Issue #83, opened 2014-02-05
Add ability to perform a "range join" to RDDs of ADAMRecords. 


tdanford:
Right now, we can use Spark's RDD 'join' method to perform equi-joins on ADAMRecords or other types -- joining records based on the equality of one or more fields. 

However, for several different genomics applications will require a "range join" -- joining records based on whether they overlap within a genomic coordinate system.  This issue is written for the task of extending an RDD[ADAMRecord] to support range join on ADAMRecords or other record types which contain a reference chromosome and start/end coordinates.

PR #117, opened 2014-02-19
Add a "range join" method.

tdanford:
Fixes ISSUE #83.
- This allows for joining different types which are aligned to different
  ReferenceRegions
- A record will appear in multiple different places when it overlaps
  multiple different regions

We're still tinkering with this commit -- we'll be adding more comments, tests, etc. -- but feel free to review and give us comments as soon as you can.


Tagar:
Any chance this can be merged to upstream Spark as well?

heuermh:
@Tagar Hello, thank you for your comment!

It is not likely that our implementation would be generally useful; our region join APIs work on datasets of records aligned to a genomic reference.  Each record needs to be convertable to a reference region (referenceName, start, end).

Finally, while we would like to port the region join APIs from `RDD`s to Spark SQL over `Dataset`/`DataFrame`s, we haven't the bandwidth to do so.

For more details, please see the doc at

Using ADAM’s RegionJoin API
https://adam.readthedocs.io/en/latest/api/joins/

or code starting about here

https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicDataset.scala#L1090

Tagar:
@heuermh thank you for all of those details . 

I have ran into similar issues with Spark outside of genomics. I do understand this particular pull request is for ADAM project only. 

For example, if we have two `dataframe`s that both have `date_start`, `date_end` and we want to join those two dataframes be intersections of dates.. it seems to be very similar to joining genome regions. 

https://issues.apache.org/jira/browse/SPARK-8682

Or, for example, problem of finding which IP address matches which subnet (given known subnet  - start-end IPs); there is an interesting somewhat generic solution for range join explained in 
https://stackoverflow.com/questions/37953830/spark-sql-performance-join-on-value-between-min-and-max/37955947#37955947

Thanks again. Hope to see this will be fixed at some point in upstream Spark.

PR #250, opened 2014-05-29
Adding ADAM to SAM conversion.

### LINKED ISSUE #49: ADAMRecord-->BAM converter
Currently, we have a converter that converts BAM/SAM via SamRecords to ADAMRecords. However, we do not have a converter that goes from ADAMRecords to BAM/SAM. This would perform the opposite packaging from ADAMRecords-->SamRecords, the collection of SamHeader data, and then write-out through Hadoop-BAM.

fnothaft:
This commit adds a conversion function for going from ADAM back to SAM or BAM. This work was done by @erictu; I've just gone ahead and rebased it on top of master, and done some small clean up. This resolves #49.


ankushreddy:
how to perform adam actions to convert from .adam file format to bam file.

fnothaft:
@ankushreddy you'll want to do:

```
./bin/adam-submit transform [<optional spark config> -- ] myFile.adam myFile.bam
```

Issue #297, opened 2014-07-09
PartitionAndJoin should throw an exception if it sees an unmapped read


kozanitis:
PartitionAndJoin crashes with a null pointer exception when I call it to join a set of 4 mouse-chrM coordinates with a small mouse file.

You can find my mouse.bam, mouse.adam, the coordinate test.txt file and my source code here:

https://github.com/kozanitis/misc

This is the exception stack:
2014-07-09 09:39:57 WARN  TaskSetManager:70 - Loss was due to java.lang.NullPointerException
java.lang.NullPointerException
    at org.bdgenomics.adam.rdd.NonoverlappingRegions.hasRegionsFor(RegionJoin.scala:311)
    at org.bdgenomics.adam.rdd.MultiContigNonoverlappingRegions.filter(RegionJoin.scala:365)
    at org.bdgenomics.adam.rdd.RegionJoin$$anonfun$5.apply(RegionJoin.scala:125)
    at org.bdgenomics.adam.rdd.RegionJoin$$anonfun$5.apply(RegionJoin.scala:125)
    at scala.collection.Iterator$$anon$14.hasNext(Iterator.scala:390)
    at scala.collection.Iterator$$anon$13.hasNext(Iterator.scala:371)
    at scala.collection.Iterator$class.foreach(Iterator.scala:727)
    at scala.collection.AbstractIterator.foreach(Iterator.scala:1157)
    at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:158)
    at org.apache.spark.scheduler.ShuffleMapTask.runTask(ShuffleMapTask.scala:99)
    at org.apache.spark.scheduler.Task.run(Task.scala:51)
    at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:187)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)

fnothaft:
Closed as won't fix.

PR #421, opened 2014-10-15
CLI & mappability filter for region join, custom exceptions

### LINKED ISSUE #420: Create standardized, interpretable exceptions for error reporting
I think that #297 is indicative of a slightly larger problem -- the way it manifests is that, when you run `partitionAndJoin` with some non-mappable records (i.e. records that can't be turned into a ReferenceRegion), you get a NullPointerException being thrown by an un-intelligible piece of code.

The challenge isn't just to catch this error condition earlier, it's to throw exceptions whose names/messages are interpretable to the end-user.  So, for example, we could have an UnmappedException exception, which is thrown whenever "something that should be mappable, isn't."  

Other ideas for exceptions -- 
- invalid sample identifier
- insufficient read coverage (or data coverage?) 

thoughts?

tdanford:
Fixes #420, #297 

Several changes: 
* adding a filter for 'mappability' to the ``partitionAndJoin`` method, fixing (I hope) Christos's problem
* creating a CLI tool for accessing/testing RegionJoin (and for re-creating the bug) 
* adding a custom exception, for throwing errors related to unmapped reads or records (see #420) 
* added a ``hasReferenceRegion`` to ``ReferenceMapping`` -- to explicitly check for whether a value _can_ be converted to a ReferenceRegion

### DIFF
--- a/adam-cli/src/main/scala/org/bdgenomics/adam/cli/ADAMMain.scala
+++ b/adam-cli/src/main/scala/org/bdgenomics/adam/cli/ADAMMain.scala
@@ -35,13 +35,13 @@ object ADAMMain extends Logging {
         CountReads,
         CalculateDepth,
         CountKmers,
-	Adam2Fastq,
+        Adam2Fastq,
         Transform,
         /* TODO (nealsid): Reimplement in terms of new schema
 	  ComputeVariants
 	*/
-          PluginExecutor
-        )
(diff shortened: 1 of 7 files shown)


fnothaft:
@tdanford what's the status on this? Close and reopen later?

fnothaft:
Closing due to inactivity.

PR #472, opened 2014-11-12
[ADAM-463] AvroMultiParquetRDD loads multiple Parquet files

tdanford:
[ADAM-463] Added AvroMultiParquetRDD

First, updated the FileLocator to include a childLocators method, and added its implementation in all the child FileLocator implementations.

Next, added an HDFSFileLocator implementation. 

Finally, added AvroMultiParquetRDD, which takes a FileLocator which indicates the _parent_ directory (in HDFS, or S3) of a set of parquet files, and which loads _all_ row groups from all the parquet files that are immediate children of that FileLocator.
        
(Joint work with Carl Yeksigian.)


fnothaft:
Closing this as after #478 this will need to be moved to the bdg-utils repo.

Issue #511, opened 2014-11-28
ReferenceRegion shouldn't extend Ordered


tdanford:
Currently, ReferenceRegion extends Ordered (everything here also applies, mutatis mutandis, to ReferenceRegionWithOrientation).  

This is a problem semantically, as according to Scala's docs (http://www.scala-lang.org/api/current/index.html#scala.math.Ordered) Ordered should only be used for types that have "a single, natural ordering" -- while we certainly can (and will) want to sort ReferenceRegions in multiple different ways.

Second, it has become a problem concretely as well: in defining some natural operations on RDDs of ReferenceRegions, I'm coming across a "diverging implicit expansion" error (http://stackoverflow.com/questions/4883906/what-is-a-diverging-implicit-expansion-error) that seems to be related to using Ordered on ReferenceRegions and subclasses of ReferenceRegion.

The solution I'm proposing to this is threefold: 
1. removed Ordered from the superclasses of ReferenceRegion, 
2. move the compareTo method in ReferenceRegion into the compare method of a ReferenceRegionOrdering, which will extend the Ordering trait (the correct trait we should be using when there is more than one ordering on a type), and 
3. provide an implicit Ordering in a new object, ReferenceRegionContext, that can be imported by classes that need the Ordering available.

fnothaft:
Resolved since at least #624.

Issue #512, opened 2014-11-28
Need a RegionRDDFunctions class


tdanford:
I want a RegionRDDFunctions class, which wraps an RDD of ReferenceRegions, and provides "common operations" on the regions within the RDD.  The common operations I'd like to see are: 

* join by overlap: join with another RDD of regions, the output of the join being pairs of regions (one from the wrapped RDD, one from the other RDD) that overlap
* join by range: similar to join by overlap, except that regions need to be within a particular 'range' of each other (where the range is a distance measured in bases).
* filter by overlap: filters the regions of the wrapped RDD, keeping only those regions that overlap _at least one_ region of a second RDD of regions
* filter by range: similar to the filter by overlap, except that we substitute (as we did with joins) 'within X bp' for 'overlap'
* group by overlap: basically, does a join by overlap followed by a groupBy
* group by range: same as above
* window counts: calculate fixed windows of a given size across the genome, and count how many regions from the input RDD overlap each window

Any other suggestions?

heuermh:
I believe this issue should be marked as closed by #517.

fnothaft:
Agreed. Closed by #517.

PR #517, opened 2014-12-02
Adding RegionRDDFunctions

tdanford:
This PR contains two commits fixing #511 and #512.  

First, it updates ReferenceRegion so that the class doesn't extend Ordered -- rather, we providing _an_ Ordering which can be used in the same contexts (and which we can replace by other orderings, if we want to order the regions in a different way).  

Second, it provides three new RDD wrapper (*Functions) classes, around RDDs of 
1. ReferenceRegions
2. Reference-mappable values
3. Pairs whose first element is a ReferenceRegion

In each case, the wrapper class provides a standard set of operations (join, filter, groupBy, windowCounts) which operate on sets of geometric operations (and are implemented under the hood, here, with RegionJoin.partitionAndJoin).  

These PRs are necessary pre-cursors to the Mutect work that's going on over in the avocado repository.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
@@ -52,7 +52,7 @@ object ReferenceRegionWithOrientation {
  *                       reverse strand of the reference region.
  */
 case class ReferenceRegionWithOrientation(region: ReferenceRegion,
-                                          negativeStrand: Boolean) extends Ordered[ReferenceRegionWithOrientation] {
+                                          negativeStrand: Boolean) {
   def width: Long = region.width
 
   def contains(other: ReferencePositionWithOrientation): Boolean = {
@@ -67,13 +67,30 @@ case class ReferenceRegionWithOrientation(region: ReferenceRegion,
     region.overlaps(other.region) && negativeStrand == other.negativeStrand
   }
 
-  def compare(that: ReferenceRegionWithOrientation): Int = {
-    val regionCompare = region.compare(that.region)
-    if (regionCompare != 0) {
-      regionCompare
-    } else {
-      negativeStrand.compare(that.negativeStrand)
-    }
+  /**
+   * Creates a new ReferenceRegionWithOrientation, with either the start or
+   * end field modified so that the overall region has a width equal to the
+   * given target length.
+   *
+   * If the region is on the negative strand, then the start value is modified,
+   * otherwise, if the region is on the positive strand, then the end value is
+   * incremented.
+   *
+   * @param targetLength the desired width of the new region
+   * @return The new region whose width==targetLength
+   */
+  def extend(targetLength: Long): ReferenceRegionWithOrientation = {
+    require(targetLength >= 0)
+    require(targetLength > width)
+
+    val diff = targetLength - width
+
+    if (negativeStrand)
+      ReferenceRegionWithOrientation(
+        ReferenceRegion(region.referenceName, region.start - diff, region.end), negativeStrand)
+    else
+      ReferenceRegionWithOrientation(
+        ReferenceRegion(region.referenceName, region.start, region.end + diff), negativeStrand)
   }
 
   def toReferenceRegion: ReferenceRegion = region
@@ -83,6 +100,23 @@ case class ReferenceRegionWithOrientation(region: ReferenceRegion,
   def start: Long = region.start
 
   def end: Long = region.end
+}
+
+object ReferenceRegionContext {
+  implicit val referenceRegionOrdering = ReferenceRegionOrdering
+  implicit val referenceRegionWithOrientationOrdering = ReferenceRegionWithOrientationOrdering
+}
+
+object ReferenceRegionWithOrientationOrdering extends Ordering[ReferenceRegionWithOrientation] {
+
+  def compare(this_ : ReferenceRegionWithOrientation, that: ReferenceRegionWithOrientation): Int = {
+    val regionCompare = ReferenceRegionOrdering.compare(this_.region, that.region)
+    if (regionCompare != 0) {
+      regionCompare
+    } else {
+      this_.negativeStrand.compare(that.negativeStrand)
+    }
+  }
 }
 
 object ReferenceRegion {
@@ -135,6 +169,17 @@ object ReferenceRegion {
   }
 }
 
+object ReferenceRegionOrdering extends Ordering[ReferenceRegion] {
+
+  override def compare(this_ : ReferenceRegion, that: ReferenceRegion): Int =
+    if (this_.referenceName != that.referenceName)
+      this_.referenceName.compareTo(that.referenceName)
+    else if (this_.start != that.start)
+      this_.start.compareTo(that.start)
+    else
+      this_.end.compareTo(that.end)
+}
+
 /**
  * Represents a contiguous region of the reference genome.
  *
@@ -144,10 +189,10 @@ object ReferenceRegion {
  *            which is <i>not</i> in the region -- i.e. [start, end) define a 0-based
  *            half-open interval.
  */
-case class ReferenceRegion(referenceName: String, start: Long, end: Long) extends Ordered[ReferenceRegion] with Interval {
+case class ReferenceRegion(referenceName: String, start: Long, end: Long) extends Interval {
 
-  assert(start >= 0)
-  assert(end >= start)
+  assert(start >= 0, "ReferenceRegion doesn't allow a negative start coordinate (here provided %d)".format(start))
+  assert(end >= start, "ReferenceRegion requires end >= start (here provided %d,%d)".format(start, end))
 
   def width: Long = end - start
 
@@ -254,14 +299,6 @@ case class ReferenceRegion(referenceName: String, start: Long, end: Long) extend
 
   def overlaps(other: ReferenceRegion): Boolean =
     referenceName == other.referenceName && end > other.start && start < other.end
-
-  def compare(that: ReferenceRegion): Int =
-    if (referenceName != that.referenceName)
-      referenceName.compareTo(that.referenceName)
-    else if (start != that.start)
-      start.compareTo(that.start)
-    else
-      end.compareTo(that.end)
 
   def length(): Long = {
     end - start
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/RegionRDDFunctions.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/RegionRDDFunctions.scala
@@ -1,0 +1,182 @@
+/**
+ * Licensed to Big Data Genomics (BDG) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The BDG licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.bdgenomics.adam.rdd
+
+import org.apache.spark.SparkContext._
+import org.apache.spark.rdd.RDD
+import org.bdgenomics.adam.models.{ SequenceDictionary, ReferenceRegionWithOrientation, ReferenceRegion }
+import org.bdgenomics.adam.models.ReferenceRegionContext._
+import scala.reflect.ClassTag
+import scala.math.max
+
+/**
+ * Functions (joins, filters, groups, counts) that are often invoked on RDD[R] where R is a
+ * type which extends ReferenceRegion (and is often just ReferenceRegion directly).
+ *
+ * @param rdd The argument RDD
+ * @param kt The type of the argument RDD must be explicit
+ * @tparam Region The type of the argument RDD
+ */
+class RegionRDDFunctions[Region <: ReferenceRegion](rdd: RDD[Region])(implicit kt: ClassTag[Region])
+    extends Serializable {
+
+  import RegionRDDFunctions._
+
+  /**
+   * Performs an 'overlap join' on the input RDD, returning pairs whose first element is a member of the
+   * input RDD and whose second element is a member of the argument RDD which overlaps the first element
+   * of the pair.
+   *
+   * @param that The argument RDD
+   * @param kt2 The type of the values in the argument RDD must be explicit.
+   * @tparam R2 The explicit type of the regions in the argument RDD
+   * @return The RDD of pairs (f, s) where f comes from the input RDD and s comes from 'that', the argument
+   *         RDD.
+   */
+  def joinByOverlap[R2 <: ReferenceRegion](that: RDD[R2])(implicit kt2: ClassTag[R2]): RDD[(Region, R2)] =
+    BroadcastRegionJoin.partitionAndJoin[Region, Region, R2, R2](rdd.keyBy(r => r), that.keyBy(r => r))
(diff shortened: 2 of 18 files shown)


tdanford:
@laserson I'm going to resurrect this bad boy.

tdanford:
I'm going to start re-reviewing this, to make sure that (a) I met all the requests/suggestions above, and (b) that it's got reasonable test coverage.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/RegionRDDFunctions.scala
```diff
@@ -0,0 +1,187 @@
...
+ * @param rdd The argument RDD
+ * @param kt The type of the argument RDD must be explicit
+ * @tparam Region The type of the argument RDD
+ */
+class RegionRDDFunctions[Region <: ReferenceRegion](rdd: RDD[Region])(implicit kt: ClassTag[Region])
+    extends Serializable {
+
+  import RegionRDDFunctions._
+
+  private def regionToReferenceRegion(region: Region): Option[ReferenceRegion] = Option(region)
```
fnothaft: I am still strongly -1 having the option here.

fnothaft:
I'm -1 on adding any new code that uses `ReferenceMapping`. I echo @laserson's earlier comment:

> Can we eliminate ReferenceMapping entirely? Could we instead just use a function that takes some type and returns a ReferenceRegion?

tdanford:
Should I wait for your PR that eliminates ReferenceMappings altogether?

fnothaft:
@tdanford I won't be able to get to that until later this week, at earliest. I don't think it should effect this code, as it is all new. Instead of relying on `ReferenceMappable[T]`, RDDs should be `Tuple2[ReferenceMappable, T]`.

tdanford:
@fnothaft I've rebased and then will fix these errors...

fnothaft:
I would like to merge #592 before this PR.

adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceMapping.scala
```diff
@@ -17,7 +17,7 @@
  */
 package org.bdgenomics.adam.models
 
-trait ReferenceMapping[T] {
+trait ReferenceMapping[T] extends Serializable {
```
laserson: Ok, this is obviously gone now :)

adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
```diff
@@ -85,6 +102,23 @@ case class ReferenceRegionWithOrientation(region: ReferenceRegion,
...
+    val regionCompare = ReferenceRegionOrdering.compare(this_.region, that.region)
+    if (regionCompare != 0) {
+      regionCompare
+    } else {
+      this_.negativeStrand.compare(that.negativeStrand)
+    }
+  }
+}
+
 object ReferenceRegion {
```
laserson: Minor point, but could we move `ReferenceRegionWithOrientation` below `ReferenceRegion`?  Just seems more natural to me.

adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
```diff
@@ -67,13 +67,30 @@ case class ReferenceRegionWithOrientation(region: ReferenceRegion,
...
+   * given target length.
+   *
+   * If the region is on the negative strand, then the start value is modified,
+   * otherwise, if the region is on the positive strand, then the end value is
+   * incremented.
+   *
+   * @param targetLength the desired width of the new region
+   * @return The new region whose width==targetLength
+   */
+  def extend(targetLength: Long): ReferenceRegionWithOrientation = {
```
laserson: Should you be able to extend in either direction?  Perhaps there should be a flag or a separate method to allow the user to extend the 5' or the 3' end.

adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
```diff
@@ -85,6 +102,23 @@ case class ReferenceRegionWithOrientation(region: ReferenceRegion,
   def end: Long = region.end
 }
 
+object ReferenceRegionContext {
+  implicit val referenceRegionOrdering = ReferenceRegionOrdering
+  implicit val referenceRegionWithOrientationOrdering = ReferenceRegionWithOrientationOrdering
```
laserson: I don't think I understand what these two implicits are.

fnothaft: These are for #511.

adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
```diff
@@ -144,10 +189,10 @@ object ReferenceRegion {
  *            which is <i>not</i> in the region -- i.e. [start, end) define a 0-based
  *            half-open interval.
  */
-case class ReferenceRegion(referenceName: String, start: Long, end: Long) extends Ordered[ReferenceRegion] with Interval {
+case class ReferenceRegion(referenceName: String, start: Long, end: Long) extends Interval {
```
laserson: Thoughts about eliminating `Interval`?  Are there any use cases where you need an `Interval` and are not dealing with `ReferenceRegion`s?

fnothaft: In my dream world, I would eliminate both `Interval` and (more controversially) `ReferencePosition`, and move `ReferenceRegion` to be an Avro model.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/RegionRDDFunctions.scala
```diff
@@ -0,0 +1,187 @@
...
+
+/**
+ * Functions (joins, filters, groups, counts) that are often invoked on RDD[R] where R is a
+ * type which extends ReferenceRegion (and is often just ReferenceRegion directly).
+ *
+ * @param rdd The argument RDD
+ * @param kt The type of the argument RDD must be explicit
+ * @tparam Region The type of the argument RDD
+ */
+class RegionRDDFunctions[Region <: ReferenceRegion](rdd: RDD[Region])(implicit kt: ClassTag[Region])
```
laserson: Using "Region" for a type parameter is confusing to me.  Perhaps change to `R` or `T`?

fnothaft: +1 @laserson

adam-core/src/main/scala/org/bdgenomics/adam/rdd/RegionRDDFunctions.scala
laserson: I'll re-+1 Frank's -1 to having an `Option` here :)

adam-core/src/main/scala/org/bdgenomics/adam/rdd/RegionRDDFunctions.scala
```diff
@@ -0,0 +1,187 @@
...
+   * input RDD and whose second element is a member of the argument RDD which overlaps the first element
+   * of the pair.
+   *
+   * @param that The argument RDD
+   * @param kt2 The type of the values in the argument RDD must be explicit.
+   * @tparam R2 The explicit type of the regions in the argument RDD
+   * @return The RDD of pairs (f, s) where f comes from the input RDD and s comes from 'that', the argument
+   *         RDD.
+   */
+  def joinByOverlap[R2 <: ReferenceRegion](that: RDD[R2])(implicit kt2: ClassTag[R2]): RDD[(Region, R2)] =
```
laserson: Consider removing, as there are multiple join ops, and this is something simple enough that a user can express themselves.  (Evidence: it's a one-liner)

adam-core/src/main/scala/org/bdgenomics/adam/rdd/RegionRDDFunctions.scala
```diff
@@ -0,0 +1,187 @@
...
+   *
+   * @param that The argument RDD
+   * @param range A long integer; if two regions are within 'range' of each other (along the same chromosome),
+   *              then they are present in the returned RDD.
+   * @param kt2 The type of the values in the argument RDD must be explicit.
+   * @tparam R2 The explicit type of the regions in the argument RDD
+   * @return The RDD of pairs (f, s) where f comes from the input RDD and s comes from 'that', the argument
+   *         RDD.
+   */
+  def joinWithinRange[R2 <: ReferenceRegion](that: RDD[R2], range: Long)(implicit kt2: ClassTag[R2]): RDD[(Region, R2)] =
```
laserson: Actually, I'll switch my "nice!" to a "remove this".

laserson: Another alternative is to move these fns from here to the corresponding join impls.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/RegionRDDFunctions.scala
fnothaft: It also locks the user in to the `BroadcastRegionJoin`. At the least, the user should be able to specify a join strategy.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/RegionRDDFunctions.scala
fnothaft: Can we specify the joins via some Join trait? This is layered on top of the join implementation, and thus would fit well in a trait.

laserson:
Actually, looking at this again, I think perhaps it'd be best to nix this PR entirely (almost).  Perhaps we can provide a `trait` for a join implementation that the broadcast and shuffle joins both implement, and include some of the functions provided here?

fnothaft:
> Actually, looking at this again, I think perhaps it'd be best to nix this PR entirely (almost). Perhaps we can provide a trait for a join implementation that the broadcast and shuffle joins both implement, and include some of the functions provided here?

+1

adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
laserson: But what does this actually do?

adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
laserson: +1, what's stopping us?  Are there compelling reasons against?  There is something nice about having `ReferenceRegion` as a case class I think, at least syntactically.

adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
fnothaft: Oh. Good question. I'm not sure... _Howaaayver_, I think that this is necessary to provide an implicit conversion from `ReferenceRegion` to `Ordering[ReferenceRegion]`, which is otherwise provided (???) by having `ReferenceRegion` extend `Ordered`. TBH, IDK.

laserson: Do we know if this fails to build if we delete it?

adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
fnothaft: I don't know if there's anything stopping us, per se. I've just developed these feelings over the past few weeks (been playing around with a lot of stuff in avocado). The syntactical stuff can be fixed by having a `ReferenceRegion` object, I think, which we'd want anyways to implement the current functionality that we've got baked into `ReferenceRegion`.

adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
fnothaft: I _think_ but cannot prove that it would fail to build. @tdanford is probably the best person to comment, but I do think this is necessary for #511, and I do agree that #511 is something that should be fixed. We'd need to fix #511 if we were to move `ReferenceRegion` into Avro, for instance.

adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
laserson: k, in that case, I'd support removing `Interval` and `ReferencePosition`, and hold off moving `ReferenceRegion` to Avro.

fnothaft: Well, IMO, there's a bit more to the story than that. The big advantage of moving `ReferenceRegion` to Avro, is that we could clean up our Avro records. `AlignmentRecord` would not have start, end anymore. It would have a `ReferenceRegion`. Ditto for `Variant`, `Feature`, `NucleotideContigFragment`, etc.

fnothaft:
Adding a few things to @laserson's comment above.

* I'm hesitant to add any code that doesn't have a clear use case right now.
* I'm also hesitant to merge large changes without profiling for performance.
* I would strongly prefer a `Join` trait that would contain these functions.
* I don't understand why many of these functions take just a `ReferenceRegion` as an input. Or, to rephrase this question, what is the meaning of an `RDD[ReferenceRegion]`? I can think of a few special cases, but they're pretty rare.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/RegionRDDFunctions.scala
tdanford: I think I wrote this originally while BroadcastRegionJoin was still the only option in town.

fnothaft:
@tdanford shall we close this and rebuild via the trait in #600?

laserson:
+1 to closing this.  I think the landscape has changed a bit...

adam-core/src/main/scala/org/bdgenomics/adam/rdd/RegionRDDFunctions.scala
```diff
@@ -0,0 +1,216 @@
...
+ * Functions (joins, filters, groups, counts) that are often invoked on RDD[R] where R is a
+ * type which extends ReferenceRegion (and is often just ReferenceRegion directly).
+ *
+ * @param rdd The argument RDD
+ * @param kt The type of the argument RDD must be explicit
+ * @tparam Region The type of the argument RDD
+ */
+class RegionRDDFunctions[Region <: ReferenceRegion](rdd: RDD[Region])(implicit kt: ClassTag[Region])
+    extends Serializable {
+
```
tdanford: @fnothaft this is the class -- I can see you're already all over it.

fnothaft:
@tdanford Thanks for the ping on this. I'm -1 on merging this until some fairly substantive changes are made. Specifically:

* I don't think a class that provides primitives for operating on types `R <: ReferenceRegion` is valuable. I would be a lot more interested in this pull request if everything except for the `RegionKeyedRDDFunctions` class was eliminated.
* Any join-based algorithm must be able to choose a generic join engine (anything that extends the `RegionJoin` trait). I realize that the `RegionJoin` trait was added after you started this pull request, but I think the correct approach would be to add these methods to the `RegionJoin` trait.
* Unless this code is performance validated or solves a clear problem, I'm hesitant to merge it into ADAM core.

I suppose the bigger problem at the end of the day is that I'm not convinced about the benefit of adding this code versus the cost of maintaining it. I agree that `joinByOverlap` is useful, but as Uri points out, it is a one liner. There's also application-specific reasons that `joinByOverlap` isn't as general as one would hope (e.g., you often need to actually expand the object you're joining against, not the region where it sits). There are also methods that look very expensive (filterOnOverlaps) that don't solve a problem that I'm aware of. I do realize though that this could just be my personal ignorance, so let me know if I'm overlooking something. Instead of merging into the ADAM core, I think it would make sense to "incubate" this code into the downstream repository that needs the features, while working on performance and the API.

fnothaft:
Closing due to inactivity.

Issue #524, opened 2014-12-17
Parquet/S3 integration tests fail


fnothaft:
Paging @tdanford @carlyeks 

It looks like the `S3Test` tagged tests were never getting run. Somewhere along the way, they started failing. This is what's causing #472 to fail as well, I believe. I'll open a PR momentarily that demonstrates the bug.

fnothaft:
Closing this, as after #478 this issue is now against bdg-utils.

PR #478, opened 2014-11-14
Move non-genomics code

fnothaft:
This resolves #475 by removing the non-genomics specific code which has been migrated to [bdg-utils](https://www.github.com/bigdatagenomics/bdg-utils), and pointing at bdg-utils where necessary. Specifically, we've moved the:

* SparkFunSuite
* Parquet reimplementation
* Metrics
* Instrumentation

PR #525, opened 2014-12-17
Attempted to clean up issues in pom.xml.

fnothaft:
This reveals #524. I was playing around with the AvroParquetRDD and ran into a few issues; specifically, the Apache Commons HTTP library that we were depending on _in ADAM_ is in conflict with the version that the AWS SDK needs. Removing this conflict causes the compile to fail. I also found that we weren't running the `S3Test` tagged unit tests at all, which is fun.

Also, as an aside, why was the AWS SDK library marked as `provided`? Am I missing something (i.e., is there a way to side step the dependency conflict above), or...?

CC @tdanford @carlyeks

### DIFF
--- a/adam-core/pom.xml
+++ b/adam-core/pom.xml
@@ -236,7 +236,7 @@
             <groupId>org.scalatest</groupId>
             <artifactId>scalatest-maven-plugin</artifactId>
             <configuration>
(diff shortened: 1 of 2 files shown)


massie:
Jenkins, test this please.

fnothaft:
Closing this, as after #478, this will need to be moved to bdg-utils.

PR #535, opened 2014-12-24
[ADAM-441] put a check in for Nothing.  Throws an IAE if no return type is provided

calvertj:
Let me know if you want me to change anything.  I also updated the
deprecated code and specified a return type for all implicits in
ADAMContext.scala

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
@@ -49,7 +49,7 @@ object ADAMContext {
   implicit def sparkContextToADAMContext(sc: SparkContext): ADAMContext = new ADAMContext(sc)
 
   // Add generic RDD methods for all types of ADAM RDDs
-  implicit def rddToADAMRDD[T <% SpecificRecord: Manifest](rdd: RDD[T]) = new ADAMRDDFunctions(rdd)
+  implicit def rddToADAMRDD[T](rdd: RDD[T])(implicit ev1: T => SpecificRecord, ev2: Manifest[T]): ADAMRDDFunctions[T] = new ADAMRDDFunctions(rdd)
 
   // Add implicits for the rich adam objects
   implicit def recordToRichRecord(record: AlignmentRecord): RichAlignmentRecord = new RichAlignmentRecord(record)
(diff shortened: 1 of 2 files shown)


adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -49,7 +49,7 @@ object ADAMContext {
   implicit def sparkContextToADAMContext(sc: SparkContext): ADAMContext = new ADAMContext(sc)
 
   // Add generic RDD methods for all types of ADAM RDDs
-  implicit def rddToADAMRDD[T <% SpecificRecord: Manifest](rdd: RDD[T]) = new ADAMRDDFunctions(rdd)
+  implicit def rddToADAMRDD[T](rdd: RDD[T])(implicit ev1: T => SpecificRecord, ev2: Manifest[T]): ADAMRDDFunctions[T] = new ADAMRDDFunctions(rdd)
```
laserson: I'm a relative scala noob.  Can you explain the significance of this change?

laserson:
@calvertj Could you rebase this on master?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
calvertj: @laserson sure!  He is my attempt to explain it.

View Bounds, for which the syntax is as follows with <% :
def myFun[A <% B](a: A) : B = a
basically mean that A can be treated like B. These are deprecated in 2.10, here is the original thread:
https://groups.google.com/forum/#!topic/scala-internals/hNow9MvAi6Q/discussion.
Instead we require that A must have an implicit method which converts A to B with the new syntax:
def myFun[A](a: A)(implicit ev1: A => B): B = a
ev1 here is short for evidence 1.
Here is a good article on it:
http://jatinpuri.com/2014/03/replace-view-bounds/
I think the newer syntax is clearer, but thats really the difference to us mere mortals, syntax.

laserson: seems pretty clear...thanks!

calvertj: Glad to hear it, I rebased.

laserson:
Thanks, @calvertj!

calvertj:
My pleasure, I love the mission, and Scala, plus you guys have a nice code base. Hope I can find the time to do some more.

PR #548, opened 2015-01-20
Multiline FASTQ input

cket:
[ADAM-386] Adds support for multiline FASTQ input

This is my first pull request, so let please let me know of any improvements I should/can make! 
Thanks,
Chris


massie:
Add to whitelist. Jenkins, test this please.

cket:
Oops.. I accidentally changed ".adam" to "adam", causing the test to fail. Everything should work fine now!

massie:
Jenkins, test this please.

adam-core/src/main/java/org/bdgenomics/adam/io/InterleavedFastqInputFormat.java
```diff
@@ -123,16 +127,15 @@ public InterleavedFastqRecordReader(Configuration conf, FileSplit split) throws
          */
         private void positionAtFirstRecord(FSDataInputStream stream) throws IOException {
             Text buffer = new Text();
-
+            int bytesRead = 0;
+            
             if (true) { // (start > 0) // use start>0 to assume that files start with valid data
```
fnothaft: If this is now `if(true)`, shouldn't we just eliminate the if clause?

adam-core/src/main/java/org/bdgenomics/adam/io/InterleavedFastqInputFormat.java
```diff
@@ -234,7 +241,43 @@ public float getProgress() {
...
+
+                //read is 1 line. Read 2 more lines to set position the reader at the next @title line
+                bytesRead = reader.readLine(buffer,(int)Math.min(MAX_LINE_LENGTH, end - start));
+                bytesRead = reader.readLine(buffer,(int)Math.min(MAX_LINE_LENGTH, end - start));
+                return sequenceLines; // all good!
+            } else {
+
+                String bufferString = buffer.toString();
+                //check if the record is multiline and we are still reading the sequence data
+                while(bytesRead > 0 && buffer.getLength() > 0 && bufferString.matches("[A-Z|a-z]*")){
```
fnothaft: I think you can tighten this down to the IUPAC codes.

adam-core/src/main/java/org/bdgenomics/adam/io/InterleavedFastqInputFormat.java
```diff
@@ -246,15 +289,30 @@ protected boolean lowLevelFastqRead(Text readName, Text value) throws IOExceptio
...
+                //it has not yet been called
+                sequenceLines = getLineInfo();
+            }else{
+                //it has been called. Take line info from readLines and set it to -2.
+                sequenceLines=readLines;
+                readLines=-2;
+            }
 
             // sequence
+            for(int x = 0; x<sequenceLines; x++){
```
fnothaft: This is a small detail, but in general, in scala it is more efficient to do `(0 to sequenceLines).foreach(x => {` instead of `for(int x = 0; x<sequenceLines; x++) {`

adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
```diff
@@ -74,8 +82,34 @@ class FastqRecordConverter extends Serializable with Logging {
         .build())
   }
 
+  def parseMultiLine(element: String): Array[String] = {
+    val fastqRegex = "\\n(?=\\+)"
+
+    val splitOne = element.split("\\n", 2)
```
fnothaft: What is all the splitting necessary here for? If you could add a few inline comments, that would help to clarify.

adam-core/src/test/scala/org/bdgenomics/adam/rdd/read/ADAMAlignmentRecordContextSuite.scala
```diff
@@ -152,7 +172,7 @@ class ADAMAlignmentRecordContextSuite extends SparkFunSuite {
    same location as createTempFile() uses, so therefore the returned path from this method
    should be suitable for adamParquetSave().
    */
-  def tempLocation(suffix: String = ".adam"): String = {
+  def tempLocation(suffix: String = "adam"): String = {
```
fnothaft: This change is what caused the unit test failure; IMO, it should probably go back to the old version.

fnothaft:
@cket good work! I dropped a few comments inline. When you make changes, can you run `scripts/format-source`? This script will reformat your code to match the project's formatting guidelines.

cket:
Thanks! I've all the changes except the foreach loop, which I wanted to check before I made another commit. You mentioned that the foreach is more efficient in scala, but that code is java. Did you mean I should use the java foreach instead? What is the advantage of the foreach over the for in this situation?

fnothaft:
> Thanks! I've all the changes except the foreach loop, which I wanted to check before I made another commit. You mentioned that the foreach is more efficient in scala, but that code is java. Did you mean I should use the java foreach instead? What is the advantage of the foreach over the for in this situation?

D'oh! Sorry about the confusion there; that's the problem with squeezing code review in early in the AM... Since it _is_ Java, you are correct.

The Scala for/foreach issue has been discussed elsewhere on the internet, but the TL;DR is that `for` loops in Scala are less efficient than loops implemented via `foreach`, `while`, or `@tailrec`. There are a variety of reasons for this, but IIRC it mostly boils down to 1) Scala's `for` loops have higher object allocation overhead 2) Scala's compiler emits less efficient code for `for` loops. I'm not sure how this has tracked in newer versions of Scala, but a few folks at Berkeley benchmarked this on 2.10.3 (?) and saw a meaningful performance overhead.

cket:
Ok, I've made the changes. I'm not sure what happened with the last commit though. The code is fine, but message is wrong (it was called 'multiline FASTQ support') and it is saying that I added them yesterday. I pulled from my remote to my local repository earlier, but that is showing up as a commit as well. I'm sorry about that, I don't know what happened.

massie:
@cket You submitted this pull request (PR) using the `master` branch of your fork. Any changes that you make (and push) to this branch will be reflected in this PR.

In the future, I'd recommend that you create a topic branch for each PR you submit.

fnothaft:
@cket if you'd like to clean this up, I'd suggest moving your work to another branch, and then squashing down. *Read all the way through before deciding to do this ;).* You can do this by creating a new branch that starts from your current commit:

```
git checkout -b multiline-fastq
```

Then, you can squash the work you've done down into a single commit. [This is a good guide](http://git-scm.com/book/en/v2/Git-Tools-Rewriting-History#Squashing-Commits), but the *TL;DR* is that you'll run:

```
git rebase -i HEAD~3
```

and keep the first commit as `pick` and change the remaining commits to `squash`. When you do this, you can also rewrite your commit message; if you could change it to something like "[ADAM-386] Adds support for multiline FASTQ input" (to quote from your original PR's message), that'd be great!

Now, to fix your master branch, let's switch back to that branch:

```
git checkout master
```

We're going to reset your master branch to match the main fork's branch. *Before you do this, make 100% sure that the `multiline-fastq` branch has your work!* First, let's add this repo as an upstream repository:

```
git remote add upstream https://www.github.com/bigdatagenomics/adam.git
```

Then, we will force reset it:

```
git fetch upstream
git reset --hard upstream/master
```

Since this blows away the work on the [cket/adam/master](https://github.com/cket/adam/tree/master) branch, you'll need to reopen this PR (not a big deal). You should rebase on [top-of-tree](https://github.com/bigdatagenomics/adam/tree/master) first:

```
git checkout -b multiline-fastq
git rebase upstream/master
git push origin multiline-fastq
```

_Alternatively_, if this whole process above sounds like a bit of a pain, you can just do the squash and rebase part:

```
git rebase -i HEAD~3
git remote add upstream https://www.github.com/bigdatagenomics/adam.git
git fetch upstream
git rebase upstream/master
```

To echo @massie, I would suggest doing your work on a topic branch in the future. You can even ensure that it is based off of ToT by creating the branch as:

```
git checkout -b my_branchname upstream/master
```

PR #557, opened 2015-01-25
Randomize ports and turn off Spark UI to reduce bind exceptions in tests

massie:

PR #558, opened 2015-01-25
Makes Features2ADAMSuite use SparkFunSuite

fnothaft:
Spawned by comments in #557; `Features2ADAMSuite` was using `FunSuite` and `Features2ADAM.run()` was creating a `SparkContext`. This led to address bind errors. This PR changes it so that we pass our own `sc` to `Features2ADAM`, and that this `sc` is created by a `SparkFunSuite`.


fnothaft:
Whoops, error on my end.

massie:
Thanks, Frank!

PR #559, opened 2015-01-25
Use a lock file to allow only one jenkins test script at a time

massie:


massie:
Looking at this I realized that Spark 1.1.0 is hard-coded into out jenkins test script. We'll need to address that for Spark 1.2.0 support. We are also downloading Spark each time we run the script; we should probably cache the tarball locally instead.

massie:
Bother. `lockfile` isn't available on the workers. I'll look into this tomorrow with Shane.

fnothaft:
Jenkins, test this please.

fnothaft:
Ping @massie @shaneknapp to install `lockfile` on Jenkins.

shaneknapp:
python, i assue...  right?

fnothaft:
Nah, tis a command: [man lockfile](http://linux.die.net/man/1/lockfile)

massie:
I'm starting to question this approach now. A better fix might be to configure Jenkins to guarantee one build per worker. I'll chat with Shane about this more later. I'm going to close this for now.

PR #564, opened 2015-02-03
[ADAM-563] Add warning message when saving Parquet files with incorrect extension

### LINKED ISSUE #563: Warn when saving Parquet files without "*.adam" extension
Equivalent to #523, but on save. I'm electing to print a warning instead of erroring out, since this is an issue you could hit after running a long job.

fnothaft:
Resolves #563. Prints warning if file does not have "*.adam" extension.


fnothaft:
Rebased on ToT.

Issue #456, opened 2014-11-03
BQSR throws error if VCF is not provided in Transform


fnothaft:
If you do not provide a VCF file with known sites to BQSR, it throws an error:

```
Exception in thread "main" java.lang.IllegalArgumentException: Can not create a Path from a null string
        at org.apache.hadoop.fs.Path.checkPathArg(Path.java:123)
        at org.apache.hadoop.fs.Path.<init>(Path.java:135)
        at org.apache.spark.SparkContext.newAPIHadoopFile(SparkContext.scala:638)
        at org.bdgenomics.adam.rdd.variation.VariationContext.adamVCFLoad(VariationContext.scala:48)
        at org.bdgenomics.adam.cli.Transform.run(Transform.scala:133)
        at org.bdgenomics.adam.cli.ADAMSparkCommand$class.run(ADAMCommand.scala:58)
        at org.bdgenomics.adam.cli.Transform.run(Transform.scala:90)
        at org.bdgenomics.adam.cli.ADAMMain$.main(ADAMMain.scala:96)
        at org.bdgenomics.adam.cli.ADAMMain.main(ADAMMain.scala)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:606)
        at org.apache.spark.deploy.SparkSubmit$.launch(SparkSubmit.scala:328)
        at org.apache.spark.deploy.SparkSubmit$.main(SparkSubmit.scala:75)
        at org.apache.spark.deploy.SparkSubmit.main(SparkSubmit.scala)
```

fnothaft:
Closed by #567.

PR #567, opened 2015-02-04
[ADAM-553] Unify contexts

### LINKED ISSUE #553: Simplify `ADAMContext`
I am finding `ADAMContext` to be more complicated than I'd like, especially when I am thinking about the Python API and the `JavaADAMContext`.  Here are a few questions:

* What is the purpose of `adamLoad`?  We only have a relatively small number of types.  Does it make sense to break `adamLoad` out to the non-generic cases we are interested in?  We already have a `loadAlignments` function.

* What is the purpose of `adamDictionaryLoad`?  It looks like it's basically an agg on an `RDD[AlignmentRecord]`.  Why don't we have the user load the RDD themselves, and then have a function `RDD[AlignmentRecord] => SequenceDictionary`?  If the use wants to do it on a `.bam` file, she can use `adamBamDictionaryLoad`.

* What is the purpose of `maybeLoadBam`?  Make the user check that the file has the proper extension.  How is it useful to return an `Option`?

* What's the purpose of `FeaturesContext`?  It just adds some loading functions.  Could be moved to `ADAMContext`.  Same with `VariationContext`.

* What's the purpose of `GeneContext`?  Seems simpler to implement it by loading GTF `Feature` objects and then transforming the `RDD[Feature]` to `RDD[Gene]`.  Actually, this is what it does, but the implicit pimp-my-lib architecture just seems more confusing.

* `VariantContext` is named strangely as well.

* It's be great to consolidate some of the `*RDDFunctions` whenever possible, though not as obvious to me yet exactly what/how.

A lot of this automagically calling functions by looking at extensions seems like over-complication, and not really that useful.  How often will it happen that the user doesn't know whether they're loading data from a `.bam` file or an `.adam` file?

Why are there so many `*Context` objects?

Thoughts on cleanup/simplification?

fnothaft:
Resolves #553 and #456. First 5 commits remove the public `*Context` APIs, and the last commit removes the private `*Context` APIs. All functionality is consolidated down to `ADAMContext`, or moved into the `*RDDFunctions` classes.


fnothaft:
Rebased on ToT.

massie:
Manually pushed. Thanks, Frank!

fnothaft:
@massie I don't think this was merged properly... these commits aren't in the [master branch](https://github.com/bigdatagenomics/adam/commits/master). Can you double check this? I'll rebase these commits.

fnothaft:
Jenkins, retest this please.

massie:
I just manually merged this but correctly this time. Sorry I didn't get it right the first time.

fnothaft:
No problem! Thanks for the quick fix!

PR #568, opened 2015-02-04
Moved the Timers object from bdg-utils back to ADAM

nfergu:
This moves the Timers object back from bdg-utils to ADAM, as it contains ADAM-specific timers. 

I'll raise another PR to remove the Timers object from bdg-utils, but I don't think it's necessary to make a new bdg-utils release or anything.


fnothaft:
Merged! Thanks for catching this @nfergu!

PR #569, opened 2015-02-05
Typo in the narrowPeak parser

nikhilRP:


laserson:
Good catch! Thanks, @nikhilRP!

PR #576, opened 2015-02-08
Changed hashCode implementations to improve performance of BQSR

nfergu:
Some notes:

 * When looking at the performance of BQSR the `Util.hashCombine` method appeared to perform poorly, so I have replaced its usage in `CovariateKey` and `QualityScore` with hash codes based on the recipe from Programming in Scala (http://www.artima.com/pins1ed/object-equality.html). 
 * On my small test data set of about 90,000 reads this change makes `ObservationAccumulator.+=` about 3.5 times faster, making BQSR about 50% faster in total (not including times for loading, saving files etc). I'm aware that this is not a particularly realistic test, but I think it's probably sufficient for such a low-risk change.
 * I have also removed `Util.hashCombine` method and replaced its other usages, since others may use it without being aware of the performance penalty.
 * This change also included some new timers in the `Timers` object

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/instrumentation/Timers.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/instrumentation/Timers.scala
@@ -48,7 +48,11 @@ object Timers extends Metrics {
   // Recalibrate Base Qualities
   val BQSRInDriver = timer("Base Quality Recalibration")
   val CreateKnownSnpsTable = timer("Create Known SNPs Table")
+  val ComputeCovariates = timer("Compute Covariates")
+  val ObservationAccumulatorComb = timer("Observation Accumulator: comb")
+  val ObservationAccumulatorSeq = timer("Observation Accumulator: seq")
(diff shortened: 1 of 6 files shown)


fnothaft:
Nice catch @nfergu ! The aggregate in BQSR is notably slow on large datasets. I'm going to be doing a big cluster run later this week, so I'll give this a run on the high coverage NA12878 to benchmark perf and concordance.

adam-core/src/main/scala/org/bdgenomics/adam/instrumentation/Timers.scala
```diff
@@ -48,7 +48,11 @@ object Timers extends Metrics {
   // Recalibrate Base Qualities
   val BQSRInDriver = timer("Base Quality Recalibration")
   val CreateKnownSnpsTable = timer("Create Known SNPs Table")
+  val ComputeCovariates = timer("Compute Covariates")
```
massie: You know. Whoever wrote that timer library deserves kudos; it sure is useful. :)

massie:
Thanks @nfergu !

PR #579, opened 2015-02-10
[ADAM-578] Update end of read when trimming.

### LINKED ISSUE #578: Reference position vs. number of bases check fails if trimming before BQSR
Reported on the [mailing list](https://groups.google.com/forum/#!topic/adam-developers/UR5UVQ5-f3E).

It looks like inside of [DecadentRead](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rich/DecadentRead.scala) the `require` at line 70 is tripping. I _think_ this is because we're not updating a start/end position correctly when running trimming on an aligned read.

fnothaft:
Resolves #578. Adds logic to track the end of the read when trimming. Evidently, this wasn't in there. I think this code may have been in the repository before we added the `end` field back, and we forgot to tack the `end` logic back on.


massie:
Thanks, Frank!

PR #580, opened 2015-02-14
Add support for converting contigs into reads

### LINKED ISSUE #571: Load FASTA contigs into AlignmentRecords
This is pretty common for assembly related tasks. We should be able to go `NucleotideContigFragment` to `AlignmentRecord` (which is pretty straightforward), which also allows us to do FASTA to `AlignmentRecord`.

fnothaft:
Resolves #571. Adds support for loadings FASTA files and `*.contig.adam` files in as `AlignmentRecords`. Depends on #567.

### DIFF
--- a/adam-core/src/test/scala/org/bdgenomics/adam/converters/FragmentConverterSuite.scala
+++ b/adam-core/src/test/scala/org/bdgenomics/adam/converters/FragmentConverterSuite.scala
@@ -1,0 +1,168 @@
+/**
+ * Licensed to Big Data Genomics (BDG) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The BDG licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


fnothaft:
Jenkins, test this please.

adam-core/src/test/scala/org/bdgenomics/adam/converters/FragmentConverterSuite.scala
```diff
@@ -0,0 +1,168 @@
...
+    val ctg3Reads = reads.filter(_.getContig.getContigName == "ctg3")
+    assert(ctg3Reads.length === 1)
+
+    val ctg3Read = ctg3Reads.head
+    assert(ctg3Read.getSequence === "AATTCCGGCCTTAA")
+    assert(ctg3Read.getContig.getContigName === "ctg3")
+    assert(ctg3Read.getStart === 14L)
+    assert(ctg3Read.getEnd === 28L)
+  }
+}
```
massie: How does this code handle the case of overlapping contigs? I don't see a test case for that.

fnothaft: @massie for right now, we assume that we don't have overlapping contigs. That is mostly because we are importing from FASTA and we assume that contigs are unique.

fnothaft: I think that we shouldn't add support for overlapping contigs now. If this is an important case for folks in the future, we can address it then.

massie:
Thanks, Frank!

PR #584, opened 2015-02-21
[ADAM-583] Add k-mer counting functionality for nucleotide contig fragments

### LINKED ISSUE #583: Add support for counting k-mers from contigs
This is a follow on from #571. It isn't efficient to convert contigs to reads and _then_ count _k_-mers.

fnothaft:
Resolves #583. This steals some code from https://github.com/bigdatagenomics/avocado/pull/127 that adds flanking sequence to adjacent fragments. Without adding the adjacent sequence, we won't get correct values.


massie:
Thanks, Frank!

PR #593, opened 2015-02-28
Performance enhancements for Recalibrator class

nfergu:
This PR includes some performance enhancements for the `Recalibrator` class, which is used for BQSR.

The `RecalibrationTable` class has been restructured so that instead of having three separate maps (one for the global values, one for the quality values, and one for the extra values) we now have a 3 level tree-like structure with the global values at the root, the quality values at the next level and the extra values at the bottom (a map of maps of maps). This reduces the complexity of the keys that we need to use to look-up the values. For example, when looking up the extra values we no longer have to use the read group and quality score as part of key, since they have been looked-up already.

This makes the `Recalibrator.computeQual` method around 1.7 times faster on my small sample of around 90,000 reads, leading to around a 17% improvement in the overall performance of BQSR.

Unfortunately I don't think the new version is as elegant as the old one. There's more state to be passed around, and the code is slightly more complex. But hopefully the performance improvements are worth it.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/instrumentation/Timers.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/instrumentation/Timers.scala
@@ -53,6 +53,7 @@ object Timers extends Metrics {
   val ObservationAccumulatorSeq = timer("Observation Accumulator: seq")
   val RecalibrateRead = timer("Recalibrate Read")
   val ComputeQualityScore = timer("Compute Quality Score")
+  val GetExtraValues = timer("Get Extra Values")
 
   // Realign Indels
   val RealignIndelsInDriver = timer("Realign Indels")
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/recalibration/ObservationTable.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/recalibration/ObservationTable.scala
@@ -114,14 +114,6 @@ class ObservationTable(
     val space: CovariateSpace,
     val entries: Map[CovariateKey, Observation]) extends Serializable {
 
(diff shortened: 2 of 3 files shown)


fnothaft:
Nice! I'll give this a run on the high coverage NA12878 this week to see what the performance is like on a very large dataset. I'll try to give the code a look through tomorrow. I was doing some performance benchmarking recently and was surprised to notice that the recalibration was more expensive than the aggregation, so I'm excited to see your changes.

hannes-ucsc:
As per Frank's request, we are going to run this PR against NA12878 today.

beaunorgeot:
Ran transform -bqsr with nfergu fix #593 using the 233GB NA12878.bam as input. Job finished successfully in 60mins

fnothaft:
@beaunorgeot nice! That would compare to the 74min run we had on the prior implementation of bqsr. Did you get similar timings (~75min) for the prior implementation?

beaunorgeot:
I don't have a time point for the prior implementation, but I can
run/record that in the next couple of days to provIde a sheet comparison

beaunorgeot:
Your time trial for the prior implementation used a .adam for input, so
it's possible that the #593 fix represents a speed up of even more then 15
mins

fnothaft:
@beaunorgeot nice! I didn't realize that you were running directly from .bam input. I'm going to go ahead and rebase this and merge it manually.

fnothaft:
Committed manually via https://github.com/bigdatagenomics/adam/commit/32af274e849767aec9d9d0d7f75e4bef4e8966a3. This is wonderful code, @nfergu! Thanks for pushing the change, and apologies for the slow merge from our end.

nfergu:
Great, thanks everybody, glad to see we got a reasonable speed-up here.

PR #607, opened 2015-03-02
Added ReferenceFile trait, modified TwoBitFile to extend it

ananth-pallaseni:


fnothaft:
Jenkins, add to whitelist.

laserson:
I'm fine with this. @ananth-pallaseni: are you implementing another alternative to a 2bit file?

Looks like there may also be formatting problems?  Try running `scripts/format-source`

fnothaft:
@laserson some testing work for RNAdam. Essentially, we need to mock a 2bit file.

Issue #609, opened 2015-03-02
Eliminate `applyPredicate` from `ADAMContext`


laserson:
In our recent effort to polish and tighten up ADAM, the focus has been to use the ADAM repo to store primarily lower-level ETL tools that make it easy to load, store, filter, join, etc. all different kinds of genomics data.  Ideally, we'd use the lowest level abstractions given to us, which are the Spark RDDs.  As such, if you get back an RDD of genomics objects and want to filter them, you should call `.filter()` on the RDD, rather than some other ADAM-specific function that essentially wraps the same functionality.  The only load functions that should take predicates are ones that push the predicate into the Parquet code.

tdanford:
Can you provide a description though, uri?

fnothaft:
Closed by #616.

Issue #565, opened 2015-02-04
Upgrade to Parquet filter2 API


fnothaft:
Currently, we're using the old `UnboundRecordPredicate` API for reading data from Parquet. We should migrate to the newer `filter2` API, which supports block statistics.

fnothaft:
Discussion point: If we did this, could we remove the `org.bdgenomics.adam.predicates` package?

tdanford:
Ping @carlyeks

tomwhite:
The main advantage of the filter2 API is that it filters out entire row groups. This means that queries for a range of positions, say, are very efficient since they only load the relevant row groups. I wrote a test to demonstrate this: https://github.com/bigdatagenomics/adam/compare/master...tomwhite:ADAM-565-filter2. (Note that the test can't assert that only particular row groups were loaded, but I observed this in the debugger.)

On the other hand, there is currently an outstanding Parquet issue to investigate why the filter2 API is slower than the original (https://issues.apache.org/jira/browse/PARQUET-98). It would be good for that to be resolved before switching.

I wasn't sure how to take this further - should the load methods on ADAMContext be changed to take a FilterPredicate? (I.e. is ADAMPredicate being removed?) This would be in line with https://github.com/bigdatagenomics/adam/issues/608.

fnothaft:
Hi @tomwhite! Great to have you on here. The new statistics functionality in filter2 is pretty excellent. I'm hopeful that if we moved over to filter2, we'd be able to avoid needing indices to serve very small slices of data (e.g., everyone's favorite "look up dbSNP site RS12314" queries).

> I wasn't sure how to take this further - should the load methods on ADAMContext be changed to take a FilterPredicate? (I.e. is ADAMPredicate being removed?) This would be in line with #608.

I personally would like to remove `ADAMPredicate`. If we did go whole hog with #608, `ADAMPredicate` could be removed trivially. I'm about halfway in agreement with #608. I would prefer to maintain the predicate functions on the "magic" functions, but I suppose I can be swayed. That being said, it's not clear if we _could_ remove `ADAMPredicate` without going all-in on #608.

In short,  I may be starting to agree with @laserson on #608.

> On the other hand, there is currently an outstanding Parquet issue to investigate why the filter2 API is slower than the original (https://issues.apache.org/jira/browse/PARQUET-98). It would be good for that to be resolved before switching.

[...]

PR #616, opened 2015-03-12
[ADAM-565] Upgrade to Parquet filter2 API.

tomwhite:
See discussion on https://github.com/bigdatagenomics/adam/issues/565

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
@@ -31,7 +31,6 @@ import org.bdgenomics.adam.converters._
 import org.bdgenomics.adam.instrumentation.Timers._
 import org.bdgenomics.adam.io._
 import org.bdgenomics.adam.models._
-import org.bdgenomics.adam.predicates.ADAMPredicate
 import org.bdgenomics.adam.projections.{ AlignmentRecordField, NucleotideContigFragmentField, Projection }
 import org.bdgenomics.adam.rdd.contig.NucleotideContigFragmentRDDFunctions
 import org.bdgenomics.adam.rdd.features._
@@ -44,7 +43,7 @@ import org.bdgenomics.utils.instrumentation.Metrics
 import org.seqdoop.hadoop_bam.util.SAMHeaderReader
 import org.seqdoop.hadoop_bam._
 import parquet.avro.{ AvroParquetInputFormat, AvroReadSupport }
-import parquet.filter.UnboundRecordFilter
+import parquet.filter2.predicate.FilterPredicate
 import parquet.hadoop.ParquetInputFormat
 import parquet.hadoop.util.ContextUtil
 import scala.collection.JavaConversions._
@@ -126,7 +125,7 @@ class ADAMContext(val sc: SparkContext) extends Serializable with Logging {
    * @tparam T The type of records to return
    * @return An RDD with records of the specified type
    */
-  private[rdd] def adamLoad[T, U <: UnboundRecordFilter](filePath: String, predicate: Option[Class[U]] = None, projection: Option[Schema] = None)(implicit ev1: T => SpecificRecord, ev2: Manifest[T]): RDD[T] = {
+  private[rdd] def adamLoad[T](filePath: String, predicate: Option[FilterPredicate] = None, projection: Option[Schema] = None)(implicit ev1: T => SpecificRecord, ev2: Manifest[T]): RDD[T] = {
     //make sure a type was specified
     //not using require as to make the message clearer
     if (manifest[T] == manifest[scala.Nothing])
@@ -139,7 +138,7 @@ class ADAMContext(val sc: SparkContext) extends Serializable with Logging {
 
     if (predicate.isDefined) {
       log.info("Using the specified push-down predicate")
-      ParquetInputFormat.setUnboundRecordFilter(job, predicate.get)
+      ParquetInputFormat.setFilterPredicate(job.getConfiguration, predicate.get)
     }
 
     if (projection.isDefined) {
@@ -208,7 +207,7 @@ class ADAMContext(val sc: SparkContext) extends Serializable with Logging {
         throw new IllegalArgumentException("If you're reading a BAM/SAM file, the record type must be Read")
 
     } else {
-      val projected: RDD[T] = adamLoad[T, UnboundRecordFilter](filePath, None, projection = Some(projection))
+      val projected: RDD[T] = adamLoad[T](filePath, None, projection = Some(projection))
 
       val recs: RDD[SequenceRecord] =
         if (isADAMRecord) {
@@ -227,12 +226,8 @@ class ADAMContext(val sc: SparkContext) extends Serializable with Logging {
     }
   }
 
-  private def applyPredicate[T, U <: ADAMPredicate[T]](reads: RDD[T], predicateOpt: Option[Class[U]])(implicit ev1: T => SpecificRecord, ev2: Manifest[T]): RDD[T] =
-    predicateOpt.map(_.newInstance()(reads)).getOrElse(reads)
-
-  def loadBam[U <: ADAMPredicate[AlignmentRecord]](
-    filePath: String,
-    predicate: Option[Class[U]] = None): RDD[AlignmentRecord] = {
+  def loadBam(
+    filePath: String): RDD[AlignmentRecord] = {
 
     // We need to separately read the header, so that we can inject the sequence dictionary
     // data into each individual Read (see the argument to samRecordConverter.convert,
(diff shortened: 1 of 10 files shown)


massie:
Jenkins, add to whitelist and test please.

Really appreciate the contribution, @tomwhite !

adam-core/src/test/scala/org/bdgenomics/adam/util/ADAMFunSuite.scala
```diff
@@ -25,6 +25,7 @@ trait ADAMFunSuite extends SparkFunSuite {
   override val properties: Map[String, String] = Map(("spark.serializer", "org.apache.spark.serializer.KryoSerializer"),
     ("spark.kryo.registrator", "org.bdgenomics.adam.serialization.ADAMKryoRegistrator"),
     ("spark.kryoserializer.buffer.mb", "4"),
-    ("spark.kryo.referenceTracking", "true"))
+    ("spark.kryo.referenceTracking", "true"),
+    ("spark.hadoop.parquet.task.side.metadata", "false"))
```
massie: We can't use `parquet.task.side.metadata` with the `filter2` API?  Or just better to not use it in general?

massie:
This PR looks good to me. Not only will be get a performance bump but the code is cleaner and easier to reason about. Thanks, Tom.

adam-core/src/test/scala/org/bdgenomics/adam/util/ADAMFunSuite.scala
tomwhite: Filtering will still work, it's just that row group elimination only occurs on the client side. We could remove this change (leave parquet.task.side.metadata set to true) and the test will still pass.

massie: In `ParquetInputFormat.java`, there is the following comment,

```java
/**
   * key to turn on or off task side metadata loading (default true)
   * if true then metadata is read on the task side and some tasks may finish immediately.
   * if false metadata is read on the client which is slower if there is a lot of metadata but tasks will only be spawn if there is work to do.
   */
  public static final String TASK_SIDE_METADATA = "parquet.task.side.metadata";
```

which I read to mean that setting this to `false` would mean that metadata is read on the client side instead of the task side. I thought that "task side" is the opposite of "client side" but I could very well be misunderstanding this too.

tomwhite: Task side is on the cluster nodes, client side is where the Spark client is running. The reason that task side is the default now for reading metadata is to avoid situations where a large number (thousands) of Parquet files are being processed and the client has to read all the footers before the job can start. See https://issues.apache.org/jira/browse/PARQUET-139

Row group filtering will occur in both cases (client side and task side) - I was wrong to say above that it only occurs on the client side. So I've made a minor adjustment to the code to remove this change, and keep the Parquet default of task side row group elimination. I used the debugger to confirm that row group elimination was still happening in the test.

massie:

fnothaft:
I am +1 to merge, _preconditioned on one change_.

I think we should prefer to remove predicate pushdown from the automagical loaders instead of warning that the predicate pushdown is ignored. I believe that @laserson would favor this (doing this would resolve #609). My reasoning is that ignoring a projection is OK, because you only pay a performance penalty; getting "more" fields back doesn't change the logic of your query, it just hurts your performance. Ignoring a predicate should be verboten, because you'll get a different set of records back than if the predicate had been applied.

I'll comment in the diff to indicate where these changes would go.

fnothaft:
I think this looks great, BTW. I'm stoked to move over to the new filter2 API.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -378,132 +369,131 @@ class ADAMContext(val sc: SparkContext) extends Serializable with Logging {
...
     projection: Option[Schema] = None): RDD[DatabaseVariantAnnotation] = {
-    adamLoad[DatabaseVariantAnnotation, U](filePath, predicate, projection)
+    adamLoad[DatabaseVariantAnnotation](filePath, predicate, projection)
   }
 
-  def loadVariantAnnotations[U <: ADAMPredicate[DatabaseVariantAnnotation]](
+  def loadVariantAnnotations(
     filePath: String,
-    predicate: Option[Class[U]] = None,
+    predicate: Option[FilterPredicate] = None,
```
fnothaft: Suggest removal.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -378,132 +369,131 @@ class ADAMContext(val sc: SparkContext) extends Serializable with Logging {
...
-      loadParquetVariantAnnotations[U](filePath, predicate, projection)
+      loadParquetVariantAnnotations(filePath, predicate, projection)
     }
   }
 
-  def loadFeatures[U <: ADAMPredicate[Feature]](
+  def loadFeatures(
     filePath: String,
-    predicate: Option[Class[U]] = None,
+    predicate: Option[FilterPredicate] = None,
```
fnothaft: Suggest removal.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -378,132 +369,131 @@ class ADAMContext(val sc: SparkContext) extends Serializable with Logging {
...
-      loadParquetFeatures[U](filePath, predicate, projection)
+      loadParquetFeatures(filePath, predicate, projection)
     }
   }
 
-  def loadGenes[U <: ADAMPredicate[Feature]](filePath: String,
-                                             predicate: Option[Class[U]] = None,
-                                             projection: Option[Schema] = None): RDD[Gene] = {
+  def loadGenes(filePath: String,
+                predicate: Option[FilterPredicate] = None,
```
fnothaft: Suggest removal.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -378,132 +369,131 @@ class ADAMContext(val sc: SparkContext) extends Serializable with Logging {
...
+                projection: Option[Schema] = None): RDD[Gene] = {
     import ADAMContext._
     loadFeatures(filePath, predicate, projection).asGenes()
   }
 
-  def loadSequence[U <: ADAMPredicate[NucleotideContigFragment]](
+  def loadSequence(
     filePath: String,
-    predicate: Option[Class[U]] = None,
+    predicate: Option[FilterPredicate] = None,
```
fnothaft: Suggest removal.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -378,132 +369,131 @@ class ADAMContext(val sc: SparkContext) extends Serializable with Logging {
...
-      loadParquetFragments[U](filePath, predicate, projection)
+      loadParquetFragments(filePath, predicate, projection)
     }
   }
 
-  def loadGenotypes[U <: ADAMPredicate[Genotype]](
+  def loadGenotypes(
     filePath: String,
-    predicate: Option[Class[U]] = None,
+    predicate: Option[FilterPredicate] = None,
```
fnothaft: Suggest removal.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -378,132 +369,131 @@ class ADAMContext(val sc: SparkContext) extends Serializable with Logging {
...
-      loadParquetGenotypes[U](filePath, predicate, projection)
+      loadParquetGenotypes(filePath, predicate, projection)
     }
   }
 
-  def loadVariants[U <: ADAMPredicate[Variant]](
+  def loadVariants(
     filePath: String,
-    predicate: Option[Class[U]] = None,
+    predicate: Option[FilterPredicate] = None,
```
fnothaft: Suggest removal.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -378,132 +369,131 @@ class ADAMContext(val sc: SparkContext) extends Serializable with Logging {
...
-      loadParquetVariants[U](filePath, predicate, projection)
+      loadParquetVariants(filePath, predicate, projection)
     }
   }
 
-  def loadAlignments[U <: ADAMPredicate[AlignmentRecord]](
+  def loadAlignments(
     filePath: String,
-    predicate: Option[Class[U]] = None,
+    predicate: Option[FilterPredicate] = None,
```
fnothaft: Suggest removal.

massie:
This is an issue with Jenkins that @shaneknapp is looking into.

shaneknapp:
jenkins, test this please

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
massie: Just to clarify. You want the `predicate` argument removed? If so, how will the predicate be available to the final `loadParquetVarientAnnotations` which can use predicate pushdown? Does it make more sense to fast fail here when there is `Some(predicate)` but the format is `vcf` instead of Parquet?

fnothaft: Yeah; this is the discussion that @laserson raised on #609. Uri makes good arguments on the ticket for that issue. I disagreed at first, but have come around to agree with Uri's arguments.

As an aside, if you are going to fail if a predicate is provided and the format isn't VCF, there's not really any benefit in allowing people to provide a predicate to the automagical function. Essentially, the automagical function just exists as a shim of convenience, so it should be difficult to make the automagical loaders throw an error.

massie: Looks like these methods are mostly used for the `Adam2*` functions anyway. Agree it makes sense to just remove predicates altogether since they aren't being used (or useful).

tomwhite:
Added a new commit with @fnothaft's suggested changes. Please let me know if I should squash the commits.

fnothaft:
@tomwhite I would probably squash the first three into the [ADAM-565] commit, and would relabel the last commit with [ADAM-609].

tomwhite:
Done.

fnothaft:
Thanks @tomwhite! I will merge once the new @AmplabJenkins run completes.

fnothaft:
Merged! Thanks @tomwhite!

tomwhite:
Thanks for your help @massie and @fnothaft!

laserson:
Beautiful!  Great work @tomwhite :)

PR #617, opened 2015-03-16
fix for gtfs without exonid

antonkulaga:
Some gtfs (esp. produced by cuffmerge) have no exonid that breaks asGenes method. That if why here I create artificial exondid from exon number


fnothaft:

fnothaft:
Thanks @antonkulaga! Manually committed via https://github.com/bigdatagenomics/adam/commit/0d5b4c4a942cbac30f3aa64c98685252c083ac1d.

PR #620, opened 2015-03-16
[ADAM-600] Adding RegionJoin trait.

### LINKED ISSUE #600: Consolidate join functionality via a trait
This was discussed in #517. Instead of layering the functionality via an additional `RegionRDDFunctions` layer, @laserson proposed layering this on top of the various region join implementations via a trait.

It might also make sense to implement some sort of cardinality estimation to select a join strategy, but that should probably be follow on work.

fnothaft:
Resolves #600. @tdanford @laserson thoughts about this?

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/RegionJoin.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/RegionJoin.scala
@@ -1,0 +1,43 @@
+/**
+ * Licensed to Big Data Genomics (BDG) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The BDG licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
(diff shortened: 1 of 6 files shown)


adam-core/src/main/scala/org/bdgenomics/adam/rdd/RegionJoin.scala
```diff
@@ -0,0 +1,44 @@
...
+import org.apache.spark.SparkContext._
+import scala.Predef._
+import org.apache.spark.SparkContext
+import scala.reflect.ClassTag
+
+trait RegionJoin {
+  /**
+   * Performs a region join between two RDDs.
+   *
+   * @param sc A SparkContext for the cluster that will perform the join
```
laserson: delete this?

laserson:
+1.  I'll wait for possible comments before merging

adam-core/src/main/scala/org/bdgenomics/adam/rdd/RegionJoin.scala
fnothaft: Thanks; fixing...

laserson:
Thanks @fnothaft!

PR #623, opened 2015-03-18
[ADAM-601] Consolidate ReferencePosition

### LINKED ISSUE #601: Consolidate ReferenceRegion, Position, etc.
We had discussed this a bit on another thread; this winds up being a consolidation of #511, #517, #592, Here's what I propose:

1. Eliminate `ReferencePosition`. A `ReferencePosition` is just a special case of a `ReferenceRegion` (region with width = 1). For the cost of 1 `Long` field, we could eliminate a lot of duplicated code.
2. Eliminate the `*WithOrientation` models. We have a `Strand` enum: https://github.com/bigdatagenomics/bdg-formats/blob/master/src/main/resources/avro/bdg.avdl#L720.
3. Move `ReferenceRegion` to Avro. This will allow us to consolidate fields in `AlignmentRecord`, `Variant`, and `NucleotideContigFragment`. I think we'd keep the `org.bdgenomics.adam.models.ReferenceRegion` object, and consolidate an `apply` method, as well as the `overlaps`, `isAdjacent`, etc methods there.
4. As proposed in #511, add an `Ordering` for `ReferenceRegion`.

fnothaft:
Resolves #601. Consolidates ReferencePosition so that it extends ReferenceRegion. Additionally, consolidates ordered classes down into unordered classes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ build
 *.log
 .*.swp
 .DS_Store
+
+*#*
--- a/adam-core/src/main/scala/org/bdgenomics/adam/models/ReferencePosition.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/models/ReferencePosition.scala
@@ -17,193 +17,43 @@
  */
 package org.bdgenomics.adam.models
 
-import org.bdgenomics.formats.avro._
-import org.bdgenomics.adam.rdd.ADAMContext._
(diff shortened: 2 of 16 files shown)


fnothaft:
Rebased.

fnothaft:
Rebased. Ping for review/merge.

fnothaft:
Merged via #624.

PR #624, opened 2015-03-18
Conceptual fix for duplicate marking and sorting stragglers

### LINKED ISSUE #574: Remove wrapping by Option where throwing an exception is preferable
This is pretty closely related to #297, #420, #421. There are a few places where we're wrapping by an option where we should really be throwing an exception if we would return `None`. The ones that stand out are:

* `ReferencePosition*` and `ReferenceRegion*`: If you provide an unmapped read/contig/genotype/X, you get a `None`.
* `RichAlignmentRecord`: Similar to the above...

Currently, I don't think we're optimizing for the _common case_. In my world, if you're asking for the position of an object, you have already checked to make sure that the object is mapped. If I'm doing variant calling, I'm going to filter out all unmapped reads first`*`. `ReferenceRegion`'s `apply(ar: AlignmentRecord)` returns an `Option[ReferenceRegion]`. This is bad in the common case, because I'm allocating an extra object (the `Option`), and then I have to unpack the `ReferenceRegion` from the option. _If_ the `Option` is a `None` anyways, `None.get` throws an exception _that has no useful debug info_.

This isn't to say that having a method that goes `AlignmentRecord => Option[ReferenceRegion]` isn't useful; IIRC, we need something like this for duplicate marking. My point is mostly that:

1. By default, we should assume that the common case is that the user is _providing good input_.
2. When the user provides _bad_ input, we should throw a detailed exception.

Evidently, I'm just in the mood to stir the pot tonight! I'd love to hear thoughts on this.

`*`: This assumes that I'm not going to do something like try to rescue unmapped reads. However, I believe that most unmapped read rescue algorithms rely on the mate read in the fragment being mapped, so the unmapped-ness of the read is irrelevant anyways.

fnothaft:
Resolves #574, depends on #623. In this pull request, we associate unmapped reads with a reference position that is derived from either their read name or their read sequence during the sort and duplicate marking phases. Specifically, sort uses `ZZZ<read_name>` and duplicate marking uses `<read_sequence>` as the contig name. Unmapped reads are placed at position 0 on the contig. Conceptually, this change should also enable duplicate marking on unmapped reads. I haven't tested this yet, but will spin a cluster up sometime soon.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ build
 *.log
 .*.swp
 .DS_Store
+
+*#*
--- a/adam-cli/src/main/scala/org/bdgenomics/adam/cli/CalculateDepth.scala
+++ b/adam-cli/src/main/scala/org/bdgenomics/adam/cli/CalculateDepth.scala
@@ -89,8 +89,8 @@ class CalculateDepth(protected val args: CalculateDepthArgs) extends ADAMSparkCo
     val variantNames = vcf.collect().toMap
 
     val joinedRDD: RDD[(ReferenceRegion, AlignmentRecord)] =
-      if (args.cartesian) BroadcastRegionJoin.cartesianFilter(variantPositions.keyBy(v => v), mappedRDD.keyBy(ReferenceRegion(_).get))
-      else BroadcastRegionJoin.partitionAndJoin(variantPositions.keyBy(v => v), mappedRDD.keyBy(ReferenceRegion(_).get))
+      if (args.cartesian) BroadcastRegionJoin.cartesianFilter(variantPositions.keyBy(v => v), mappedRDD.keyBy(ReferenceRegion(_)))
+      else BroadcastRegionJoin.partitionAndJoin(variantPositions.keyBy(v => v), mappedRDD.keyBy(ReferenceRegion(_)))
 
     val depths: RDD[(ReferenceRegion, Int)] =
       joinedRDD.map { case (region, record) => (region, 1) }.reduceByKey(_ + _).sortByKey()
--- a/adam-core/src/main/scala/org/bdgenomics/adam/models/ReferencePosition.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/models/ReferencePosition.scala
@@ -17,193 +17,43 @@
  */
 package org.bdgenomics.adam.models
 
-import org.bdgenomics.formats.avro._
-import org.bdgenomics.adam.rdd.ADAMContext._
 import com.esotericsoftware.kryo.{ Kryo, Serializer }
 import com.esotericsoftware.kryo.io.{ Input, Output }
-import scala.annotation.tailrec
+import org.bdgenomics.formats.avro._
 
-object ReferencePositionWithOrientation {
-
-  def apply(record: AlignmentRecord): Option[ReferencePositionWithOrientation] = {
-    if (record.getReadMapped) {
-      Some(new ReferencePositionWithOrientation(ReferencePosition(record).get, record.getReadNegativeStrand))
-    } else {
-      None
-    }
-  }
-
-  def fivePrime(record: AlignmentRecord): Option[ReferencePositionWithOrientation] = {
-    if (record.getReadMapped) {
-      Some(new ReferencePositionWithOrientation(ReferencePosition.fivePrime(record).get, record.getReadNegativeStrand))
-    } else {
-      None
-    }
-  }
-  /**
-   * Given an index into a sequence (e.g., a transcript), and a sequence
-   * of regions that map this sequence to a reference genome, returns a
-   * reference position.
-   *
-   * @param idx Index into a sequence.
-   * @param mapping Sequence of reference regions that maps the sequence to a
-   * reference genome.
-   * @return Returns a lifted over reference position.
-   *
-   * @throws AssertionError Throws an assertion error if the strandedness does
-   * not match between the reference position and the mapping regions.
-   * @throws IllegalArgumentException Throws an exception if the index given
-   * is out of the range of the mappings given.
-   */
-  def liftOverToReference(idx: Long,
-                          mapping: Seq[ReferenceRegionWithOrientation]): ReferencePositionWithOrientation = {
-    val negativeStrand = mapping.head.negativeStrand
-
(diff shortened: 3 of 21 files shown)


hannes-ucsc:
Getting this when running against NA12878.hiseq.wgs.bwa.raw.bam on 98 slaves:
```
java.lang.AssertionError: assertion failed
	at scala.Predef$.assert(Predef.scala:165)
	at org.bdgenomics.adam.models.ReferenceRegion.<init>(ReferenceRegion.scala:127)
	at org.bdgenomics.adam.models.ReferencePosition.<init>(ReferencePosition.scala:86)
	at org.bdgenomics.adam.models.ReferencePosition$.apply(ReferencePosition.scala:82)
	at org.bdgenomics.adam.rich.RichAlignmentRecord.fivePrimeReferencePosition(RichAlignmentRecord.scala:134)
	at org.bdgenomics.adam.models.ReferencePositionPair$$anonfun$apply$1.org$bdgenomics$adam$models$ReferencePositionPair$$anonfun$$getPos$1(ReferencePositionPair.scala:38)
	at org.bdgenomics.adam.models.ReferencePositionPair$$anonfun$apply$1$$anonfun$apply$2.apply(ReferencePositionPair.scala:45)
	at org.bdgenomics.adam.models.ReferencePositionPair$$anonfun$apply$1$$anonfun$apply$2.apply(ReferencePositionPair.scala:45)
	at scala.Option.map(Option.scala:145)
	at org.bdgenomics.adam.models.ReferencePositionPair$$anonfun$apply$1.apply(ReferencePositionPair.scala:45)
	at org.bdgenomics.adam.models.ReferencePositionPair$$anonfun$apply$1.apply(ReferencePositionPair.scala:30)
	at org.apache.spark.rdd.Timer.time(Timer.scala:57)
	at org.bdgenomics.adam.models.ReferencePositionPair$.apply(ReferencePositionPair.scala:30)
	at org.bdgenomics.adam.rdd.read.MarkDuplicates$$anonfun$apply$1.apply(MarkDuplicates.scala:79)
	at org.bdgenomics.adam.rdd.read.MarkDuplicates$$anonfun$apply$1.apply(MarkDuplicates.scala:79)
	at org.apache.spark.rdd.RDD$$anonfun$keyBy$1.apply(RDD.scala:1227)
	at org.apache.spark.rdd.RDD$$anonfun$keyBy$1.apply(RDD.scala:1227)
	at scala.collection.Iterator$$anon$11.next(Iterator.scala:328)
	at scala.collection.Iterator$$anon$11.next(Iterator.scala:328)
```

fnothaft:
@hannes-ucsc looking now.

hannes-ucsc:
Confirmed working on CCLE exome. 

10 r3.xlarge, s.d.p 40

Stage 2 task duration: average 40.75, stdev 5.74, median 41

fnothaft:
@hannes-ucsc _Excellent_! Any chance you'd know what the variance is without the patch?

hannes-ucsc:
this PR, Stage 2 task duration in min: 
Median        6
Average        5.9875
StdDev        0.20778132
Variance        0.04317307692

0.16.0, Stage 2 task duration in min: 
Median        5.6
Average        5.67
StdDev        0.2523733498
Variance        0.06369230769

hannes-ucsc:
I don't think this is conclusive. I mainly ran this to catch bugs. Will run WGS next week.

BTW: I used this BAM: https://browser.cghub.ucsc.edu/details/ebdb53ae-6386-4bc4-90b1-4f249ff9fcdf/

fnothaft:
Yeah, this doesn't appear to be demonstrating stragglers. Let's wait for a WGS run.

hannes-ucsc:
I'm happy to report that the WGS finished successfully. The per-stage running time in min was 32, 51, 34 respectively. All tasks in the last stage (the one that always had the two stragglers) were within a minute of each other.

So this is great news since it marks the first successful WGS run for us. Can I just say one thing: the "ZZZ" smells bad to me. Is there a way that instead of sorting by the tuple ```( library, position )```  we can sort by the triple ```( is_mapped, if is_mapped library else read_name, if is_mapped position else 0 )```?

fnothaft:
@hannes-ucsc w00t! Great to hear. There is a similar straggler problem in INDEL realignment; I'll have a PR for that upcoming (possibly by the end of the week? not sure).

> Can I just say one thing: the "ZZZ" smells bad to me. Is there a way that instead of sorting by the tuple ( library, position ) we can sort by the triple ( is_mapped, if is_mapped library else read_name, if is_mapped position else 0 )?

I agree that the `ZZZ` is a smell; I think your approach is reasonable. Let me see if there's a clean way to refactor it.

massie:
As I understand it, the Spark `Partitioner` uses the `hashCode()` method for assigning partitions.

In Avro, the `hashCode()` method is in `GenericData.java` and reads...

```java
/** Compute a hash code according to a schema, consistent with {@link
   * #compare(Object,Object,Schema)}. */
  public int hashCode(Object o, Schema s) {
    if (o == null) return 0;                      // incomplete datum
    int hashCode = 1;
    switch (s.getType()) {
    case RECORD:
      for (Field f : s.getFields()) {
        if (f.order() == Field.Order.IGNORE)
          continue;
        hashCode = hashCodeAdd(hashCode,
                               getField(o, f.name(), f.pos()), f.schema());
      }
      return hashCode;
    case ARRAY:
      Collection<?> a = (Collection<?>)o;
      Schema elementType = s.getElementType();
      for (Object e : a)
        hashCode = hashCodeAdd(hashCode, e, elementType);
      return hashCode;
    case UNION:
      return hashCode(o, s.getTypes().get(resolveUnion(s, o)));
    case ENUM:
      return s.getEnumOrdinal(o.toString());
    case NULL:
      return 0;
    case STRING:
      return (o instanceof Utf8 ? o : new Utf8(o.toString())).hashCode();
    default:
      return o.hashCode();
    }
  }
```

If we just use an `AlignmentRecord` with all fields NULL except the reference, position and map flag then we should be ok. The Avro Record also support `Comparable` for sorting too.

fnothaft:
Rebased. Ping for review/merge of this and #623.

massie:
Thanks, Frank!

PR #628, opened 2015-03-22
[ADAM-627] Makes ReferenceFile trait extend Serializable.

### LINKED ISSUE #627: ReferenceFiles should extend Serializable
Extends #618. If a `ReferenceFile` is going to be used as a broadcast variable (which is the main use case for a `ReferenceFile`) it needs to extend `Serializable`.

fnothaft:
Resolves #627.


fnothaft:
Rebased.

fnothaft:
Rebased.

massie:
Thanks, Frank!

PR #630, opened 2015-03-26
Add flatten command.

tomwhite:
This adds a 'flatten' command to the CLI for removing nesting from ADAM files in Parquet format. This is useful for querying ADAM data using tools like Impala, which can't currently read nested fields.


fnothaft:
Looks great @tomwhite ! Can you just run `format-source` and push an update?

tomwhite:
That seemed to fix it. Thanks for taking a look @fnothaft

fnothaft:
Merged! Thanks @tomwhite!

PR #639, opened 2015-04-02
[ADAM-636] Removed assert in annotation loader and replaced with correct wiring

### LINKED ISSUE #636: Clean up connections in VCF annotation loader
There's a TODO in the VCF annotation loader that causes an assert to fire (https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantContextConverter.scala#L181). It looks like there's just a small bit of code that needs to be hooked up here to fix the TODO.

fnothaft:
Resolves #636. Still needs a bit more testing; I'm going to see if I can track down the person who reported the bug to get their input dataset.


massie:
Thanks, Frank!

PR #650, opened 2015-04-14
[ADAM-649] Support transform of all BAM/SAM files in a directory.

tomwhite:


fnothaft:
Merged! Thanks @tomwhite.

Issue #410, opened 2014-10-09
Replace hard-coded sequence utils with an Alphabet model


tdanford:
As part of the work on Gene models, we introduced the SequenceUtils object (in adam.utils), which is just a hacky container for some simple "complement" operations on DNA sequences themselves.  

This should be replaced by an actual Alphabet model, with support for ambiguity codes, different alphabets, etc.

Agent007:
Has anyone taken a look at https://github.com/bioscala/bioscala/tree/master/src/main/scala/bio/sequence ?

fnothaft:
@Agent007 nice! Thanks for the heads up there; I wasn't aware of that, but it's definitely worth a look.

fnothaft:
Closed by #653.

PR #653, opened 2015-04-19
ADAM issue #410

bryanjj:
I was interested in contributing to this project and learning scala.
I am attempting to partially address https://github.com/bigdatagenomics/adam/issues/410

I created a new alphabet class which operates on symbols.  I based the functionality on the existing SequenceUtils class which supports the following functionality:
1. basic DNA symbols
2. case-insensitivity
3. reverse complement function which leaves unknown chars unchanged

My main concerns with this pull request are
1. I'm just learning scala so might be doing something dumb
2. I'm not sure which package this class should go in
3. I think performance should be ok because the String map function uses a StringBuilder internally, but not sure about overhead on case class creation or exception handling for unknown characters
4. This issue also mentions "ambiguity codes", but I'm not sure what they are but should be able to add later if they're needed

this is my first github contribution ever so let me know what you think


massie:

adam-core/src/main/scala/org/bdgenomics/adam/models/Alphabet.scala
```diff
@@ -0,0 +1,106 @@
...
+  /** the symbols in this alphabet */
+  val symbols: Seq[Symbol]
+
+  /**
+   * flag if symbols are case-sensitive.
+   * if true, this alphabet will treat symbols representing upper and lower case symbols as distinct
+   * if false, this alphabet will treat upper or lower case chars as the same for its symbols
+   *
+   */
+  val caseSensitive: Boolean
```
fnothaft: +1

adam-core/src/main/scala/org/bdgenomics/adam/models/Alphabet.scala
```diff
@@ -0,0 +1,106 @@
...
+
+  /** map of the symbol char to the symbol */
+  lazy val symbolMap: Map[Char, Symbol] =
+    if (caseSensitive)
+      symbols.map(x => x.label -> x).toMap
+    else
+      symbols.flatMap(x => Seq(x.label.toLower -> x.toLower, x.label.toUpper -> x.toUpper)).toMap
```
fnothaft: My understanding of the `symbolMap` is that it'll canonicalize an input symbol. In that case, shouldn't the `flatMap`ped Seq be `Seq(x.label.toLower -> x.toUpper, x.label.toUpper -> x.toUpper)`?

adam-core/src/main/scala/org/bdgenomics/adam/models/Alphabet.scala
```diff
@@ -0,0 +1,106 @@
...
+
+  /**
+   *
+   * @param s Each char in this string represents a symbol on the alphabet.
+   *          If the char is not in the alphabet then the unknown char is treated as the complement
+   * @return the reversed complement of the given string.
+   */
+  def reverseComplement(s: String): String = {
+    s.map(x => Try(apply(x)).getOrElse(Symbol(x, x)).complement).reverse
```
fnothaft: The `Try` here is to check if char `x` is in your alphabet, correct? If you get an `x` that isn't in the alphabet as input, wouldn't you want to throw an exception? I agree with wrapping in some form of try/catch here, but I would think it to be better to use the try/catch to throw a better exception than the `KeyNotFound` exception you'd get back from the symbol map (e.g., convert the `KeyNotFound` into `throw new IllegalArgumentException("Character %s not found in alphabet.".format(x))`). At least, that's my general thought.

fnothaft:
+1, this is good code. Thanks for the contribution @bryanjj! I left a few comments inline, but I think we can merge this soon. As a small nit, can you [squash the commits down](http://makandracards.com/makandra/527-squash-several-git-commits-into-a-single-commit) into a single commit, and prefix that commit with `[ADAM-410]`?

adam-core/src/main/scala/org/bdgenomics/adam/models/Alphabet.scala
bryanjj: Yes I think that makes sense.  The reason I did it this way was to match the existing behavior in the SequenceUtils match statement, where case "a" => "t" and case "A" => "T" but not sure if that is a real requirement

adam-core/src/main/scala/org/bdgenomics/adam/models/Alphabet.scala
bryanjj: Ok.  so the Try in this case will catch the KeyNotFoundException and replace it with the Symbol(x,x).  This is to maintain the existing behavior in the SequenceUtils class where an unrecognized character is just kept as is.  The reverseComplementExact will throw the KeyNotFound but I can convert that to an IllegalArgument instead

bryanjj:
Thanks for the comments!  I will make the changes and see how it looks

bryanjj:
I addressed the comments.  Please let me know what you think.

After the squash, the build succeeded for hadoop 1.0 and 2.2 but failed for 2.3  I'm not sure if it's related to my change or something random.

fnothaft:
Jenkins, retest this please.

fnothaft:
@laserson @tdanford any thoughts on this? I would like to merge by Monday for the 0.17.0 release.

fnothaft:
Thanks @bryanjj! I've rebased and manually merged via https://github.com/bigdatagenomics/adam/commit/6ce4ef7c3c23580b1f083642a5c679ff86f59da5.

PR #654, opened 2015-04-19
Fix imports for RealignIndels to allow instrumentation to work

nfergu:
This removes an import of org.apache.spark.SparkContext._ which was causing indel realignment not to be instrumented, and replaces it with an import of org.apache.spark.rdd.MetricsContext._

Note that enabling instrumentation for indel realignment exposes a bug in BDG utils which leads to errors with negative durations. This bug is fixed in this PR: https://github.com/bigdatagenomics/utils/pull/35 but we will need to merge that and do another release of BDG utils before instrumentation will work properly for indel realignment.

I've also removed a couple of stray unused imports I added in a previous commit.


fnothaft:
Thanks @nfergu! I've manually rebased this and committed it as https://github.com/bigdatagenomics/adam/commit/95a0952e68b8524017993b1991a9aeaaef265ade.

PR #664, opened 2015-04-30
Move to org.bdgenomics.utils 0.2.0 release

fnothaft:
Resolves #536, #581, #603, #657. In anticipation of the 0.17.0 release, this moves to the org.bdgenomics.utils 0.2.0 release. This leads to:

* Unblocks #654.
* Moves from the `org.bdgenomics.utils.parquet.io` package to the `org.bdgenomics.utils.io` package
* Eliminates the CLI classes that had already moved to utils.
* Adds the _infrastructure_ for building Scala 2.10 and 2.11 artifacts. This includes infrastructure for _releasing_ both artifacts, so the ADAM 0.17.0 release will have both 2.10 and 2.11 artifacts! Yay!

PR #670, opened 2015-05-03
kill Job arg to ADAMSparkCommand

### LINKED ISSUE #662: Kill `job` argument to ADAMSparkCommand.run?
[The second arg here](https://github.com/bigdatagenomics/adam/blob/492fb3b04fc1ebe279e8aaa95f7b290cef911f27/adam-cli/src/main/scala/org/bdgenomics/adam/cli/ADAMCommand.scala#L46), `job: Job`, is never used in any ADAM command afaict.

Some tests seem to pass a value in, as if some dependency-injection is happening, but I can't see how that value would ever be used in the current world.

Am I missing something?

ryan-williams:
Working plan is that this will not be merged as these changes will make their way in to #664 via [fnothaft/adam#5](https://github.com/fnothaft/adam/pull/5), but since I made this anyway to try to do some testing without biting off the entire `utils-0.2.1` upgrade, here it is for now.

fixes #662


fnothaft:
Closing as this change merged in via #678.

Issue #657, opened 2015-04-26
Move from utils-parquet to utils-io.


fnothaft:
Since https://github.com/bigdatagenomics/utils/pull/33 and #645 have merged, we should cut over from `org.bdgenomics.utils.parquet` to `org.bdgenomics.utils.io`.

tomwhite:
@fnothaft I think we need a release of bdg-utils (0.1.3) to do this, don't we?

fnothaft:
Ah! Yes, good call. I'm a bit swamped today and tomorrow but will cut utils/0.1.3 on Wednesday.

fnothaft:
I just cut utils/0.2.0, which will hit in Maven Central tomorrow. I'll open a PR tomorrow AM to fix this and unblock #654.

fnothaft:
This was closed by #678.

Issue #634, opened 2015-03-27
Remove bdg-utils-parquet


tomwhite:
Most of the Parquet code in bdg-utils-parquet is unused as far as I can tell, so we should remove it to remove confusion about what it's for. If changes are needed to Parquet then I would encourage them to be made in the Parquet project itself. However, I don't think any Parquet changes are actually required at this point (please correct me if I'm wrong).

The two uses I found in ADAM are ParquetLister and the org.bdgenomics.utils.parquet.io package.

ParquetLister is only used from two CLI tests, so I would suggest simplifying it and moving it to the test package (here's a commit showing how: https://github.com/bigdatagenomics/adam/commit/a78ac65a9ff7b6ee8b4bc046930a3196a67b21ee). I would expect production code to use a ParquetReader or ParquetInputFormat directly.

The org.bdgenomics.utils.parquet.io package doesn't actually contain any Parquet code. ByteAccess is a general purpose IO class used by TwoBitFile in ADAM. Byte access and its associated classes could either move into org.bdgenomics.adam.io in ADAM (I know it was moved from there a few months ago), or into a new bdg-utils-io module in a org.bdgenomics.utils.io package. Any thoughts on this?

Once these changes have been made bdg-utils-parquet can be removed.

fnothaft:
@tomwhite I'm going to close this, as I believe we've purged this. Please reopen if I am in error.

Issue #603, opened 2015-03-01
Move command structure to bdg-utils?


fnothaft:
The command infrastructure is useful for much more than ADAM, and isn't really specific to genomics. For example, I myself use it in avocado, PacMin, and a few non-genomics projects. I'd like to move it to bdg-utils. At the same point in time, I think we could roll the build information work (see [here](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/util/BuildInformation.scala) and [here](https://github.com/bigdatagenomics/adam/blob/master/adam-cli/src/main/scala/org/bdgenomics/adam/cli/BuildInformation.scala)) to the main command (e.g., as a command line argument, `-v`/`--version`).

This is similar to how we pulled `SparkFunSuite` out of ADAM. It's really useful code, but for projects where I don't need to bring `ADAM` in as a dependency (e.g., the non-omics ones referred to earlier), I'd really prefer not to pull it in.

tdanford:
As an alternative, move the command structure further _up_ (or down, depending on your orientation) the food chain.  I ended up replicating a subset of ``ADAMMain`` as a class called ``QCMain`` in the qc-metrics repository.  

Maybe all the support classes (e.g. ``ADAMCommandCompanion``) can go in bdg-utils, and all the actual Main classes and commands can go in downstream repos?

fnothaft:
@tdanford 

> Maybe all the support classes (e.g. ADAMCommandCompanion) can go in bdg-utils, and all the actual Main classes and commands can go in downstream repos?

That'd be my preference as well.

tdanford:
+1.

fnothaft:
@tdanford I've opened a PR upstream: https://github.com/bigdatagenomics/bdg-utils/pull/30. I'll open a PR against ADAM tomorrow showing what these changes would look like.

laserson:
+1 to factoring out the common stuff

fnothaft:
This was closed by #678.

Issue #581, opened 2015-02-14
Upgrade to newer bdg-utils library


fnothaft:
The current bdg-utils version we're pointing at is built for Spark 1.1.0. A Spark method signature changed from 1.1.0 to 1.2.0, which leads to errors with the metrics.

heuermh:
Can this issue be closed?  The bdg-utils dependency version is the most recent version 0.2.2 as of commit a70b4ca12299a9639275671cb20f7887c63ecb09.

fnothaft:
Yes, it can. Thanks for pinging @heuermh!

PR #678, opened 2015-05-26
upgrade to bdg utils 0.2.1

### LINKED ISSUE #536: scala 2.11
I see that Spark 1.20 supports scala 2.11, it would be great if adam would also move to it.

### LINKED ISSUE #662: Kill `job` argument to ADAMSparkCommand.run?

ryan-williams:
Forked from #664 

Fixes #662, #657, #634, #603, #581.

#664 claimed to fix #536 as well but it's not obvious to me that this does in fact support scala 2.11 so I'm not including that, though I'm interested to see whether Github is smart enough to parse that from this text.


laserson:
I didn't check this as carefully as Frank's original PR.  I'm guessing this is good to go on your end?  I'll just wait for the green light from jenkins.

ryan-williams:
Yea it's just his PR with two commits on it, one was a merge with one trivial conflict that I resolved, and one is a couple of unused import removals.

PR #611, opened 2015-03-05
[ADAM-610] Fix FASTQ input format split picking.

fnothaft:
Resolves #610. This is a fairly extensive rewrite of the FASTQ input format code. I started debugging said bug, and found a lot of cruft and repeated code in the FASTQ input formats. I don't _love_ the new code, but I think it is substantially cleaner and has the plus side of working. I'm going to revalidate it on the ERR000096 FASTQs tomorrow; I had a cluster up, but alas they were spot nodes and disappeared into the EC2 mist.

PR #551, opened 2015-01-22
[ADAM-386] Multiline FASTQ input

cket:
[ADAM-386] Adds support for multiline FASTQ input

PR #679, opened 2015-05-26
[ADAM-610] Revert multiline FASTQ input change

### LINKED ISSUE #610: FASTQ input format picks bad splits
Reported on the [mailing list](https://groups.google.com/forum/#!topic/adam-developers/CMzA8nH7FXg). This seems to be a regression that came in from https://github.com/bigdatagenomics/adam/commit/2052521fba00da4993276161c150bc6b812359c7.

fnothaft:
Resolves #610. Reverts #551. Adding support for multiline FASTQ input made it difficult to support single line FASTQ. I'd proposed a fix in #611, but it was very slow. I'm in favor of reverting the multiline FASTQ support, since most long read technologies are only emitting single line FASTQ (as far as I can tell).


adam-core/src/main/java/org/bdgenomics/adam/io/InterleavedFastqInputFormat.java
```diff
@@ -127,13 +123,15 @@ public InterleavedFastqRecordReader(Configuration conf, FileSplit split) throws
          */
         private void positionAtFirstRecord(FSDataInputStream stream) throws IOException {
             Text buffer = new Text();
-            int bytesRead = 0;
-            
+
+            if (true) { // (start > 0) // use start>0 to assume that files start with valid data
```
tdanford: Just out of curiosity, why is this wrapped in "if (true)"?  Is it just because you're reverting and "this is the way it used to be?"

fnothaft: Correct. I'm hoping to do more of a cleanup later (e.g., similar to the cleanup that I did in #611), but wanted to get this back to working first.

massie:
Thanks, Frank!

PR #689, opened 2015-06-01
adam2bam - command for converting from adam format right to regular BAM/SAM file

antonstamov:


ryan-williams:
quick drive-by comment: can't `transform` do this?

antonstamov:
It' based on transform, but it replace outputPath directory by part-* file
located in

пн, 1 июня 2015, 18:21, Ryan Williams <notifications@github.com>:

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Adam2Bam.scala
```diff
@@ -0,0 +1,66 @@
...
+  @Args4jOption(required = false, name = "-sort_fastq_output", usage = "Sets whether to sort the FASTQ output, if saving as FASTQ. False by default. Ignored if not saving as FASTQ.")
+  var sortFastqOutput: Boolean = false
+}
+
+class Adam2Bam(protected val args: Adam2BamArgs) extends BDGSparkCommand[Adam2BamArgs] with Logging {
+  val companion = Adam2Bam
+
+  def run(sc: SparkContext) {
+    if (sc.loadParquetAlignments(args.inputPath).adamSave(args)) {
+      //make regular BAM/SAM file (not hadopped)
```
ryan-williams: `s/hadopped/hadoop/` maybe?

ryan-williams:
cool, that makes sense @antonstamov.

One suggestion: what if we made this "replace directory with `part-r0000` file" functionality an option to [`adamSAMSave`](https://github.com/bigdatagenomics/adam/blob/33edc701ea09864948acbbe4776f5a2d7adbe956/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala#L125)?

I would certainly use this capability during longer `adam-shell` sessions, for example.

Such functionality would probably be useful beyond just saving SAMs/BAMs, too, e.g. for VCFs and various other file formats.

I'm not saying I think you need to do *all* of that here, we can do it later, but if you wanted to that would be great, and I'm going to file an issue for it either way.

ryan-williams:
Filed #692 ^^

fnothaft:
> One suggestion: what if we made this "replace directory with part-r0000 file" functionality an option to adamSAMSave?

That would be my preference, instead of adding a second command.

ryan-williams:
That shouldn't be too difficult; how does it sound, @antonstamov?

antonstamov:
please checkout last commit

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala
```diff
@@ -156,6 +158,16 @@ class AlignmentRecordRDDFunctions(rdd: RDD[AlignmentRecord])
           conf
         )
     }
+    if (asRegularFile) {
+      log.info("make regular BAM/SAM file (not hadoop)")
+      val conf = new Configuration()
+      val fs = FileSystem.get(conf)
+      val ouputParentDir = filePath.substring(0, filePath.lastIndexOf("/") + 1)
+      val tmpPath = ouputParentDir + "tmp" + System.currentTimeMillis().toString
+      fs.rename(new Path(filePath + "/part-r-00000"), new Path(tmpPath))
```
ryan-williams: I thought I saw an instance where the format of the `part*` filename has varied across Hadoop versions; anyone know if that's a concern?

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
```diff
@@ -80,6 +79,8 @@ class TransformArgs extends Args4jBase with ADAMSaveAnyArgs with ParquetArgs {
   var forceLoadIFastq: Boolean = false
   @Args4jOption(required = false, name = "-force_load_parquet", usage = "Forces Transform to load from Parquet.")
   var forceLoadParquet: Boolean = false
+  @Args4jOption(required = false, name = "-as_regular_file", usage = "Saves OUTPUT as regular file instead of using Hadoop OutputFormat")
```
ryan-williams: I might place a weak vote for `s/regular/single/g`, i.e. "asSingleFile"? I feel like there's an even better way to describe this out there; will think about it.

ryan-williams: "regular file" is fine if we can't come up with something else

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala
ryan-williams: A little digging implies that "old" (v1?) Hadoop wrote filenames like `part-00000`, cf. [this thread](http://hadoop-common.472056.n3.nabble.com/Is-it-always-called-part-00000-tp124567p124595.html) and [this Hadoop FAQ](https://wiki.apache.org/hadoop/FAQ#How_do_I_change_final_output_file_name_with_the_desired_name_rather_than_in_partitions_like_part-00000.2C_part-00001.3F).

[This SO](http://stackoverflow.com/a/10666874/544236) implies that we could end up with `part-m-00000` files if the output was from a map phase, which seems to never happen with Spark?

Since ADAM still supports Hadoop 1.0.4, we should maybe confirm that this does the right thing there and doesn't look for the wrong filename.

ryan-williams: There also seem to be ways to configure Hadoop to write files in arbitrary other formats; if it's possible/easy to get that from available config objects that would be nice, though not essential by any means.

ryan-williams:
Thanks for this update @antonstamov, much better! I left a few questions, lmk what you think.

fnothaft:
Can you rebase? Thanks!

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
laserson: +1 to "single", IMO

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
```diff
@@ -80,6 +79,8 @@ class TransformArgs extends Args4jBase with ADAMSaveAnyArgs with ParquetArgs {
...
+  var asRegularFile: Boolean = false
```
laserson: same here, etc.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala
```diff
@@ -66,14 +68,14 @@ class AlignmentRecordRDDFunctions(rdd: RDD[AlignmentRecord])
     rdd.filter(overlapsQuery)
   }
 
-  def maybeSaveBam(args: SaveArgs): Boolean = {
+  def maybeSaveBam(args: SaveArgs, asRegularFile: Boolean = false): Boolean = {
```
laserson: Can this instead take `args: ADAMSaveAnyArgs`?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala
```diff
@@ -93,7 +95,7 @@ class AlignmentRecordRDDFunctions(rdd: RDD[AlignmentRecord])
   }
 
   def adamSave(args: ADAMSaveAnyArgs) = {
-    maybeSaveBam(args) || maybeSaveFastq(args) || { rdd.adamParquetSave(args); true }
+    maybeSaveBam(args, asRegularFile = args.asRegularFile) || maybeSaveFastq(args) || { rdd.adamParquetSave(args); true }
```
laserson: With above suggestion, you can just pass `args`

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
laserson: This help message seems a bit inaccurate to me, because `transform` will continue to use an `OutputFormat`.  I'd reword it.

ryan-williams:
Another thing that occurred to me: can we force a `.coalesce(1)` before writing the "single file"? Or are we already guaranteed that it will just have one partition somehow? 

I just had a BAM get written out in many partitions yesterday by ADAM and was coalescing manually after that since I hasn't realized I needed that step previously; saving that step and doing the file rename will be a very nice improvement to this flow!

laserson:
+1 to @ryan-williams's point.  I don't see how this feature can work without coalescing.

antonstamov:
Can somebody make code review of last changes and merge this pull request?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala
```diff
@@ -188,6 +189,16 @@ class AlignmentRecordRDDFunctions(rdd: RDD[AlignmentRecord])
           conf
         )
     }
+    if (asSingleFile) {
+      log.info("make regular BAM/SAM file (not hadoop)")
```
ryan-williams: I don't think this is accurate... we can and will want to write single files on Hadoop as well as non-Hadoop, no?

Also `s/regular/single/`.

Maybe:
```
log.info(s"Writing single ${if (asSam) "SAM" else "BAM"} file (not Hadoop-style directory)")
```

antonstamov:
@ryan-williams, ok I fixed this message

ryan-williams:
Cool, thanks! I'm just finishing running a test of this functionality reading from and writing to HDFS…

ryan-williams:
OK this looks good. It did take me >1hr to write a 14GB bam file this way though, which is sort of a shame… I wonder if we wrote it to many `part-*` files, without the header on ones past the first, and then used hadoop `cat` to make them one file, if that would be faster?

Anyway that can be follow-on work, having this work slowly is a nice start.

I'll wait for @fnothaft or someone to chime in before merging.

Thanks @antonstamov

fnothaft:
LGTM! I will squash down and merge.

fnothaft:
Thanks @antonstamov! Manually committed as https://github.com/bigdatagenomics/adam/commit/dd9dc3aaf13b816d2c5c634eb15caddf3fc512c7.

antonstamov:
@ryan-williams , do you have any idea how to save BAM/SAM `"without the header on ones past the first"`?

ryan-williams:
@antonstamov good q; it seems like it would be possible to `mapPartitionsWithIndex` and set some state on the first partition, off the top of my head.

ryan-williams:
seems like we'd have to make changes to hadoop-bam to allow it to write partitions without headers though: https://github.com/HadoopGenomics/Hadoop-BAM/blob/7.1.0/src/main/java/org/seqdoop/hadoop_bam/KeyIgnoringAnySAMOutputFormat.java#L121-L123

antonstamov:
@ryan-williams yes, and instead of `mapPartitionsWithIndex` that I don't understand how to use for this case, I found solution with: https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/rdd/PartitionPruningRDD.scala

ryan-williams:
ADAM [currently uses `mapPartitionsWithIndex`](https://github.com/bigdatagenomics/adam/blob/a68a7e1ad3d970111d007b6a4011365e15e213a6/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala#L135) to set the SAM/BAM headers on each executor, which is relevant to the code paths you've been looking at.

Have you used a `PartitionPruningRDD` to write a SAM/BAM file with only a header on the first partition? If so, how? It seems to me that hadoop-bam's insistence on every partition having a header means we can't do it unless we change it there or reimplement that functionality.

antonstamov:
I create two `PartitionPruningRDD`s, one with head partition and second with tail partitions, and save them with `saveAsNewAPIHadoopFile`, but for tail I use modified output format with writeHeader_=false flag from https://github.com/HadoopGenomics/Hadoop-BAM/blob/master/src/main/java/org/seqdoop/hadoop_bam/KeyIgnoringAnySAMOutputFormat.java

antonstamov:
I tested this with SAM format, but I'm not sure that merging of `part-r-*` files will work with BAM files, cause of their compression. Anyway tommorrow I will test this and in case of success will make PR.

ryan-williams:
awesome, thanks for continuing to push on this!

Issue #651, opened 2015-04-15
Support Hive-style partitioning


tomwhite:
It's common to partition sequence data by locus. This change would make it possible to partition genotypes (alignments, etc) using a Hive-compatible directory structure like chr=M/pos=N (N is something like `floor(position/10^K)`).

Querying would then be more efficient since the SQL engine would only need to read files in the partitions of interest (typically one partition when doing a point query).

fnothaft:
Don't [these partitioners](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicPartitioners.scala) satisfy the requirements?

tomwhite:
Possibly, although I think there's more work to write out data in a Hive-compatible directory structure. Or have you done that kind of thing already?

fnothaft:
Ah, no, haven't done that!

tomwhite:
OK, I'll take a look to see how to do that.

tomwhite:
Here's an initial go that uses Kite to do partitioning:

https://github.com/bigdatagenomics/adam/compare/master...tomwhite:ADAM-651-hive-partitions

The idea is to add a partition command that takes a partition strategy file that defines which fields are used in constructing the partitions, and how the records map to partition values. For example, [this partition strategy](https://github.com/bigdatagenomics/adam/blob/df6355dfabed5c87f6076a06398deef30e11c129/adam-cli/src/test/resources/genotypes-partition-strategy.json) partitions by chromosome and bucketed position.

This is not quite ready yet, as it depends on a snapshot version of Kite for https://issues.cloudera.org/browse/CDK-986 and https://issues.cloudera.org/browse/CDK-988.

fnothaft:
Nice; looks good! I hadn't seen Kite before; I'll need to take more of a look.

laserson:
As discussed, let's add command line options that specify a few different baked in partition schemes, such as locus-partitioned or sample-partitioned.

tomwhite:
I updated my [branch](https://github.com/bigdatagenomics/adam/compare/master...tomwhite:ADAM-651-hive-partitions) to use a partitioner to ensure that each reducer only writes to a small number of Parquet files. This makes sure that they reducers don't run out of memory.

[...]

PR #690, opened 2015-06-01
upgrade to Spark 1.3.1

### LINKED ISSUE #665: upgrade to Spark 1.3.1
any reason not to?

ryan-williams:
seems smooth, but a couple old jackson deps needed excluding

fixes #665

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -19,7 +19,7 @@
   <properties>
     <java.version>1.7</java.version>
     <avro.version>1.7.6</avro.version>
-    <spark.version>1.2.0</spark.version>
+    <spark.version>1.3.1</spark.version>
     <parquet.version>1.8.1</parquet.version>
     <!-- Edit the following line to configure the Hadoop (HDFS) version. -->
     <hadoop.version>2.2.0</hadoop.version>
@@ -477,12 +477,6 @@
         <groupId>org.slf4j</groupId>
         <artifactId>slf4j-log4j12</artifactId>
         <version>1.7.5</version>
-      </dependency>
-      <dependency>
-        <groupId>com.amazonaws</groupId>
-        <artifactId>aws-java-sdk</artifactId>
-        <version>1.7.5</version>
-        <scope>provided</scope>
       </dependency>
       <dependency>
         <groupId>org.apache.httpcomponents</groupId>
--- a/adam-core/pom.xml
+++ b/adam-core/pom.xml
@@ -192,10 +192,6 @@
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>com.amazonaws</groupId>
-      <artifactId>aws-java-sdk</artifactId>
-    </dependency>
-    <dependency>
       <groupId>org.apache.httpcomponents</groupId>
       <artifactId>httpclient</artifactId>
     </dependency>
--- a/adam-core/src/main/java/org/bdgenomics/adam/io/SingleFastqInputFormat.java
+++ b/adam-core/src/main/java/org/bdgenomics/adam/io/SingleFastqInputFormat.java
@@ -26,12 +26,12 @@ import org.apache.hadoop.io.Text;
 import org.apache.hadoop.io.compress.CompressionCodec;
 import org.apache.hadoop.io.compress.CompressionCodecFactory;
 import org.apache.hadoop.mapreduce.InputSplit;
-import org.apache.hadoop.mapreduce.JobContext;
 import org.apache.hadoop.mapreduce.RecordReader;
 import org.apache.hadoop.mapreduce.TaskAttemptContext;
 import org.apache.hadoop.mapreduce.lib.input.FileInputFormat;
 import org.apache.hadoop.mapreduce.lib.input.FileSplit;
 import org.apache.hadoop.util.LineReader;
+import org.bdgenomics.adam.util.ADAMHadoopUtil;
 
 import java.io.EOFException;
 import java.io.IOException;
@@ -288,7 +288,7 @@ public class SingleFastqInputFormat extends FileInputFormat<Void,Text> {
     public RecordReader<Void, Text> createRecordReader(
             InputSplit genericSplit,
             TaskAttemptContext context) throws IOException, InterruptedException {
-        context.setStatus(genericSplit.toString());
+        ADAMHadoopUtil.setTaskAttemptContextStatus(context, genericSplit.toString());
 
         // cast as per example in TextInputFormat
         return new SingleFastqRecordReader(context.getConfiguration(), 
--- a/adam-core/src/main/scala/org/bdgenomics/adam/util/ADAMHadoopUtil.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/util/ADAMHadoopUtil.scala
@@ -1,0 +1,29 @@
+/**
+ * Licensed to Big Data Genomics (BDG) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The BDG licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.bdgenomics.adam.util
+
+import org.apache.hadoop.mapreduce.TaskAttemptContext
+
+object ADAMHadoopUtil {
+
+  def setTaskAttemptContextStatus(context: TaskAttemptContext, status: String): Unit = {
+    val method = context.getClass.getMethod("setStatus", classOf[java.lang.String])
+    method.invoke(context, status: java.lang.String)
+  }
+
+}


ryan-williams:
```[ERROR]     Non-resolvable parent POM: Could not find artifact org.bdgenomics.adam:adam-parent:pom:0.16.1-SNAPSHOT and 'parent.relativePath' points at wrong local POM @ line 4, column 13 -> [Help 2]```

some kind of POM error? Can't tell if it's my fault, doesn't seem like it is from what I can tell…

File: pom.xml [hunk 1]
```diff
@@ -473,6 +473,16 @@
         <artifactId>aws-java-sdk</artifactId>
```
fnothaft: Are we actually using this dependency right now? I don't think we are.

fnothaft: Might be preferable to just remove the dependency (as opposed to excluding things).

ryan-williams:
rebased

ryan-williams:
let me try removing that dep

ryan-williams:
removed the dep, and excluding it from being brought in by `utils-io`.

on second thought, that might not be the correct thing to do? does `utils-io` actually need that dep? maybe I should just more narrowly exclude the `jackson.core` that comes with `utils-io`'s AWS dep?

pom.xml
ryan-williams: side note: Github is doing a really unfortunate thing and showing your old comments next to my new diffs

ryan-williams:
I'm going to do what I said there and switch to just excluding the jackson things, not the aws-java-sdk, since utils-io does seem to use it.

ryan-williams:
In other news I can't get a break with jenkins today!



The only thing that looks like an error is:
```
[ERROR] Error formatting /home/jenkins/workspace/ADAM-prb/SCALAVER/2.10/hadoop.version/1.0.4/label/centos/adam-cli/src/main/scala/org/bdgenomics/adam/cli/PrintADAM.scala: java.nio.charset.MalformedInputException: Input length = 1
```

[Link](https://amplab.cs.berkeley.edu/jenkins/job/ADAM-prb/729/SCALAVER=2.10,hadoop.version=1.0.4,label=centos/console).

I've seen that a bunch of times today on different PRs, no idea what it means.

ryan-williams:
OK, I think this is a better version!

fnothaft:
@ryan-williams 

> I'm going to do what I said there and switch to just excluding the jackson things, not the aws-java-sdk, since utils-io does seem to use it.

That sounds reasonable here. I've opened https://github.com/bigdatagenomics/utils/issues/41to remove the AWS SDK from `utils-io`. We do need the dependency right now, but I believe the piece of code that we need it for is dead.

fnothaft:
> OK, I think this is a better version!

+1, agreed!

ryan-williams:
If the code that uses it in `utils-io` is never exercised (which is what I interpreted "dead code" to mean), then the coarser exclusion might have "worked" here, but yea, this seems safer/better.

fnothaft:
@ryan-williams yeah, that's what I was meaning. The `utils-io` code that depends on the AWS SDK was part of the old S3 Parquet loader, which doesn't exist anymore. I agree though that this approach is safer.

ryan-williams:
[Jenkins](https://amplab.cs.berkeley.edu/jenkins/job/ADAM-prb/732/SCALAVER=2.10,hadoop.version=1.0.4,label=centos/console)!



What does it mean?

fnothaft:
Jenkins, retest this please.

fnothaft:
@ryan-williams I don't think the error you're seeing in:

> Jenkins!
>
> What does it mean?

Is the error. I'm seeing:

```
2015-06-01 17:59:57 WARN  LoadSnappy:46 - Snappy native library not loaded
2015-06-01 17:59:57 ERROR Executor:96 - Exception in task 0.0 in stage 0.0 (TID 0)
java.lang.IncompatibleClassChangeError: Found class org.apache.hadoop.mapreduce.TaskAttemptContext, but interface was expected
	at org.apache.spark.mapred.SparkHadoopMapRedUtil$.commitTask(SparkHadoopMapRedUtil.scala:95)
	at org.apache.spark.SparkHadoopWriter.commit(SparkHadoopWriter.scala:106)
	at org.apache.spark.rdd.PairRDDFunctions$$anonfun$13.apply(PairRDDFunctions.scala:1082)
	at org.apache.spark.rdd.PairRDDFunctions$$anonfun$13.apply(PairRDDFunctions.scala:1059)
	at org.apache.spark.scheduler.ResultTask.runTask(ResultTask.scala:61)
	at org.apache.spark.scheduler.Task.run(Task.scala:64)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:203)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
	at java.lang.Thread.run(Thread.java:745)
2015-06-01 17:59:57 ERROR SparkUncaughtExceptionHandler:96 - Uncaught exception in thread Thread[Executor task launch worker-0,5,main]
java.lang.IncompatibleClassChangeError: Found class org.apache.hadoop.mapreduce.TaskAttemptContext, but interface was expected
	at org.apache.spark.mapred.SparkHadoopMapRedUtil$.commitTask(SparkHadoopMapRedUtil.scala:95)
	at org.apache.spark.SparkHadoopWriter.commit(SparkHadoopWriter.scala:106)
	at org.apache.spark.rdd.PairRDDFunctions$$anonfun$13.apply(PairRDDFunctions.scala:1082)
	at org.apache.spark.rdd.PairRDDFunctions$$anonfun$13.apply(PairRDDFunctions.scala:1059)
	at org.apache.spark.scheduler.ResultTask.runTask(ResultTask.scala:61)
	at org.apache.spark.scheduler.Task.run(Task.scala:64)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:203)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
	at java.lang.Thread.run(Thread.java:745)
2015-06-01 17:59:57 WARN  TaskSetManager:71 - Lost task 0.0 in stage 0.0 (TID 0, localhost): java.lang.IncompatibleClassChangeError: Found class org.apache.hadoop.mapreduce.TaskAttemptContext, but interface was expected
	at org.apache.spark.mapred.SparkHadoopMapRedUtil$.commitTask(SparkHadoopMapRedUtil.scala:95)
	at org.apache.spark.SparkHadoopWriter.commit(SparkHadoopWriter.scala:106)
	at org.apache.spark.rdd.PairRDDFunctions$$anonfun$13.apply(PairRDDFunctions.scala:1082)
	at org.apache.spark.rdd.PairRDDFunctions$$anonfun$13.apply(PairRDDFunctions.scala:1059)
	at org.apache.spark.scheduler.ResultTask.runTask(ResultTask.scala:61)
	at org.apache.spark.scheduler.Task.run(Task.scala:64)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:203)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
	at java.lang.Thread.run(Thread.java:745)

2015-06-01 17:59:57 ERROR TaskSetManager:75 - Task 0 in stage 0.0 failed 1 times; aborting job
```

This is only for the Hadoop 1.0.4 builds. I'm wondering if this is an issue in Spark?

fnothaft:
I'm going to check this out locally and run with 1.0.4 and see if this failure reproduces.

fnothaft:
Ahhhh... this fails locally (with the error Jenkins is reporting) for me when running:

```
mvn package -Dhadoop.version=1.0.4
```

Surely, they didn't break Hadoop 1.0.4 support in Spark 1.3.1, did they?

fnothaft:
@ryan-williams I'm hoping to cut a 0.17.0 release tomorrow. Do you want to check and see if you can resolve the Spark 1.3.1/Hadoop 1.0.4 issue by tomorrow? If you can't, shall we slip this to 0.18.0?

ryan-williams:
good call, thanks for catching. I'll see what I can do, though not sure how much time I'll have :-\

fnothaft:
No worries. We can take more of a hack at this later.

ryan-williams:
So, I am tentatively thinking that depending on Spark 1.3.1 excludes the possibility to build against Hadoop 1.0.4. [This user@spark thread is relevant](http://search-hadoop.com/m/q3RTtFIt3G1pYNH1&subj=spark+1+3+1+jars+in+repo1+maven+org), and I just responded cc'ing @fnothaft, explaining the failure we're seeing and linking to this PR.

Spark 1.3.1 is "provided" and not bringing in the Hadoop2.2 dependency that it [is hard-coded into it](https://repo1.maven.org/maven2/org/apache/spark/spark-core_2.10/1.3.1/spark-core_2.10-1.3.1.pom), which means that any Spark code-paths we exercise run the risk of using bits of Hadoop2 that might not have existed in a compatible way in Hadoop1.0.4.

I suspect there's a way to shade Hadoop2.2 into Spark so that this would work for us, but I assume that would have to be done in Spark's POM; I've not seen Maven functionality related to shading an arbitrary version of a dep's dep into said dep, if that makes sense. 

Curious to hear others' thoughts!

* How bad is the prospect of dropping Hadoop1.0.4 support?
* ADAM-1.0 bad? 
* or just ADAM-0.1[78].0 bad?
* or: [fork Spark, shade Hadoop2 into it, and publish and link ADAM against that] bad?

fnothaft:
> So, I am tentatively thinking that depending on Spark 1.3.1 excludes the possibility to build against Hadoop 1.0.4. This user@spark thread is relevant, and I just responded cc'ing @fnothaft, explaining the failure we're seeing and linking to this PR.

Yeah... I am moderately frustrated here. Either we're doing something wrong with how we're building against Spark 1.3.1, or their testing infrastructure is really weak. While it would be a lie to say that running on Hadoop 1.0.4 has always been seamless, it's generally worked OK.

What is possible is that we're running into a "build-only" problem. E.g., it is possible that the Maven artifact for Spark `org.apache.spark/spark-core:1.3.1` is only compatible with Hadoop 2.x, but the Apache release artifacts are valid (e.g., http://d3kbcqa49mib13.cloudfront.net/spark-1.3.1-bin-hadoop1.tgz will not have errors). If this is the case, perhaps we could resolve it by moving Spark and Hadoop from `provided` to `system` dependencies? We'd then need something in our build to check that `SPARK_HOME` is defined and is valid (or something like that), but that _should_ be straightforward.

> Spark 1.3.1 is "provided" and not bringing in the Hadoop2.2 dependency that it is hard-coded into it, which means that any Spark code-paths we exercise run the risk of using bits of Hadoop2 that might not have existed in a compatible way in Hadoop1.0.4.

We run a matrix build in Jenkins that sweeps the Hadoop version across 1.0.4, 2.2.0, 2.3.0. I believe this is sufficient for testing; it did catch the issue we would've run into here, and has caught similar past issues. Perhaps I'm misunderstanding what you're getting at?

> I suspect there's a way to shade Hadoop2.2 into Spark so that this would work for us, but I assume that would have to be done in Spark's POM; I've not seen Maven functionality related to shading an arbitrary version of a dep's dep into said dep, if that makes sense.

I don't think there's a straightforward way to do that. I think the best solution would be a fix to how Spark distributes resources. E.g., [Spark provides pre-built artifacts for different Hadoop versions](spark.apache.org/downloads.html). While they're building these artifacts, they could also deploy those same artifacts to Maven Central, with a [classifier that corresponds to the Hadoop version](https://maven.apache.org/plugins/maven-deploy-plugin/examples/deploying-with-classifiers.html).

ryan-williams:
I think we have the same understanding though we may have talked past each other a bit.

> it is possible that the Maven artifact for Spark org.apache.spark/spark-core:1.3.1 is only compatible with Hadoop 2.x

Exactly. @srowen basically asserts this in his first email on [the thread I linked previously](http://search-hadoop.com/m/q3RTtFIt3G1pYNH1&subj=spark+1+3+1+jars+in+repo1+maven+org) (I can't link directly to it on that site; [here it is on apache.org](http://mail-archives.apache.org/mod_mbox/incubator-spark-user/201505.mbox/%3CCAMAsSd+8w0tU4L5y1jL_TsVroRercH2pRNaAUXx4-pk7TZ94sg@mail.gmail.com%3E)):

> Yes, the published artifacts can only refer to one version of anything (OK, modulo publishing a large number of variants under classifiers).

So Spark is deliberately only publishing artifacts built for one version of Hadoop at a time, though they test against a matrix similar to ADAM's, I think. OTOH, Spark and ADAM both follow the common convention of actually shipping separate artifacts for Scala `2.10` vs. `2.11`. 

Finally, to your implication that Spark doesn't run tests for multiple versions of Hadoop, I also didn't think that was the case but can't find evidence to the contrary poking around Spark PRs and Jenkins setup; could be a reasonable other question to ask if we ever get a response on the aforelinked thread.

> we could resolve it by moving Spark and Hadoop from provided to system dependencies

From the little I know of Maven and `provided` vs. `system` deps, I think I would prefer that we just built the Spark we want and linked against that.

The best case is that there is an already-published `spark-core` `1.3.1` JAR that works for Hadoop1 that we can link against, *and* that we believe that whatever published that JAR will continue to do so for future Spark versions.

On that note, I'm curious about where you found the [http://d3kbcqa49mib13.cloudfront.net/spark-1.3.1-bin-hadoop1.tgz](http://d3kbcqa49mib13.cloudfront.net/spark-1.3.1-bin-hadoop1.tgz) you linked to?

srowen:
Not quite -- the idea is that Spark's Hadoop dependency in the POM is irrelevant, since it's consumed as "provided". So there is just one set of artifacts (which happens to depend on 2.2). It is supposed to work with 1.0.4 onwards at runtime.

The deployed copy of Spark of course needs some Hadoop bindings, and that version can be 1.0.4, or at least I know of no reason it shouldn't work. The use of Hadoop should still be compatible with 1.0.4; just because it happens to link vs 2.2 doesn't mean it can't link to 1.0.4. So, the current artifact is what you need.

Yes Scala has its own convention that doesn't use classifiers, and in that case, there's no way to support two Scala versions at one time in the same way.

This is the full set of Hadoop tests for Spark, buried in https://amplab.cs.berkeley.edu/jenkins Click into "pre YARN" builds in particular.

I kind of doubt 1.x support will survive much longer, but I don't know of any sudden moves in that direction.

fnothaft:
Apologies if this reply is a bit rushed; I'm running between sessions at a conference...

@srowen 

> Not quite -- the idea is that Spark's Hadoop dependency in the POM is irrelevant, since it's consumed as "provided". So there is just one set of artifacts (which happens to depend on 2.2). It is supposed to work with 1.0.4 onwards at runtime.
> 
> The deployed copy of Spark of course needs some Hadoop bindings, and that version can be 1.0.4, or at least I know of no reason it shouldn't work. The use of Hadoop should still be compatible with 1.0.4; just because it happens to link vs 2.2 doesn't mean it can't link to 1.0.4. So, the current artifact is what you need.

From my observation, that's like 75% correct. Even with proper shading, you _generally_ need to custom build ADAM for Hadoop 1.0.4 to run on Hadoop 1.0.4 because of how some of the various interfaces in Hadoop get pulled in. Specifically, the custom input formats that we have break.

That being said, this has traditionally not been an issue with the Spark artifacts. But, when trying to upgrade to Spark 1.3.1, we find that there is a Spark internal interface (seems to be somewhere related to the `PairRDDFunctions`, IIRC) that gets compiled as specific to Hadoop 2. If we exclude Hadoop from Spark and explicitly depend on Hadoop 1.0.4, we get this error:

```
java.lang.IncompatibleClassChangeError: Found class org.apache.hadoop.mapreduce.TaskAttemptContext, but interface was expected
    at org.apache.spark.mapred.SparkHadoopMapRedUtil$.commitTask(SparkHadoopMapRedUtil.scala:95)
    at org.apache.spark.SparkHadoopWriter.commit(SparkHadoopWriter.scala:106)
    at org.apache.spark.rdd.PairRDDFunctions$$anonfun$13.apply(PairRDDFunctions.scala:1082)
    at org.apache.spark.rdd.PairRDDFunctions$$anonfun$13.apply(PairRDDFunctions.scala:1059)
    at org.apache.spark.scheduler.ResultTask.runTask(ResultTask.scala:61)
    at org.apache.spark.scheduler.Task.run(Task.scala:64)
    at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:203)
    at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
    at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
    at java.lang.Thread.run(Thread.java:745)
```

I agree with you that this is likely not a problem in the packaged Spark artifacts available from spark.apache.org/downloads.html. My point is mostly that unless Spark publishes separate Maven artifacts for Hadoop 1.0.4, 2.2, etc. under a classifier, it's kind of difficult to build a downstream app simultaneously against Spark 1.3.1 and Hadoop 1.0.4.

> I kind of doubt 1.x support will survive much longer, but I don't know of any sudden moves in that direction.

I assume then that the `spark-ec2` scripts would be rewritten for Hadoop 2? The only reason that we still support Hadoop 1.0.4 in our build is because that's the default Hadoop version in the `spark-ec2` scripts. Additionally, the support for Hadoop 2 in the `spark-ec2` scripts is _rough_. We're trying to move away from `spark-ec2`, but that's further downstream.

@ryan-williams 
> On that note, I'm curious about where you found the http://d3kbcqa49mib13.cloudfront.net/spark-1.3.1-bin-hadoop1.tgz you linked to?

spark.apache.org/downloads.html has prebuilt release copies of Spark for the various Hadoop streams. That means we wouldn't need to build our own Spark.

> From the little I know of Maven and provided vs. system deps, I think I would prefer that we just built the Spark we want and linked against that.
>
> The best case is that there is an already-published spark-core 1.3.1 JAR that works for Hadoop1 that we can link against, and that we believe that whatever published that JAR will continue to do so for future Spark versions.

Then reason I'm suggesting using a `system` dependency (which I really _don't like_, but seems like the only solution that would allow our build to pass _and_ to properly build the input formats) is that these artifacts are available (by pulling down one of the Spark pre-built tarballs). This is an ugly hack to make things work _outside_ of pure Maven, but I _think_ it'd work.

ryan-williams:
Thanks for that info @srowen. As @fnothaft said, it seems like we are in fact running into a Spark 1.3.1 code path that **is** incompatible with Hadoop 1.0.4.

Do we know that any, or "most", Spark apps can run against Spark 1.3.1 and Hadoop 1.0.4 and *not* run into this, or other incompatibilities?

Our crash happens in `PairRDDFunctions.saveAsHadoopDataset`, [line 1082](https://github.com/apache/spark/blob/v1.3.1/core/src/main/scala/org/apache/spark/rdd/PairRDDFunctions.scala#L1082); AFAICT I would expect this code path to always crash in this way, and all `PairRDDFunctions` "save" methods seem to funnel there (not to mention many `RDD` "save" methods).

Given all of that, I'm curious to hear if we have existence proofs of apps running 1.3.1 against 1.0.4 and writing to Hadoop?

ryan-williams:
Wow, thanks @zsxwing for [apache/spark#6599](https://github.com/apache/spark/pull/6599) which will apparently fix this.

I guess we can wait for that to make it into a Spark `1.3.2` or `1.4.1`, and then upgrade?

srowen:
In the first instance, you're talking about ADAM itself right? yes in general applications that use Hadoop need a recompile to work on 1.x versus 2.x. Spark also has to be compiled separately to run on Hadoop 1.x vs 2.x. But applications that run on Spark should not since they just interface with Spark, and the bits of the Hadoop API that Spark exposes don't vary from 1.x to 2.x. Of course, if an app separately used Hadoop APIs that did change from 1.x to 2.x they'd face the same problem.

The error you show indicates that the deployed Spark is expecting 2.x but is getting 1.x at runtime. What are you deploying? it should be a Spark build for the 1.x you run in this case. The Maven artifacts aren't relevant.

This is an internal API call in Spark so how can it affect your app? I think the disconnect is either that you are trying to package Spark, or, that ADAM is more than an app (like it has direct Hadoop dependencies) and that is causing similar but separate issues. Yes you'd have to cross-compile too, yes you'd have to match up with the right Spark cross-compilation at runtime.

But this doesn't have much to do with the Maven artifacts.
I can imagine some strange corner cases though where you might accidentally have, say, both old and new MapReduce or Hadoop jars on the compile-time classpath since they have different names, and the wrong one wins, and you end up compiling against Hadoop 2.x when you didn't mean to, and that is in a way due to the Spark Maven artifact. But that's a matter of blocking out the Hadoop deps from Spark maybe to be sure they never enter your build, and then declare your own Hadoop deps directly.

I don't agree with https://github.com/apache/spark/pull/6599 for the same reasons. There is no such thing as using the Maven artifact on Hadoop 1. Deployable assemblies are what gets run on a cluster. I guess I am not getting the problem yet, at least given the deployment model Spark requires.

fnothaft:
@ryan-williams 
I do believe (but am not 100% sure) that I've run avocado against Spark 1.3.1/Hadoop 1.0.4 and saved data. I think there are two possibilities:

1. The Spark 1.3.1/Hadoop 1.0.4 release artifacts are OK, but Spark 1.3.1/Hadoop 2.2.0 Maven artifacts cannot be used with Hadoop 1.0.4 (exclude Hadoop from Spark, provide explicit dependency on Hadoop 1.0.4). Since Spark and Hadoop are a provided dependency, the unit tests fail (since they rely on the artifacts that Maven pulls down) but the end-to-end pipeline is OK (because it depends on the provided Spark 1.3.1/Hadoop 1.0.4 release artifacts).
2. Spark 1.3.1 cannot support Hadoop 1.0.4 because of the issue fixed by https://github.com/apache/spark/pull/6599.

@srowen 

> In the first instance, you're talking about ADAM itself right? yes in general applications that use Hadoop need a recompile to work on 1.x versus 2.x. Spark also has to be compiled separately to run on Hadoop 1.x vs 2.x. But applications that run on Spark should not since they just interface with Spark, and the bits of the Hadoop API that Spark exposes don't vary from 1.x to 2.x. Of course, if an app separately used Hadoop APIs that did change from 1.x to 2.x they'd face the same problem.

Yes, agreed. I think my point was unclear in my original comment, but what I meant this comment to illustrate is _why_ we needed an explicit dependency on Hadoop in our pom.

> The error you show indicates that the deployed Spark is expecting 2.x but is getting 1.x at runtime. What are you deploying? it should be a Spark build for the 1.x you run in this case. The Maven artifacts aren't relevant.

Our unit tests are running using the Maven artifacts for Spark 1.3.1, with Hadoop excluded. We then have an explicit dependency on Hadoop version 1.0.4, so that we can build the correct input/output formats. The Maven artifacts _are_ relevant because this error is coming up during a unit test run. Although the `provided` designation in Maven means that the final produced artifacts will _not_ contain Hadoop or Spark, the unit tests will rely on the Maven artifacts for Hadoop and Spark.

> This is an internal API call in Spark so how can it affect your app? I think the disconnect is either that you are trying to package Spark, or, that ADAM is more than an app (like it has direct Hadoop dependencies) and that is causing similar but separate issues. Yes you'd have to cross-compile too, yes you'd have to match up with the right Spark cross-compilation at runtime.

Historically, excluding Hadoop from Spark and then explicitly setting the Hadoop version was OK and would build and run unit tests fine. I _think_ the problem we're running into here isn't an actual problem when we run ADAM as an app. I think it is solely a build problem that causes our unit tests to fail. I'm _guessing_ that we could fix the issue by moving Spark from a `provided` to a `system` dependency, and then requiring the [Spark published release](spark.apache.org/downloads.html) to exist on the system. This _would_ make us fully independent of the Spark Maven artifacts, which is kind of ugly.

> I don't agree with apache/spark#6599 for the same reasons. There is no such thing as using the Maven artifact on Hadoop 1. Deployable assemblies are what gets run on a cluster. I guess I am not getting the problem yet, at least given the deployment model Spark requires.

I'm in agreement here. I don't think we're seeing a problem that would manifest during _deployment_ (since the dependancies are provided), rather I think we're seeing a build issue (that causes our unit tests to fail on Hadoop 1.0.4). That being said, I think that the way we'd need to fix it downstream (in ADAM, that is) is really ugly (the above solution with `system` dependencies), and I think the best way to handle this would be for Spark to publish Maven artifacts (under a selector) for the spectrum of Hadoop versions.

srowen:
Yeah, so you're making an ad-hoc "deployment" from Maven artifacts for tests. As far as I know that was never guaranteed to work and would surprise me if it had been across 1.x and 2.x at the same time. Maybe it very nearly does, and theory aside, a little change might make it still happen to work and that's worth it. It's brittle though and I expect that 1.x support will just go away more readily than 2-3x more artifacts are published for this. No idea about spark-ec2. It's a different rhetorical question, but that's such an old Hadoop version, is it important?

fnothaft:
> Yeah, so you're making an ad-hoc "deployment" from Maven artifacts for tests. As far as I know that was never guaranteed to work and would surprise me if it had been across 1.x and 2.x at the same time. Maybe it very nearly does, and theory aside, a little change might make it still happen to work and that's worth it.

I don't know if it's ever been "guaranteed" to work, but we've done this split Hadoop 1.0.4/2.2.0/2.3.0 build for CI without issues since pre-Spark 1.0.0 IIRC.

> It's brittle though and I expect that 1.x support will just go away more readily than 2-3x more artifacts are published for this. No idea about spark-ec2. It's a different rhetorical question, but that's such an old Hadoop version, is it important?

The reason I mention `spark-ec2` is because `spark-ec2` is the only reason we've continued building for 1.0.4. Specifically, `spark-ec2` uses 1.0.4 and has bad support for the Hadoop 2.x stream. In any case, `spark-ec2` is fragile, and we're trying to move away from it.

The reason I suggest publishing more artifacts is that Spark is publishing these artifacts _already_ (to the Spark download page); it just isn't publishing them to Maven. It seems to me like this should be a fairly seamless change to make, but:

1. My assumptions about the ease of adding this to the Spark release process may be incorrect.
2. It sounds like Spark is planning to formally drop Hadoop 1.x support (although, has this been communicated anywhere? I'm not _surprised_ that this is the plan, but I hadn't heard about it), which would render the issue moot.

Anyways, Spark isn't going to retroactively push Spark 1.3.1 artifacts to Maven with a selector for Hadoop 1.x, so this discussion is somewhat hypothetical. On a more concrete level, it sounds like our choices are:

1. Drop official Hadoop 1.x support in ADAM post 0.17.0 (and, implicitly drop support for running ADAM on a `spark-ec2` cluster)
2. Do a hacky build that would depend on `system` dependencies for Spark in Maven.

srowen:
I should probably direct comment to https://github.com/apache/spark/pull/6599 since here I'm just speaking for myself, and that's not necessarily representative. I can see the very practical argument that a small fix is worth it to preserve some existing behavior even if not guaranteed; my only open question is whether that's really all this needs, but it may be so.

Yeah I don't know much about spark-ec2. I would assume it can / will work with Hadoop 2 if needed, especially since Hadoop 1 is getting so old.

I think it's only moderately hard to add more Maven artifacts but it sure makes a mess -- Spark would deploy almost 100 artifacts per version. Now you have to change the artifacts you even depend on based on the version of another thing -- that's hard to write in Maven. It also exacerbates the Scala 2.10/2.11 problem for which there's really no other answer. This is why I can appreciate that anything reasonable to avoid this is worthwhile.

I speak for myself when I say I think 1.x support is going away, but I do think it's true in the medium term. It doesn't get much usage; 2.2 is already the default; old YARN support already went away.

ryan-williams:
@srowen your point makes sense that we are:

1. pulling in a Spark that was explicitly built against Hadoop 2, 
2. giving it Hadoop 1 at runtime,
3. observing that it doesn't work, due to the Hadoop-version mismatch.

I agree that reflection whack-a-mole in Spark to try to make all Hadoop API calls {1,2}-cross-compatible is not a good solution, even if it is just the one call in [spark#6599](https://github.com/apache/spark/pull/6599) that doesn't work today.

However, I'm curious about your hesitance to just fix this in Maven:

> Now you have to change the artifacts you even depend on based on the version of another thing

This is true, but the "another thing" in question (i.e. Hadoop version) affects whether programs will execute successfully or not; such a thing seems absolutely worth specifying explicitly, to me.

Our (ADAM's) problems here start with 1. above: we have no choice but to link against a Spark built for Hadoop 2 (or use Maven in a brittler/messier way, e.g. manually downloading Spark artifacts as system deps); given the choice, we (ADAM) would happily adjust our POMs and workflows to link against the correct Spark JAR for each Hadoop version we want to support.

> that's hard to write in Maven

IANA Maven expert, but won't this just mean we'll have to have profiles for the version of Hadoop we want (hopefully very coarse-grained, e.g. `-Phadoop1` and a default Hadoop-2.*)? I thought this was solvable and something Spark already does in cases where it's simply not possible/feasible to release one Spark JAR that will work in different runtimes.

**Most importantly**, said profiles only need exist in things that depend on Spark, e.g. ADAM; they wouldn't require any changes to Spark other than the publishing of artifacts to Maven that are already built and published to [https://spark.apache.org/downloads.html](https://spark.apache.org/downloads.html).

srowen:
I think https://github.com/apache/spark/pull/6599 is a much easier way forward, if it really does do the trick, since it only has to last until 1.x goes away.

Yes, specifying dependency versions is vital in general. This makes it significantly harder to set Hadoop version since you are not only setting Hadoop version, but must apply one Maven _profile_ or the other to get the correct Spark version. Defaults won't even work. A hard solution to a problem that must be solved is better than no solution but the PR above is much more inocuous.

ryan-williams:
@srowen fair enough. We are also assuming that [spark#6599](https://github.com/apache/spark/pull/6599) fixes the only incompatible Hadoop API call that we are going to run in to here; I'll build a Spark 1.3.1 with [spark#6599](https://github.com/apache/spark/pull/6599) patched in and run ADAM tests against it now, to check.

ryan-williams:
I ran ADAM tests against Spark-1.3.1+#6599 and Hadoop 1.0.4; the crash reported above here is fixed but the same issue crops up elsewhere:

```
java.lang.IncompatibleClassChangeError: Found class org.apache.hadoop.mapreduce.TaskAttemptContext, but interface was expected
	at org.bdgenomics.adam.io.SingleFastqInputFormat.createRecordReader(SingleFastqInputFormat.java:291)
	at org.apache.spark.rdd.NewHadoopRDD$$anon$1.<init>(NewHadoopRDD.scala:131)
	at org.apache.spark.rdd.NewHadoopRDD.compute(NewHadoopRDD.scala:104)
	at org.apache.spark.rdd.NewHadoopRDD.compute(NewHadoopRDD.scala:66)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:277)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:244)
	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:35)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:277)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:244)
	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:35)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:277)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:244)
	at org.apache.spark.rdd.MapPartitionsRDD.compute(MapPartitionsRDD.scala:35)
	at org.apache.spark.rdd.RDD.computeOrReadCheckpoint(RDD.scala:277)
	at org.apache.spark.rdd.RDD.iterator(RDD.scala:244)
	at org.apache.spark.scheduler.ResultTask.runTask(ResultTask.scala:61)
	at org.apache.spark.scheduler.Task.run(Task.scala:64)
	at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:203)
	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142)
	at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:617)
	at java.lang.Thread.run(Thread.java:745)
```

This time it's ADAM code that is calling a method on a `TaskAttemptContext`: [SingleFastqInputFormat.java:291](https://github.com/bigdatagenomics/adam/blob/4aadaf4e67ffc34f8b9c4580dd555bc2fc976059/adam-core/src/main/java/org/bdgenomics/adam/io/SingleFastqInputFormat.java#L291), so the "null hypothesis" that Spark is fixed by spark#6599 remains unrefuted.

I'll try to work around the issue in that ADAM code path and see where we are then.

ryan-williams:
OK @fnothaft, check out the 2nd commit here: it creates a wrapper for `TaskAttemptContext.setStatus` in the style of spark#6599 and uses it in `SingleFastqInputFormat`.

This gets all the tests passing, locally at least, and when run against a Spark-1.3.1 with spark#6599 patched in.

Totally mysterious to me is why [the same call in `InterleavedFastqInputFormat`](https://github.com/bigdatagenomics/adam/blob/4aadaf4e67ffc34f8b9c4580dd555bc2fc976059/adam-core/src/main/java/org/bdgenomics/adam/io/InterleavedFastqInputFormat.java#L314) is not causing tests to fail; without this hack, I saw `SingleFastqInputFormatSuite`, `AlignmentRecordRDDFunctionsSuite`, and `ADAMContextSuite` failing without this patch ([full test failure output](https://gist.github.com/ryan-williams/df57ed9d47515ae665ca)), but `SingleFastqInputFormatSuite` and `InterleavedFastqInputFormatSuite` look basically identical to me.

massie:
@ryan-williams I think that wrapping the Hadoop `Job` using `ContextUtil.getConfiguration(job)` will do the trick here. The Hadoop 1 and 2 API incompatibilities included changing a class to an interface (sigh). The `ContextUtil` helper will use reflection to work around the mess.

ryan-williams:
@massie interesting, I was not aware of the `ContextUtil` you speak of; you're referring to `parquet.hadoop.util.ContextUtil`, already used elsewhere in ADAM, I'm inferring? 

I'm afraid I don't completely follow how you're suggesting I use it here; will it more cleanly work around:

1. the internal Spark use of `TaskAttemptContext` that was crashing at `SparkHadoopMapRedUtil.scala:95`?
2. ADAM's use of `TaskAttemptContext` at `SingleFastqInputFormat.java:291` (and also possibly `InterleavedFastqInputFormat.java:314`)?

If you can provide more details I'm happy to try to use it instead, I believe that there is a better way than what I've done here!

massie:
I was referring to `parquet.hadoop.util.ContextUtil`. Your `ADAMHadoopUtil` looks like a better way to go. Sorry for the github comment noise.

ryan-williams:
ah ok, no worries, thanks for pointing that out to me, I am uninitiated to the world of Hadoop 1-vs-2 incompatibilities and workarounds thereof!

fnothaft:
@ryan-williams 

> This time it's ADAM code that is calling a method on a TaskAttemptContext: SingleFastqInputFormat.java:291, so the "null hypothesis" that Spark is fixed by spark#6599 remains unrefuted.

Just to confirm, when you built ADAM, you did provide the `-Dhadoop.version=1.0.4` on the Maven command line? If you did not force the Hadoop version in the ADAM build, I would _expect_ this error. If you do force the Hadoop version, then you should get a clean build.

@srowen 

> Yeah I don't know much about spark-ec2. I would assume it can / will work with Hadoop 2 if needed, especially since Hadoop 1 is getting so old.

Technically speaking_, `spark-ec2` does "support" Hadoop 2.0.0, but that's the only Hadoop 2 minor version it supports. Additionally, Hadoop MR is broken for Hadoop 2.0.0 in `spark-ec2` (paging @tomwhite who can provide more details, if interested), which really severely limits the utility of `spark-ec2`'s Hadoop 2.x support.

> I think it's only moderately hard to add more Maven artifacts but it sure makes a mess -- Spark would deploy almost 100 artifacts per version.

[Spark provides 7 pre-built versions](http://spark.apache.org/downloads.html), times two for the Scala 2.10/2.11 change, so it should be 14, not 100 artifacts, no? Perhaps I'm missing something.

> Now you have to change the artifacts you even depend on based on the version of another thing -- that's hard to write in Maven.

Unless I'm misunderstanding what you're proposing, this is trivial to do in Maven via either profiles or classifiers.

Keep in mind that the problem we're discussing here only pops up _if_ the project that is building against Spark _also_ needs to build against Hadoop public interfaces (i.e., they define custom input formats). In most projects, you can use Spark as a provided dependency and just rely on the transitive Hadoop dependency and be fine. Since we depend on Hadoop anyways, we already have to parametrize our build for this. Adding a second level of parametrization (parametrize which version of Spark to pull down based on a desired Hadoop version) would be easy to layer on top.

*TL;DR*: exactly what @ryan-williams said:

>> IANA Maven expert, but won't this just mean we'll have to have profiles for the version of Hadoop we want (hopefully very coarse-grained, e.g. -Phadoop1 and a default Hadoop-2.*)? I thought this was solvable and something Spark already does in cases where it's simply not possible/feasible to release one Spark JAR that will work in different runtimes.

I agree pretty much verbatim with @ryan-williams's [comment upthread](https://github.com/bigdatagenomics/adam/pull/690#issuecomment-108493276).

> It also exacerbates the Scala 2.10/2.11 problem for which there's really no other answer. This is why I can appreciate that anything reasonable to avoid this is worthwhile.

Doing split builds for Scala 2.10/2.11 purely in Maven is difficult because by convention Scala projects change the artifact name with a change in major Scala version, which Maven doesn't support. However, it's fairly straight forward to script these split builds up (e.g., see https://github.com/bigdatagenomics/adam/blob/master/scripts/release/release.sh).

srowen:
I'm talking about Maven artifacts, which are quite a separate issue from the pre-built assemblies you are linking to. These are not at all the same thing. Spark has ~22 Maven modules, times 2 Scala versions. Times 2 Hadoop versions is 88!

"Just" having to make 2 profiles and always having to specify at least 1 of them on a command line seems like a lot of pain for a dependency to require. Definitely doable but not fun, even for projects that are willing to do it. 

I'd definitely prefer to hear that a few tweaks just make this problem go away rather than bother with Hadoop-specific artifacts, which I doubt will happen. Why is it preferable to have this mess spill into the Maven artifacts? This seems a few steps down the list of backup plans.

ryan-williams:
@fnothaft good question, I'll double-check that I was seeing that error on a `-Dhadoop.version=1.0.4` ADAM.

@srowen 

> "Just" having to make 2 profiles and always having to specify at least 1 of them on a command line seems like a lot of pain for a dependency to require.

What would "always" have to be specified on the cmdline? IIUC, ADAM would default to Hadoop-2 artifacts for everything, and have a `-Phadoop-1.0.4` profile instead of the `-Dhadoop.version=1.0.4` we have today… and ADAM is a relatively complex case.

Most things that depend on Spark are probably OK supporting one Hadoop (major) version at a time; they would just link to those artifacts in their POM and be done with it.

Anyway, I agree that merging in spark#6599 is simpler in the short term, so I think we're just debating whether the Maven route is 

1. better in the long term, and 
2. better enough in the long term to warrant the cost of publishing more Maven artifacts.

Seems like @fnothaft and I definitely believe 1), while 2) is debatable, but that you are skeptical about 1) because of the complexity you think it will add to things that depend on Spark…?

srowen:
The gotcha is that all default profiles get disabled once any profile is
enabled. You can probably use a sys property to make this better I suppose.

I suspect that 1.x support will just be dropped before publishing 2x the
artifacts happens as this is much madness to support very few deployments.
IMHO.

tomwhite:
To add to @fnothaft's comment about the spark-ec2 scripts: they are great for running Spark, but the version of MR that they install is old (1.x or 2.0, not the later 2.x versions) so it's probably best to avoid running MR jobs on a spark-ec2 cluster. ADAM doesn't have any MR jobs (although see https://github.com/bigdatagenomics/adam/issues/651), so it's not a problem.

Is there any reason not to drop Hadoop 1 support? For the spark-ec2 scripts HDFS installation, the Hadoop 2 version could be used. Projects like Crunch and Kite have dropped Hadoop 1 support, or are in the process of doing so.

fnothaft:
@tomwhite 
You'll need a MR cluster up if you're going to distcp data from S3, no? It just seems to me that if MR is broken on `spark-ec2` Hadoop 2.0.0, we can't actually functionally use `--hadoop-major-version=2` for `spark-ec2`.

@srowen 
> I'm talking about Maven artifacts, which are quite a separate issue from the pre-built assemblies you are linking to. These are not at all the same thing. Spark has ~22 Maven modules, times 2 Scala versions. Times 2 Hadoop versions is 88!

I know that they're separate. I don't know how Spark is doing their Maven releases (e.g., we use the OSS Sonatype Maven release flow, which may be very different from the Apache Maven release flow), but for us, releasing the parent project (`adam-parent`) is essentially a "one click" operation that releases all 4 of our modules. For us, even releasing Scala 2.10/2.11 artifacts is effectively a "one click" operation...

In any case, it sounds like there isn't interest in Spark distributing Maven artifacts that are separate for Hadoop 1.x and 2.x, so I'm going to drop this point. I suppose that the takeaway then is that we will drop Hadoop 1.x support in the next release of ADAM. We'll then need to shake out things downstream of that.

ryan-williams:
I'm picking this back up since it sounds like we're cool with dropping Hadoop 1 support from ADAM. cc @heuermh who also expressed interest in this.

heuermh:
Great!

The jackson dependency issues should have been resolved via https://github.com/bigdatagenomics/adam/pull/744.  I'm not sure what to do about https://github.com/bigdatagenomics/utils/issues/41.

fnothaft:
@ryan-williams @laserson @heuermh should we close this in favor of #753?

laserson:
sgtm

PR #684, opened 2015-05-29
a few region-join nits

ryan-williams:
fixes #683

PR #706, opened 2015-06-03
restore clause guarding pruneCache check

ryan-williams:
removing in #684 was a mistake, assumed it was redundant but it's not!


fnothaft:
Merged! Thanks @ryan-williams, and thanks @laserson for catching the issue! My apologies for missing it the first time around.

PR #712, opened 2015-06-04
unify header-setting between BAM/SAM and VCF

### LINKED ISSUE #711: Unify header-setting between BAM/SAM and VCF
There are some inconsistencies between the fixes to #353 and #676 that we should clear up.

Also, the WARN logs should mostly be made INFOs.

ryan-williams:
fixes #711


ryan-williams:
bump!

fnothaft:
Merged! Thanks @ryan-williams!

PR #719, opened 2015-06-25
[ADAM-718] Use filesystem path to get underlying file system.

### LINKED ISSUE #718: ADAM Fails to find FS when converting BAM file
See [mailing list post](https://groups.google.com/forum/#!topic/adam-developers/UuZeJrowVyw).

A fix was proposed by @jfeala, which modified [ADAMContext](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala):

```
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
@@ -232,8 +232,8 @@ class ADAMContext(val sc: SparkContext) extends Serializable with Logging {
   def loadBam(
     filePath: String): RDD[AlignmentRecord] = {

-    val fs = FileSystem.get(sc.hadoopConfiguration)
     val path = new Path(filePath)
+    val fs = FileSystem.get(path.toUri, sc.hadoopConfiguration)
     val bamFiles = if (fs.isDirectory(path)) fs.listStatus(path) else fs.globStatus(path)
     val (seqDict, readGroups) = bamFiles
       .map(fs => fs.getPath)
```

fnothaft:
Resolves #718. Uses @jfeala's proposed fix. @jfeala, let me know if you'd like credit on the commit and I'll modify the commit author.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
@@ -232,8 +232,8 @@ class ADAMContext(val sc: SparkContext) extends Serializable with Logging {
   def loadBam(
     filePath: String): RDD[AlignmentRecord] = {
 
-    val fs = FileSystem.get(sc.hadoopConfiguration)
     val path = new Path(filePath)
+    val fs = FileSystem.get(path.toUri, sc.hadoopConfiguration)
     val bamFiles = if (fs.isDirectory(path)) fs.listStatus(path) else fs.globStatus(path)
     val (seqDict, readGroups) = bamFiles
       .map(fs => fs.getPath)


massie:
+1 Looks good to me.

jfeala:
sure, my github cred could use a boost. Thanks @fnothaft!

fnothaft:
Done! Thanks @jfeala!

laserson:
Manually merged in a66b6381e72eb821872aca4b325e6fe6d94af827

fnothaft:
@laserson are you sure this merged correctly? I don't see a66b638 in master.

laserson:
oops, I think you're right.

fnothaft:
https://github.com/fnothaft/adam/commit/a66b6381e72eb821872aca4b325e6fe6d94af827 is linearly based off of https://github.com/fnothaft/adam/commit/b0ff134b658aaf77d470683fcb0a7452a0c3a7b5 so it should be OK to just use the github merge button.

laserson:
Done.  Actually, is there an easy way to tell whether a PR is linearly branched off the latest upstream master?

fnothaft:
@laserson I'd been frustrated that there wasn't an easy way to tell that in Github for _the longest time_, but I literally just realized how. If you look at a single commit in github (e.g., https://github.com/fnothaft/adam/commit/a66b6381e72eb821872aca4b325e6fe6d94af827), it lists the parent commit(s). If there is a single parent, and that parent commit is master/HEAD, then the commit is good. If the PR has multiple commits, then you just need to walk the commits in succession.

E.g., for this PR, the parent (https://github.com/bigdatagenomics/adam/commit/b0ff134b658aaf77d470683fcb0a7452a0c3a7b5) is immediately right to the long commit hash for the single commit.
screen shot 2015-06-25 at 3 50 03 pm

laserson:
But does GitHub display anywhere that a particular commit is master/HEAD?  Or do I still need to crossreference that with git log in my terminal to determine that?

laserson:
I guess I can browse to https://github.com/bigdatagenomics/adam/commits/master

laserson:
Still not as simple as could be...

fnothaft:
@laserson eh! It is what it is...

fnothaft:
I think the best approach would be for Github to have an option that _only_ allowed fast-forward merges via the PR page (which I think would disallow non-linear merges), but for whatever reason Github has chosen to only allow non-fast-forward merges via the PR page (which leads to a merge commit per PR merged via Github's merge button, even if the merge could've been a fast-forward).

laserson:
Apparently that's gitflow best practices, though it's a little bizarre.

fnothaft:
What's a bit bizarre?

PR #720, opened 2015-06-26
Updated Variant Projection Schema, changed Field classes to objects

erictu:
This allows Projections to be used for Variant and VariantCallingAnnotations


fnothaft:
@erictu can you run `./scripts/format-source`? That's why the build is failing. I will merge when fixed.

fnothaft:
@AmplabJenkins is down, so I verified this manually (I've dropped @shaneknapp an email) and merged as https://github.com/bigdatagenomics/adam/commit/5c82d9476450264dfda5cbd2c2cfe0bcaba591df.

shaneknapp:
jenkins is back up...  looks like there was a power event in the colo and
all of the workers (which aren't on UPS) went down.

PR #727, opened 2015-07-13
Update MdTag parsing for CIGAR N

### LINKED ISSUE #726: MDTag deletions are incorrect for reads with N CIGAR operators
The current `MdTag` constructor is 
```
def apply(mdTagInput: String, referenceStart: Long): MDTag
```
however, without the CIGAR, this will place mismatches/deletions at the wrong positions if the CIGAR contains `N` operators (as in an RNA read)

example:
```
sequence: CCAGCGATGAAGCCGCCCAGTTCAATACAACAAGTGAGTTTGACTCATCA
mdTag: 28^A22
cigar: 15M_2816N_13M1D22M
```

To handle this, it seems the MDTag should process the CIGAR as well?  I can add this in, but wanted to see if this was the correct approach or whether this case should be handled separately?

arahuja:
Fixes #726 

This fixes the MdTag parsing to respect N operators in the CIGAR.  It turned to out be trickier than I expected, so if this isn't a case we want to support or want to handle it differently, I'd be fine with that.

I added 2 tests for that case as well but removed one test that required an error if an MDTag did not end with a digit - is that a valid constraint?  I couldn't figure out how it would affect parsing or make it ambiguous?

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/util/MdTag.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/util/MdTag.scala
@@ -42,72 +42,105 @@ object MdTag {
    *
    * @param mdTagInput Textual MD tag/mismatchingPositions string.
    * @param referenceStart The read start position.
+   * @param cigar Cigar operators for the read
    * @return Returns a populated MD tag.
    */
-  def apply(mdTagInput: String, referenceStart: Long): MdTag = {
+  def apply(mdTagInput: String,
+            referenceStart: Long,
+            cigar: Cigar): MdTag = {
+
     var matches = List[NumericRange[Long]]()
     var mismatches = Map[Long, Char]()
     var deletions = Map[Long, Char]()
 
-    if (mdTagInput != null && mdTagInput == "0") {
+    val numCigarOperators = cigar.numCigarElements()
+    var cigarIdx = 0
+    var mdTagStringOffset = 0
+    var referencePos = referenceStart
+    var cigarReferencePosition = referenceStart
+
+    if (mdTagInput == null || mdTagInput == "0") {
       new MdTag(referenceStart, List(), Map(), Map())
-    } else if (mdTagInput != null && mdTagInput.length > 0) {
+    } else {
       val mdTag = mdTagInput.toUpperCase
-      val end = mdTag.length
-
-      var offset = 0
-      var referencePos = referenceStart
-
-      def readMatches(errMsg: String): Unit = {
-        digitPattern.findPrefixOf(mdTag.substring(offset)) match {
-          case None => throw new IllegalArgumentException(errMsg)
-          case Some(s) =>
-            val length = s.toInt
-            if (length > 0) {
-              matches ::= NumericRange(referencePos, referencePos + length, 1L)
-            }
-            offset += s.length
-            referencePos += length
-        }
-      }
-
-      readMatches("MD tag must start with a digit")
-
-      while (offset < end) {
-        val mdTagType = {
-          if (mdTag.charAt(offset) == '^') {
-            offset += 1
-            MdTagEvent.Delete
-          } else {
-            MdTagEvent.Mismatch
+      // Parse all CIGAR operators until the end of the mdTag string
+      while (cigarIdx < numCigarOperators && mdTagStringOffset < mdTagInput.length) {
+        val cigarElement = cigar.getCigarElement(cigarIdx)
+        val nextDigit = digitPattern.findPrefixOf(mdTag.substring(mdTagStringOffset))
+        (cigarElement.getOperator, nextDigit) match {
+          case (_, None) if mdTagStringOffset == 0 => throw new IllegalArgumentException(s"MdTag $mdTagInput does not start with a digit")
+          case (_, Some(matchingBases)) if matchingBases.toInt == 0 => {
+            mdTagStringOffset += 1
           }
-        }
-        basesPattern.findPrefixOf(mdTag.substring(offset)) match {
-          case None => throw new IllegalArgumentException("Failed to find deleted or mismatched bases after a match: %s".format(mdTagInput))
-          case Some(bases) =>
-            mdTagType match {
-              case MdTagEvent.Delete =>
-                bases.foreach {
-                  base =>
-                    deletions += (referencePos -> base)
-                    referencePos += 1
-                }
-              case MdTagEvent.Mismatch =>
-                bases.foreach {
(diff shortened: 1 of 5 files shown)


arahuja:
Also, while working on this, I was wondering is there a reason all the positions are stored as Long as opposed to Int? Seems the SAMRecord just uses Int?

fnothaft:
@arahuja thanks for the PR! I'll look over this in the afternoon.

> I added 2 tests for that case as well but removed one test that required an error if an MDTag did not end with a digit - is that a valid constraint? I couldn't figure out how it would affect parsing or make it ambiguous?

I suppose it would possibly make the MD tag's end position ambiguous if you have a deletion as the last operator. Otherwise, I think it's just a useful thing to check to ensure that all CIGAR operators are formatted as `XY` where _X_ is a number and _Y_ is an operator. I don't have a strong opinion here.

> Also, while working on this, I was wondering is there a reason all the positions are stored as Long as opposed to Int? Seems the SAMRecord just uses Int?

Yeah, I'm not sure why exactly we did that. I don't think there are any genomes longer than 2^32 bases. I've had this exact same question many times...

fnothaft:
Jenkins, retest this please.

fnothaft:
@arahuja it seems like this causes the BQSR unit tests to fail. Any thoughts?

arahuja:
Sure, I'll take a look - will also work on re-incorporating the constraint to end with a digit.

arahuja:
@fnothaft Updated to handle the `S` operator as well, as that was missing before.  I still did not add back in throwing an error an missing digit to end the MDTag, as I can't think of case where the reference positions would be wrong do that.  I can add it in, but it doesn't seem necessary.

I want to add some more test cases as well anyways

arahuja:
@fnothaft this is updated with a few additional tests.

Looking it over it seems `MdTag.getReference` is still wrong, but it's not clear to me what we'd want there - a giant array full of N's seems like a waste?  I'll open another issue on that.

arahuja:
While this was still open, @timodonnell found that the `toString` method needed to be updated as well.  I updated that and added tests for it as well

fnothaft:
Thanks @arahuja! I've manually committed this as https://github.com/bigdatagenomics/adam/commit/68f5b4a158049dea33513d7bce94828ff40ff993. The check failure was due to the format-source check, so I ran `./scripts/format-sources`, rebased the work on ToT, and squashed down to a single commit.

adam-core/src/main/scala/org/bdgenomics/adam/util/MdTag.scala
```diff
@@ -42,72 +42,105 @@ object MdTag {
...
+      // Parse all CIGAR operators until the end of the mdTag string
+      while (cigarIdx < numCigarOperators && mdTagStringOffset < mdTagInput.length) {
+        val cigarElement = cigar.getCigarElement(cigarIdx)
+        val nextDigit = digitPattern.findPrefixOf(mdTag.substring(mdTagStringOffset))
+        (cigarElement.getOperator, nextDigit) match {
+          case (_, None) if mdTagStringOffset == 0 => throw new IllegalArgumentException(s"MdTag $mdTagInput does not start with a digit")
+          case (_, Some(matchingBases)) if matchingBases.toInt == 0 => {
+            mdTagStringOffset += 1
+          }
+          case (CigarOperator.MATCH_OR_MISMATCH | CigarOperator.M | CigarOperator.EQ, Some(matchingBases)) => {
```
ryan-williams: isn't `CigarOperator.MATCH_OR_MISMATCH == CigarOperator.M` @arahuja ?

fnothaft: Affirmative:

```
fnothaft$ ./bin/adam-shell 
2015-07-30 08:40:38.262 java[34517:1903] Unable to load realm info from SCDynamicStore
2015-07-30 08:40:38 WARN  NativeCodeLoader:62 - Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
Welcome to
      ____              __
     / __/__  ___ _____/ /__
    _\ \/ _ \/ _ `/ __/  '_/
   /___/ .__/\_,_/_/ /_/\_\   version 1.3.1
      /_/

Using Scala version 2.10.4 (Java HotSpot(TM) 64-Bit Server VM, Java 1.7.0_51)
Type in expressions to have them evaluated.
Type :help for more information.
Spark context available as sc.
SQL context available as sqlContext.

scala> import htsjdk.samtools.{ Cigar, CigarOperator }
import htsjdk.samtools.{Cigar, CigarOperator}

scala> CigarOperator.M == CigarOperator.MATCH_OR_MISMATCH
res0: Boolean = true
```

adam-core/src/main/scala/org/bdgenomics/adam/util/MdTag.scala
```diff
@@ -42,72 +42,105 @@ object MdTag {
...
+            }
+          }
+          case (CigarOperator.DELETION, None) => {
+            mdTag.charAt(mdTagStringOffset) match {
+              case '^' => {
+                // Skip ahead of the deletion '^' character
+                mdTagStringOffset += 1
+                basesPattern.findPrefixOf(mdTag.substring(mdTagStringOffset)) match {
+                  case None => throw new IllegalArgumentException(s"No deleted bases found ${cigar.toString} in MDTag $mdTag")
+                  case Some(deletedBases) => {
```
ryan-williams: sanity check: should `deletedBases.length == cigarElement.getLength` here? i.e. the number of bases after the `^` in the MDTag should match the number prefixing the cigar operator `D`?

fnothaft: That sounds correct here.

fnothaft:
@ryan-williams just FYI, you're commenting on a PR that has been merged. ;)
I assume that you're suggesting changes that you'd like to see #748 (or some other, future pull request) address?

ryan-williams:
Hey @fnothaft (and @arahuja) yea, sorry for the late drive-by comments; Github seems to not allow commenting on arbitrary lines otherwise I'd do these on #748.

I was imagining we could fold some of these into #748. I'm preparing a PR to merge into #748 with my suggestions :)

fnothaft:
Ah, fascinating! I thought you could comment on arbitrary lines... Does Github restrict you to diff'ed lines on PRs? Anywho; that sounds good—looking forward to the changes!

ryan-williams:
Apparently / TIL: you get 3 lines either side of a diff line: 



L123-L134 there were made visible by my clicking "Expand"; they're a little darker grey and I can't comment on them.

fnothaft:
Good times, good times...

PR #732, opened 2015-07-20
Implemented BAM Indexing and Filtering

erictu:


adam-core/src/main/java/org/bdgenomics/adam/io/BAMFilteredRecordReader.scala
```diff
@@ -0,0 +1,159 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.bdgenomics.adam.io
+
+import org.apache.hadoop.conf.Configuration
```
fnothaft: Nit: can you order the imports? `htsjdk` should be before `org.*`.

adam-core/src/main/java/org/bdgenomics/adam/io/BAMFilteredRecordReader.scala
```diff
@@ -0,0 +1,159 @@
...
+import org.seqdoop.hadoop_bam.util.WrapSeekable
+import org.seqdoop.hadoop_bam.SAMRecordWritable
+import org.seqdoop.hadoop_bam.BAMRecordReader
+import org.seqdoop.hadoop_bam.FileVirtualSplit
+import org.seqdoop.hadoop_bam.BAMInputFormat
+import hbparquet.hadoop.util.ContextUtil
+import scala.annotation.tailrec
+
+object BAMFilteredRecordReader {
+  var optViewRegion: Option[ReferenceRegion] = None
```
fnothaft: I think this should be private (or at least protected).

adam-core/src/main/java/org/bdgenomics/adam/io/BAMFilteredRecordReader.scala
```diff
@@ -0,0 +1,159 @@
...
+
+  var stringency: ValidationStringency = _
+
+  var bci: BlockCompressedInputStream = _
+  var codec: BAMRecordCodec = _
+  var fileStart: Long = _
+  var virtualEnd: Long = _
+  var isInitialized: Boolean = false
+
+  override def initialize(spl: InputSplit, ctx: TaskAttemptContext) {
```
fnothaft: Can you document what is going on in this function via inline comments?

adam-core/src/main/java/org/bdgenomics/adam/io/BAMFilteredRecordReader.scala
```diff
@@ -0,0 +1,159 @@
...
+
+    val virtualStart = split.getStartVirtualOffset()
+
+    fileStart = virtualStart >>> 16
+    virtualEnd = split.getEndVirtualOffset()
+
+    bci.seek(virtualStart)
+    codec.setInputStream(bci)
+
+    if (BAMInputFormat.DEBUG_BAM_SPLITTER) {
```
fnothaft: Do we need to keep the debug code in?

adam-core/src/main/java/org/bdgenomics/adam/io/BAMFilteredRecordReader.scala
```diff
@@ -0,0 +1,159 @@
...
+    } else {
+      val r: SAMRecord = codec.decode()
+
+      // Since we're reading from a BAMRecordCodec directly we have to set the
+      // validation stringency ourselves.
+      if (this.stringency != null) {
+        r.setValidationStringency(this.stringency)
+      }
+
+      if (r == null) {
```
fnothaft: Can you document this if/else clause? Specifically, it'd be good for it to be documented that this if/else block is used to push the predicate down onto a BGZIP block that the index has said contains data of interest to us. It'd also be good to note why this function is tail recursive (needed to work around stack overflow since predicate data can be sparse).

adam-core/src/main/java/org/bdgenomics/adam/io/IndexedBamInputFormat.scala
```diff
@@ -0,0 +1,98 @@
...
+package htsjdk.samtools
+
+import htsjdk.samtools.util.BlockCompressedFilePointerUtil
+import org.apache.hadoop.fs.{ FileSystem, Path }
+import org.apache.hadoop.mapreduce.JobContext
+import org.apache.hadoop.mapreduce.RecordReader
+import org.apache.hadoop.mapreduce.InputSplit
+import org.apache.hadoop.mapreduce.TaskAttemptContext
+import org.apache.hadoop.mapreduce.lib.input.FileSplit
+import org.seqdoop.hadoop_bam.BAMInputFormat
```
fnothaft: `org.seqdoop` after `org.bdgenomics`. Same nit (alphabetization) as earlier.

adam-core/src/main/java/org/bdgenomics/adam/io/IndexedBamInputFormat.scala
```diff
@@ -0,0 +1,98 @@
...
+import org.bdgenomics.adam.models.ReferenceRegion
+import org.bdgenomics.adam.rdd.ADAMContext._
+import org.bdgenomics.adam.io.BAMFilteredRecordReader
+import org.apache.hadoop.io.LongWritable
+import java.io.File
+import scala.collection.mutable
+
+object IndexedBamInputFormat {
+
+  var optFilePath: Option[Path] = None
```
fnothaft: Same nit around protection.

adam-core/src/main/java/org/bdgenomics/adam/io/IndexedBamInputFormat.scala
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+}
+
+class IndexedBamInputFormat extends BAMInputFormat {
+
+  override def createRecordReader(split: InputSplit, ctx: TaskAttemptContext): RecordReader[LongWritable, SAMRecordWritable] = {
+    val rr: RecordReader[LongWritable, SAMRecordWritable] = new BAMFilteredRecordReader()
+    assert(IndexedBamInputFormat.optViewRegion.isDefined)
+    BAMFilteredRecordReader(IndexedBamInputFormat.optViewRegion.get)
```
fnothaft: What does this line do? Since it is an apply, I'm a little surprised that the return value isn't being used.

adam-core/src/main/java/org/bdgenomics/adam/io/IndexedBamInputFormat.scala
```diff
@@ -0,0 +1,98 @@
...
+    rr.initialize(split, ctx)
+    rr
+  }
+
+  override def getSplits(job: JobContext): java.util.List[InputSplit] = {
```
fnothaft: Nit: add inline docs.

fnothaft:
This is the version that can only process a single BAM, right?

erictu:
Yup. loadIndexedBam in ADAMContext requires that it's only a single file.

fnothaft:
OK, when this merges, can you open an issue to support multiple files?

adam-core/src/main/java/org/bdgenomics/adam/io/IndexedBamInputFormat.scala
erictu: It sets the region that the BAMFilteredRecord Reader looks in

fnothaft: Ah, OK. IMO, that should probably not be in an `apply` method.

erictu:
Yup definitely.

fnothaft:
Thanks @erictu! Merged manually as https://github.com/bigdatagenomics/adam/commit/efd6711bf07f2ea44cfdc040aa80a86294926f1c.

PR #746, opened 2015-07-28
Upgrade to Parquet 1.8.1.

fnothaft:
This fixes an issue that appears when using `and`ed predicates to drop blocks on statistics; specifically, there is a bug pre-1.8.0 where fewer blocks are dropped than should be, which leads to bad performance.


massie:
Thanks, Frank!

PR #748, opened 2015-07-29
Fix for mdtag issues with insertions

arahuja:
Sorry all, looks the code in #727 contains a few errors when it came to parsing mismatches mixed with insertions.  I added a quite a few tests to cover this case and other complex mdtags.

If you think there are interesting cases not covered in the new tests now let me know!

Thanks!

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/util/MdTag.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/util/MdTag.scala
@@ -57,8 +57,8 @@ object MdTag {
     var cigarIdx = 0
     var mdTagStringOffset = 0
     var referencePos = referenceStart
-    var cigarReferencePosition = referenceStart
-
+    var cigarOperatorIndex = 0
+    var usedMatchingBases = 0
     if (mdTagInput == null || mdTagInput == "0") {
       new MdTag(referenceStart, List(), Map(), Map())
     } else {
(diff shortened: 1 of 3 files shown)


adam-core/src/main/scala/org/bdgenomics/adam/util/MdTag.scala
```diff
@@ -122,16 +135,16 @@ object MdTag {
...
             }
           }
           case _ if cigarElement.getOperator.consumesReferenceBases() => {
             referencePos += cigarElement.getLength
-            cigarReferencePosition += cigarElement.getLength
             cigarIdx += 1
+            cigarOperatorIndex = 0
           }
           case _ => {
             cigarIdx += 1
```
ryan-williams: is the `cigarOperatorIndex = 0` not necessary here?

mind commenting these last two cases a little bit? I assume the main interesting cigar operator being passed over here is `I`/insertion, which MDTags don't care about... any other cases being handled here worth noting?

adam-core/src/test/scala/org/bdgenomics/adam/util/MdTagSuite.scala
```diff
@@ -225,6 +225,131 @@ class MdTagSuite extends FunSuite {
...
+    assert(tag.isMatch(9))
+
+    assert(tag.mismatches(10L) === 'A')
+    assert(tag.isMatch(11))
+    assert(tag.isMatch(12))
+
+    assert(tag.toString === "2A4A2")
+  }
+
+  test("Get correct matches for mdtag with intron and deletion between mismatches") {
```
ryan-williams: hm, copy mcpastington? same test as above, different name?

arahuja: Yes! I should was trying to think of more cases, but didn't add the test actually

adam-core/src/test/scala/org/bdgenomics/adam/util/MdTagSuite.scala
```diff
@@ -225,6 +225,131 @@ class MdTagSuite extends FunSuite {
...
+    (3 to 5).foreach(locus => assert(tag.isMatch(locus)))
+
+    assert(tag.deletedBase(6L) === Some('A'))
+    assert(tag.deletedBase(7L) === Some('A'))
+    assert(tag.deletedBase(8L) === Some('A'))
+
+    (9 to 10).foreach(locus => assert(tag.isMatch(locus)))
+
+    assert(tag.mismatches(11L) === 'A')
+
```
ryan-williams: nit: missing a last "match" check on idx 12?

ryan-williams:
added some of my nits in https://github.com/hammerlab/adam/pull/2 including the check that the number of deleted bases is consistent between cigar and MDTag.

thoughts on some test cases verifying that IAEs are thrown when the various assumptions are broken?

adam-core/src/main/scala/org/bdgenomics/adam/util/MdTag.scala
arahuja: `cigarOperatorIndex = 0` is not necessary, it shouldn't be updated, but just added it for clarity.

The last case is insertion or clipped bases

ryan-williams: So it doesn't matter whether you set `cigarOperatorIndex = 0` here or not? That's also confusing to me, I don't fully have my head around this.

Is the second-to-last case one that you specifically added for `N`s? Is that the main focus there?

Could you add comments clarifying which operators you expect to hit each of these last two? At this point in the `match` I'm struggling to keep in my head what's left.

adam-core/src/test/scala/org/bdgenomics/adam/util/MdTagSuite.scala
arahuja: Yea figured it was repetitive at this point, but I'll add it

adam-core/src/main/scala/org/bdgenomics/adam/util/MdTag.scala
arahuja: Yes the second to last case is for `N`

arahuja:
@ryan-williams I fixed the dupe test and added a test case for the exception you added (mdtag and cigar disagreeing on deleted bases)

I also made the match statement explicit for the operators they were catching

re: `cigarOperatorIndex

ryan-williams:
cool, thanks @arahuja. LGTM once tests pass. Last complete build failed due to needing `./scripts/format-source` btw

arahuja:
Ah ok, good to know, I didn't realize this wasn't auto-formatting, I'll run that, squash etc.

fnothaft:
LGTM as well.

PR #669, opened 2015-05-03
don't remove CLI jar in compute-adam-jars.sh

ryan-williams:
fixes #663

Issue #663, opened 2015-04-30
include adam-cli in jars to send to Spark?


ryan-williams:
[This comment in `compute-adam-jars.sh`](https://github.com/bigdatagenomics/adam/blob/492fb3b04fc1ebe279e8aaa95f7b290cef911f27/bin/compute-adam-jars.sh#L25-L26) implies that we intentionally leave the `adam-cli` JAR off of the classpath for Spark jobs/shells.

I was stymied today for a while because I was working from an `adam-shell` and, among other things, wanted to run a `FlagStat` on a BAM in a way that was faithful to the way it gets run normally / via `adam-submit`, e.g. `FlagStat(Array(filename)).run(sc, job)`.

Is it important that CLI code not be made available to `adam-shell` sessions? Can/Should we remove this trimming of the `adam-cli` jar?

fnothaft:
Hah, drat! Yes, I'd agree that it makes sense _not_ to trim the _adam-cli_ jar. @laserson?

fnothaft:
Hah, drat! Yes, I'd agree that it makes sense _not_ to trim the `adam-cli` jar. @laserson?

ryan-williams:
yea, thinking on this more I became confused about how the CLI is available to e.g. `adam-submit`; well, seems like [we add it back in manually](https://github.com/bigdatagenomics/adam/blob/492fb3b04fc1ebe279e8aaa95f7b290cef911f27/bin/adam-submit#L90) after it's [been excluded from the `compute-adam-jars` calculation](https://github.com/bigdatagenomics/adam/blob/492fb3b04fc1ebe279e8aaa95f7b290cef911f27/bin/adam-submit#L23-L24). TIL!

That would seem to support adding it back in.

laserson:
hmm, my laziness tells me there should be some reason that I explicitly excluded the CLI jar, or else I wouldn't've done it.  But try adding it back in and see if it works.

ryan-williams:
seems to be working for me locally, I'll send a PR shortly

PR #749, opened 2015-07-31
Remove adam-cli jar from classpath during adam-submit

fnothaft:
The `./bin/compute-adam-jars.sh` script does not remote the `adam-cli` JAR from the classpath. This exposes [SPARK-1921](https://issues.apache.org/jira/browse/SPARK-1921) when running on YARN.

### DIFF
--- a/bin/adam-submit
+++ b/bin/adam-submit
@@ -48,35 +48,11 @@ SCRIPT_DIR="$(cd `dirname $0`/..; pwd)"
 
 # Get list of required jars for ADAM
 ADAM_JARS=$("$SCRIPT_DIR"/bin/compute-adam-jars.sh)
+ADAM_CLI_JAR=${ADAM_JARS##*,}
+ADAM_JARS=$(echo "$ADAM_JARS" | rev | cut -d',' -f2- | rev)
 
 # append ADAM_JARS to the --jars option, if any
 SPARK_ARGS=$("$SCRIPT_DIR"/bin/append_to_option.py , --jars $ADAM_JARS $SPARK_ARGS)
-
-# Assume we're in a binary distribution
-REPO_DIR="$SCRIPT_DIR/repo"
-if [ ! -d "$REPO_DIR" ]; then
-  # Fallback to source directory
-  REPO_DIR="$SCRIPT_DIR/adam-cli/target/appassembler/repo"
-fi
-
-# Find the ADAM CLI jar
-ADAM_DIR="$REPO_DIR/org/bdgenomics/adam"
-num_versions=$(ls ${ADAM_DIR} | grep cli | wc -l)
-if [ "$num_versions" -eq "0" ]; then
-  echo "Failed to find adam-cli jar in $ADAM_DIR"
-  echo "You need to build ADAM before running this program."
-  exit 1
-fi
-if [ "$num_versions" -gt "1" ]; then
-  versions_list=$(ls "$ADAM_DIR" | grep cli)
-  echo "Found multiple ADAM CLI versions in $ADAM_DIR:"
-  echo "$versions_list"
-  echo "Please remove all but one."
-  exit 1
-fi
-CLI=$(ls "$ADAM_DIR" | grep cli)
-CLI_DIR="${ADAM_DIR}/${CLI}"
-ADAM_CLI_JAR=$(ls $CLI_DIR/*/adam-cli_2.1[01]-*.jar)
 
 # Find spark-submit script
 if [ -z "$SPARK_HOME" ]; then
--- a/bin/compute-adam-jars.sh
+++ b/bin/compute-adam-jars.sh
@@ -22,8 +22,8 @@ ADAM_REPO="$(cd `dirname $0`/..; pwd)"
 
 CLASSPATH=$("$ADAM_REPO"/bin/compute-adam-classpath.sh)
 
-# list of jars to ship with spark; trim off the first from the CLASSPATH
-# TODO: brittle? assumes appassembler always puts the $BASE/etc first
-ADAM_JARS=$(echo "$CLASSPATH" | tr ":" "," | cut -d "," -f 2-)
+# list of jars to ship with spark; trim off the first from the CLASSPATH --> this is /etc
+# TODO: brittle? assumes appassembler always puts the $BASE/etc first and the CLI jar last
+ADAM_JARS=$(echo "$CLASSPATH" | tr ":" "\n" | tail -n +2 | perl -pe 's/\n/,/ unless eof' )
 
 echo "$ADAM_JARS"


bin/compute-adam-jars.sh
```diff
@@ -24,6 +24,6 @@ CLASSPATH=$("$ADAM_REPO"/bin/compute-adam-classpath.sh)
 
 # list of jars to ship with spark; trim off the first and last from the CLASSPATH
 # TODO: brittle? assumes appassembler always puts the $BASE/etc first and the CLI jar last
-ADAM_JARS=$(echo "$CLASSPATH" | tr ":" "," | cut -d "," -f 2-)
+ADAM_JARS=$(echo "$CLASSPATH" | tr ":" "\n" | tail -n +2 | head -n -1 | perl -pe 's/\n/,/ unless eof')
```
shaneknapp: i can't make heads or tails of this statement!  *rimshot*

fnothaft: Up top! *makes high five gesture*

ryan-williams:
So this undoes the "fix" #669 of #663. Can you say more about how this exposes SPARK-1921 @fnothaft?

fnothaft:
Ah! Good catch @ryan-williams. You all are running Spark-on-YARN on a CDH5 setup at Mt. Sinai, right? Perhaps the behavior is different between our clusters (which wouldn't be surprising, I think the bug is caused by a race condition), but we are in the middle of bringing up a CDH5 cluster at Berkeley and saw the exact behavior described in SPARK-1921 (JAR timestamps differ). By removing the adam-cli jar from the JARS list when calling `spark-submit`, you don't copy the JAR twice, and thus don't have the timestamp race. It should be easy to make this work for both `adam-submit` and `adam-shell`; I'll fix this up.

fnothaft:
@ryan-williams just pushed a new commit to address your comment. Doing things correctly actually makes the `adam-submit` code a bit cleaner, for certain definitions of clean (specifically, we get rid of the brittle way of determining the CLI jar). Let me know what you think, and I'll squash down and rebase.

bin/adam-submit
```diff
@@ -22,40 +22,15 @@ SCRIPT_DIR="$(cd `dirname $0`/..; pwd)"
 
 # Get list of required jars for ADAM
 ADAM_JARS=$("$SCRIPT_DIR"/bin/compute-adam-jars.sh)
+ADAM_CLI_JAR=$(echo "$ADAM_JARS" | tail -n 1)
 
+ADAM_JARS=$(echo "$ADAM_JARS" | tail -n +1 | perl -pe 's/\n/,/ unless eof')
```
ryan-williams: hm, `tail -n +1` doesn't do anything, does it? do you mean `tail -n +2` to skip the first line?

bin/compute-adam-jars.sh
```diff
@@ -22,8 +22,8 @@ ADAM_REPO="$(cd `dirname $0`/..; pwd)"
 
 CLASSPATH=$("$ADAM_REPO"/bin/compute-adam-classpath.sh)
 
-# list of jars to ship with spark; trim off the first and last from the CLASSPATH
+# list of jars to ship with spark; trim off the first from the CLASSPATH
 # TODO: brittle? assumes appassembler always puts the $BASE/etc first and the CLI jar last
```
ryan-williams: The code now only trims the first one, right? But is that supposed to be the CLI jar?

bin/compute-adam-jars.sh
```diff
@@ -22,8 +22,8 @@ ADAM_REPO="$(cd `dirname $0`/..; pwd)"
 
 CLASSPATH=$("$ADAM_REPO"/bin/compute-adam-classpath.sh)
 
-# list of jars to ship with spark; trim off the first and last from the CLASSPATH
+# list of jars to ship with spark; trim off the first from the CLASSPATH
```
ryan-williams: Maybe clarify that this command `echo`es the list of JARs, one per line

bin/adam-submit
fnothaft: Er, fun times. You are correct. This should've been `head -n -1` (working too late in the night...)  but I swapped that because BSD `head` doesn't support the `-n -1` flag. Let me find a good, portable fix and I'll ping back.

bin/compute-adam-jars.sh
ryan-williams: Answering my own question: no, it leaves the CLI jar in intentionally, for optional removal by callers.

Expounding on what is referred to by `$BASE/etc` here would be nice too, idk what that is otomh

fnothaft: Yeah, I was confused too. Essentially, `appassembler` puts `$BASE/etc` at the start of the classpath --> this maps to `/etc` on my system.

bin/compute-adam-jars.sh
fnothaft: Will do.

bin/adam-submit
ryan-williams: What if we have `compute-adam-jars.sh` take one cmdline flag, that tells it to strip the CLI jar or not, and then have it do the `perl -pe 's/\n/,/ unless eof'`, since that's what both callers want?

Then `compute-adam-jars.sh` could do some more structured things to find and remove the CLI jar, and we'd expose a clean interface to that to the two caller scripts.

ryan-williams: e.g. put a `grep -v <some unique adam cli jar identifier>` in `compute-adam-jars.sh`

fnothaft: Conceptually, I agree with what you are saying. In practice though, I actually think I like this approach a bit better, since it allows us to remove the fragile CLI jar finding code and fall back on the appassembler classpath instead.

ryan-williams: Yea, I was missing that you had that `grep` method in this file previously; it's certainly nice to see it deleted here.

I think there are two orthogonal questions:

1. find the CLI jar via the assumption that appassembler puts it last on the classpath vs. by `grep`ing.
2. do the above in `adam-submit` vs. in `compute-adam-jars.sh`.

I'm ok with either approach to 1), but am somewhat interested in moving 2) to `compute-adam-jars.sh`, that way `adam-submit` and `adam-shell` both save a `| perl -pe 's/\n/,/ unless eof'`, and we can create a cleaner interface to the `| head -n -1` that is currently here in `adam-submit`.

Anyways, the way you have it wfm too, just articulating the options I see.

ryan-williams:
this lgtm btw, unless I've persuaded you [here](https://github.com/bigdatagenomics/adam/pull/749#discussion_r36099270) to push CLI-jar detection and removal into `compute-adam-jars.sh`

massie:
@fnothaft Do you plan to move cli-jar detection tin `compute-adam-jars.sh` or is this ready to be merged? 

@ryan-williams If this is ready to be merged, feel free to hit the button and merge it.

fnothaft:
This isn't ready to merge yet; I still need to make fixes.

fnothaft:
@ryan-williams just pushed a new round of fixes. Let me know what you think of these.

ryan-williams:
Builds failed with 
```
hudson.plugins.git.GitException: Could not checkout null with start point 49a466a4d424b6acd1730c5daafe12f8e40d26b4
```

idk what that's about. Looking at the code now

fnothaft:
Jenkins, retest this please.

@ryan-williams there's a bug in the Jenkins Git plugin where fetching force-pushes fails occasionally.

bin/adam-submit
```diff
@@ -22,40 +22,15 @@ SCRIPT_DIR="$(cd `dirname $0`/..; pwd)"
 
 # Get list of required jars for ADAM
 ADAM_JARS=$("$SCRIPT_DIR"/bin/compute-adam-jars.sh)
+ADAM_CLI_JAR=$(echo "$ADAM_JARS" | rev | cut -d',' -f1 | rev )
```
ryan-williams: huh, isn't `$ADAM_JARS` already one comma-delimited line at this point? What's the point of the two `rev`s if so?

fnothaft: Maybe I am not so great at `cut`, but it isn't clear to me that there is an easy way to take _just the last_ item from a comma separated list using `cut`. However, you can fake this by reversing the list, taking _the first_ item, and reversing the list again.

heuermh: Blindly copied from http://stackoverflow.com/questions/22727107/how-to-find-the-last-field-using-cut-linux

```bash
data=foo,bar,baz,qux
last=${data##*,}
```

ryan-williams: huh, I read `man rev` as reversing the order of lines in a file, not reversing each line's contents, my mistake @fnothaft!

@heuermh's trick above seems maybe a little cleaner, but I'm agnostic at this point and have held up this PR long enough so whatever you want to do sgtm @fnothaft!

fnothaft:
Just pushed a last commit with Michael's suggested change, and with all things squashed down and rebased on master. This should be good to merge.

ryan-williams:
lgtm

massie:
+1

(Feel free to merge @ryan-williams)

heuermh:
the latest diff looks like it adds back the ```$ADDL_JARS``` stuff -- was this intentional?
https://github.com/fnothaft/adam/blob/fix-compute-jars/bin/adam-submit#L54

fnothaft:
@heuermh thanks for the catch; not intentional, rather this comes from a rebase gone wrong... I will fix.

fnothaft:
Fixed. Sorry; the rebase on #754 was kind of nasty; lots of messy conflicts.

ryan-williams:
want to retest this @fnothaft ? looks like git force-push problem again

ryan-williams:
good catch @heuermh btw! I looked at it and just assumed it was intentional, woops

fnothaft:
Jenkins, test this please.

fnothaft:
This is passing now; can I get a merge?

Issue #132, opened 2014-02-22
Allow list of commands to be injected into adam-cli AdamMain


heuermh:
The list of commands in AdamMain is private, so outside of the PluginExecutor mechanism, adding new commands requires modifying this list and rebuilding; this also introduces a circular dependency between the adam build and an external project.

Adding a new constructor to AdamMain with a reference to the list of commands as a parameter would allow the list to be provided by dependency injection (e.g. via Guice).

https://code.google.com/p/google-guice/

heuermh:
Fixed by #762 and #854

Issue #756, opened 2015-08-05
Investigate SparkLauncher API as a replacement for adam-submit+spark-submit


heuermh:
Docs
https://spark.apache.org/docs/1.4.0/api/java/org/apache/spark/launcher/package-summary.html

[SPARK-4924] Add a library for launching Spark jobs programmatically.
https://github.com/apache/spark/pull/3916

heuermh:
Closing as WontFix, see #813

PR #762, opened 2015-08-07
[ADAM-132] allowing list of commands to be injected into adam-cli ADAMMain

heuermh:
Work in progress towards #132.

Please be brutal in code review; I'm not very confident writing in scala or using scalatest yet.

### DIFF
--- a/adam-cli/src/main/scala/org/bdgenomics/adam/cli/ADAMMain.scala
+++ b/adam-cli/src/main/scala/org/bdgenomics/adam/cli/ADAMMain.scala
@@ -18,15 +18,16 @@
 package org.bdgenomics.adam.cli
 
 import java.util.logging.Level._
+import javax.inject.Inject
+import com.google.inject.AbstractModule
+import net.codingwell.scalaguice.ScalaModule
 import org.apache.spark.Logging
 import org.bdgenomics.adam.util.ParquetLogger
 import org.bdgenomics.utils.cli._
 
-object ADAMMain extends Logging {
+object ADAMMain {
 
-  case class CommandGroup(name: String, commands: List[BDGCommandCompanion])
-
-  private val commandGroups =
+  val defaultCommandGroups =
     List(
       CommandGroup(
         "ADAM ACTIONS",
@@ -69,6 +70,17 @@ object ADAMMain extends Logging {
       )
     )
 
+  def main(args: Array[String]) {
+    new ADAMMain(defaultCommandGroups)(args)
+  }
+}
+
+case class CommandGroup(name: String, commands: List[BDGCommandCompanion])
+
+private class InitArgs extends Args4jBase with ParquetArgs {}
+
+class ADAMMain @Inject() (commandGroups: List[CommandGroup]) extends Logging {
+
   private def printCommands() {
     println("\n")
     println("""     e            888~-_              e                 e    e
@@ -87,7 +99,7 @@ object ADAMMain extends Logging {
     println("\n")
   }
 
-  def main(args: Array[String]) {
+  def apply(args: Array[String]) {
     log.info("ADAM invoked with args: %s".format(argsToString(args)))
     if (args.size < 1) {
       printCommands()
@@ -115,10 +127,14 @@ object ADAMMain extends Logging {
     args.map(escapeArg).mkString(" ")
   }
 
-  class InitArgs extends Args4jBase with ParquetArgs {}
-
   private def init(args: InitArgs) {
     // Set parquet logging (default: severe)
     ParquetLogger.hadoopLoggerLevel(parse(args.logLevel))
+  }
+}
+
+class ADAMModule extends AbstractModule with ScalaModule {
+  def configure {
+    bind[List[CommandGroup]].toInstance(ADAMMain.defaultCommandGroups)
   }
 }
--- a/adam-cli/src/test/scala/org/bdgenomics/adam/cli/ADAMMainSuite.scala
+++ b/adam-cli/src/test/scala/org/bdgenomics/adam/cli/ADAMMainSuite.scala
@@ -1,0 +1,156 @@
+/**
+ * Licensed to Big Data Genomics (BDG) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The BDG licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.bdgenomics.adam.cli
+
+import java.io.ByteArrayOutputStream
+import com.google.inject.{ AbstractModule, Guice, Injector }
+import net.codingwell.scalaguice.ScalaModule
+import net.codingwell.scalaguice.InjectorExtensions._
+import org.bdgenomics.adam.cli.ADAMMain.defaultCommandGroups
+import org.scalatest.FunSuite
+
+class ADAMMainSuite extends FunSuite {
+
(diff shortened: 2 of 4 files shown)


heuermh:
When working on https://github.com/heuermh/adam-commands, downstream to this pull request, I ran into a build problem
```bash
$ mvn package
...
[ERROR] error while loading ADAMMain, Missing dependency 'bad symbolic reference. A
signature in ADAMMain.class refers to term inject in package javax which is not
available. It may be completely missing from the current classpath, or the version on
the classpath might be incompatible with the version used when compiling ADAMMain.class.',
required by /Users/xxx/.m2/repository/org/bdgenomics/adam/adam-cli_2.10/0.17.1-SNAPSHOT/adam-cli_2.10-0.17.1-SNAPSHOT.jar(org/bdgenomics/adam/cli/ADAMMain.class)
[ERROR] one error found
```
JDK8 vs JDK7 maybe?

massie:
@heuermh Have you figured out the issue with downstream projects yet?

heuermh:
@massie Haven't been able to reproduce today, works fine for me with both JDKs 7 and 8.  Must have been some maven snapshot strangeness.

heuermh:
The real question with this pull request is how to actually run the main in the repo above.

Duplicate most of the setup in ```adam-submit``` in order to call ```spark-submit``` directly?  Accept ```--main``` as a parameter in ```adam-submit```?  Implement a new ADAMMain that can be extended or delegated to that in turn delegates to SparkLauncher (#756).

Hopefully there's enough there to demonstrate the use case.

heuermh:
This can wait until post-0.17.1 release

heuermh:
Jenkins, retest this please

adam-cli/src/test/scala/org/bdgenomics/adam/cli/ADAMMainSuite.scala
```diff
@@ -0,0 +1,136 @@
...
+    val commandGroups = injector.instance[List[CommandGroup]]
+    assert(commandGroups == defaultCommandGroups)
+  }
+
+  test("inject default command groups when called via main") {
+    val stream = new ByteArrayOutputStream()
+    Console.withOut(stream) {
+      ADAMMain.main(Array())
+    }
+    assert(stream.toString().contains("flatten"))
```
massie: Calling the ADAMMain with an empty Array() will cause the help page to output, correct? If so, we should put a comment that effect and change the assert to look for the word "help" or some such.

adam-cli/src/test/scala/org/bdgenomics/adam/cli/ADAMMainSuite.scala
```diff
@@ -0,0 +1,136 @@
...
+    }
+    assert(stream.toString().contains("Build information"))
+  }
+
+  test("command groups is empty when called via apply") {
+    val stream = new ByteArrayOutputStream()
+    Console.withOut(stream) {
+      new ADAMMain(List.empty).apply(Array())
+    }
+    assert(!stream.toString().contains("flatten"))
```
massie: Same here...

adam-cli/src/test/scala/org/bdgenomics/adam/cli/ADAMMainSuite.scala
```diff
@@ -0,0 +1,136 @@
...
+  }
+
+  test("single command group") {
+    val stream = new ByteArrayOutputStream()
+    Console.withOut(stream) {
+      new ADAMMain(List(CommandGroup("SINGLE COMMAND GROUP", List(Flatten)))).apply(Array())
+    }
+    assert(stream.toString().contains("SINGLE"))
+    assert(stream.toString().contains("flatten"))
```
massie: We should also assert that commands from the default group are *not* seen.

massie: Oops. I misunderstood. This is *and* not *or*. Ignore that.

adam-cli/src/test/scala/org/bdgenomics/adam/cli/ADAMMainSuite.scala
```diff
@@ -0,0 +1,136 @@
...
+    Console.withOut(stream) {
+      ADAMMain.main(Array("buildinfo"))
+    }
+    assert(stream.toString().contains("Build information"))
+  }
+
+  test("command groups is empty when called via apply") {
+    val stream = new ByteArrayOutputStream()
+    Console.withOut(stream) {
+      new ADAMMain(List.empty).apply(Array())
```
massie: You can remove the `.apply` here, if you like, e.g. `new ADAMain(List.empty)(Array.empty)`

adam-cli/src/test/scala/org/bdgenomics/adam/cli/ADAMMainSuite.scala
```diff
@@ -0,0 +1,136 @@
...
+    }
+    assert(stream.toString().contains("NEW"))
+  }
+
+  test("module restores default command groups when called via apply") {
+    val stream = new ByteArrayOutputStream()
+    Console.withOut(stream) {
+      val injector = Guice.createInjector(new ADAMModule())
+      val commandGroups = injector.instance[List[CommandGroup]]
+      new ADAMMain(commandGroups).apply(Array())
```
massie: You can remove the apply here too.

adam-cli/src/main/scala/org/bdgenomics/adam/cli/ADAMMain.scala
```diff
@@ -69,6 +70,17 @@ object ADAMMain extends Logging {
       )
     )
 
+  def main(args: Array[String]) {
+    new ADAMMain(defaultCommandGroups).apply(args)
```
massie: No need for the .apply here.

massie: You can just make defaultCommandGroups the default value in the `ADAMMain` constructor, which will simplify this to `new ADAMMain()(args)`

massie:
With a little cleanup, this looks ready to go. What do you think @heuermh ?

heuermh:
Yep, thank you for the review, @massie

adam-cli/src/main/scala/org/bdgenomics/adam/cli/ADAMMain.scala
massie: Maybe something like, e.g.

```
class ADAMMain @Inject() (injectedGroups: List[CommandGroup] = null) {
   val commandGroups = match Option(injectedGroups) {
      case Some(groups) =>  //*** merge the new commands into the default group and return
      case None => defaultCommmandGroup
   }
 
   ...
}
```

massie: If the `@Inject` command supports `Option`, it would be nicer to have `class ADAMMain @Inject() (injectedGroups: Option[List[CommandGroup]]) {` or somesuch. I'll stop now.

heuermh:
Note even with code fixes as described above, this still won't be usable until this or one of the other options above is addressed
* Accept ```--main``` as a parameter in ```adam-submit```?  This approach plus the ```--jars``` argument is described in [adam-commands](https://github.com/heuermh/adam-commands).

adam-cli/src/main/scala/org/bdgenomics/adam/cli/ADAMMain.scala
heuermh: Injecting an optional doesn't really fit the dependency injection model.  What gets injected into the constructor shouldn't be mucked with after the fact.  Guice forbids injecting null by default.

massie:
Jenkins, test this please.

fnothaft:
@heuermh I want to test this on the cluster. Do you have an example project showing how I could inject a new command?

heuermh:
@fnothaft The example I've been using is here https://github.com/heuermh/adam-commands.

Note the patch regarding the main class argument in the readme, which needs to be updated after the bin scripts have changed.  It is soon on my list of things to go to create such a pull request.

See also previous comment https://github.com/bigdatagenomics/adam/pull/762#issuecomment-141169280

fnothaft:
Thanks @heuermh! I've verified that this works with #854 and your adam-commands example on the cluster. Merged as https://github.com/bigdatagenomics/adam/commit/58a4823cd1b59fb5ea2aa9faa1fec98d170653e7.

Issue #752, opened 2015-08-04
Bump Spark version to 1.4


laserson:
This issue will track any progress necessary for that.

See #750.
See #659.

laserson:
It seems `spark.kryoserializer.buffer.mb` has been deprecated.

laserson:
See #751 for a change that needs to be included.

heuermh:
To clarify, this issue is to raise the compile time dependency on spark to version 1.4?

laserson:
Correct

ryan-williams:
So, I realized something awkward about this that maybe others have processed but that I hadn't yet: there's `spark.version` in ADAM's POM, and then there's whatever spark version the user's `$SPARK_HOME` has, and they're pretty much independent.

The former is only relevant to the spark classes we link against, which AFAIK have not changed in a way we care about since `1.2`, and so bumping it or not, in isolation, shouldn't make any difference to anyone.

The latter informs what scripts (e.g. Spark's `bin/utils.sh`) are available vs. not; it definitely seems like high time to allow `$SPARK_HOME` to point at a Spark >= 1.4, but there's a question of how/whether to support `$SPARK_HOME` pointing at Sparks < 1.4, which is currently being discussed on #754, I guess?

heuermh:
Yep, the compile time dependency shouldn't matter, unless it does.  :)  E.g. if a binary incompatibility in 1.x versions of Spark or one of its transitive dependencies slips through.

Is #754 backward compatible script-wise with previous versions of Spark?  A few simple examples I tried worked for me on Spark 1.3.1.  It would be best if we didn't require ```$SPARK_HOME``` to be set at all, I think it was only required to find the ```utils.sh``` script.

ryan-williams:
Good point about transitive dependencies.

If the net result of #754 ends up being that we don't need `$SPARK_HOME` set and the ADAM scripts work against arbitrary Spark versions >= 1.2, then that seems like a great step forward. I'll try to keep following along there.

fnothaft:
The transitive dependency thing is both nasty and common unfortunately...

laserson:
`SPARK_HOME` is also needed to find the `spark-submit`, `spark-shell`, and `pyspark` scripts.

heuermh:
> `SPARK_HOME` is also needed to find the `spark-submit`, `spark-shell`, and `pyspark` scripts.

Would it be a fair assumption that those should be on the user's path?

laserson:
[...]

PR #765, opened 2015-08-10
[ADAM-752] Build for many combos of Spark/Hadoop versions.

fnothaft:
Resolves build comments on #752. I have set Jenkins up as a 3D!!!!!! matrix:

screen shot 2015-08-10 at 4 32 28 pm

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -351,6 +351,12 @@
         <version>${utils.version}</version>
         <type>test-jar</type>
         <scope>test</scope>
+        <exclusions>
+          <exclusion>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>*</artifactId>
+          </exclusion>
+        </exclusions>
       </dependency>
       <dependency>
         <groupId>org.bdgenomics.utils</groupId>
@@ -364,11 +370,7 @@
         <exclusions>
           <exclusion>
             <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-core</artifactId>
-          </exclusion>
-          <exclusion>
-            <groupId>com.fasterxml.jackson.core</groupId>
-            <artifactId>jackson-annotations</artifactId>
+            <artifactId>*</artifactId>
           </exclusion>
         </exclusions>
       </dependency>
@@ -477,12 +479,6 @@
         <groupId>org.slf4j</groupId>
         <artifactId>slf4j-log4j12</artifactId>
         <version>1.7.5</version>
-      </dependency>
-      <dependency>
-        <groupId>com.amazonaws</groupId>
-        <artifactId>aws-java-sdk</artifactId>
-        <version>1.7.5</version>
-        <scope>provided</scope>
       </dependency>
       <dependency>
         <groupId>org.apache.httpcomponents</groupId>
--- a/scripts/jenkins-test
+++ b/scripts/jenkins-test
@@ -6,11 +6,10 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 PROJECT_ROOT="$DIR/.."
 
 VERSION="$(grep "<version>" "$PROJECT_ROOT/pom.xml"  | head -2 | tail -1 | sed 's/ *<version>//g' | sed 's/<\/version>//g')"
-HADOOP_VERSION="$(grep "<hadoop.version>" "$PROJECT_ROOT/pom.xml"  | head -2 | tail -1 | sed 's/ *<hadoop.version>//g' | sed 's/<\/hadoop.version>//g')"
-echo "Testing ADAM version ${VERSION} on Hadoop ${HADOOP_VERSION}"
+echo "Testing ADAM version ${VERSION} on Spark ${SPARK_VERSION} and Hadoop ${HADOOP_VERSION}"
 
 export MAVEN_OPTS="-Xmx1536m -XX:MaxPermSize=1g"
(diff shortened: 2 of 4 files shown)


fnothaft:
OK, so a few interesting things.

* We can't build for Spark 1.1.1, which I had forgotten (`repartitionAndSortWithinPartitions` doesn't exist)
* It seems like the Spark-based unit tests will only run in parallel on Spark 1.2.0. Not sure what is broken in Spark 1.3+? I've told Jenkins to just run builds sequentially, for now.

scripts/jenkins-test
```diff
@@ -26,16 +25,24 @@ export SPARK_DRIVER_MEMORY=8g
 
 pushd "$ADAM_TMP_DIR"
 
+
 if [[ $HADOOP_VERSION == "1.0.4" ]]; then
```
massie: Maybe `echo $HADOOP_VERSION | egrep '^1.0.'` would be more general, in case, we use e.g. `1.0.4` as the hadoop test version?

ryan-williams:
That looks pretty neat. I'm guessing there's some config changes you are making in Jenkins that are not depicted here that e.g. populate `$SPARK_VERSION`?

scripts/jenkins-test
```diff
@@ -26,16 +25,24 @@ export SPARK_DRIVER_MEMORY=8g
...
-    wget -q http://d3kbcqa49mib13.cloudfront.net/spark-1.1.0-bin-hadoop1.tgz
-    tar xzvf spark-1.1.0-bin-hadoop1.tgz
-    export SPARK_HOME="${ADAM_TMP_DIR}/spark-1.1.0-bin-hadoop1"
+    HADOOP=hadoop1
+elif [[ $HADOOP_VERSION == "2.6.0" ]]; then
```
massie: Maybe `echo $HADOOP_VERSION | egrep '^2.6.'` would be more general, in case, we use e.g. `2.6.3` as the hadoop test version?

scripts/jenkins-test
```diff
@@ -26,16 +25,24 @@ export SPARK_DRIVER_MEMORY=8g
...
+    HADOOP=hadoop2.6
+elif [[ $HADOOP_VERSION == "2.3.0" ]]; then
```
massie: etc

massie:
This is good stuff, Frank.

fnothaft:
Jenkins, retest this please.

scripts/jenkins-test
ryan-williams: or possibly `[[ $HADOOP_VERSION =~ ^2\.6 ]]`, likewise below, if that seems easier

massie: What @ryan-williams said. :) Much cleaner.

fnothaft:
Fun times! Guess what doesn't work?

If your guess was that running:

mvn package -Dspark.version=1.4.1

Would lead to Destruction, Terror, and Mayhem, you win $5!

I will sort this out _later_, but if anyone has seen this before or has any clues, I would love your thoughts.

heuermh:
Destruction, Terror, and Mayhem and a 45 minute wait to see if your pull request turns green.  :)

Maybe a middle ground, or set up a quick build in Travis and leave Jenkins 3D Awesome for a nightly integration test?

Would a profile work for where the ```-Dspark-version``` is failing?

fnothaft:
OK, so what seems to have been the problem is that we had a completely unused (???) dependency on `com.amazonaws:aws-java-sdk`, which was bringing in a version of `com.fasterxml.jackson.core:jackson-core` that was incompatible with Spark 1.3.0 and higher. As a result, this was giving a red herring test failure message that implied that the unit tests were crashing because we were running multiple `SparkContext`s in parallel. That unused dependency is gone now, so hopefully the build should pass now (and we don't need to run all the builds sequentially, which is good for obvious reasons).

fnothaft:
And the Spark 1.4.1/Hadoop 2.6.0 touchstone builds have passed! Huzzah! Now, for the rest of the builds.

fnothaft:
Rebased and added a commit to clean up log junk when running 1.4.1 unit tests.

fnothaft:
Cleaned up RE: the comments above around version checking.

fnothaft:
Jenkins, retest this please.

pom.xml
```diff
@@ -345,12 +345,36 @@
         <type>test-jar</type>
         <scope>test</scope>
       </dependency>
+      <!--
```
ryan-williams: any reason for this block?

fnothaft: I was testing something; this can be removed.

fnothaft:
Jenkins, retest this please.

fnothaft:
IT WORKS! IT WORKS! IT REALLY DOES!!!!!!!

heuermh:
Are all the red Failed - skipped in build 842 supposed to be Not run?  Looks like they might be combinations that are not supported, such as Hadoop 1 with Spark 1.4.1.  Or maybe I just need my 3D glasses?

fnothaft:
@heuermh correct; Spark 1.4.1/Hadoop 1.x is skipped as it won't build with the current way that Spark is packaged as maven artifacts (there was a long discussion of this on a previous ADAM pr).

fnothaft:
They show up as greyed out red because the _last_ build of that combo failed.

heuermh:
From what I can see it looks like a 15 minute build time up from 10 minutes, not too bad.  +1

massie:
Nice addition, Frank!

PR #773, opened 2015-08-12
[ADAM-772] fix some bash var quoting

### LINKED ISSUE #772: Unquoted subshell result, incorrect exit code in compute-adam-jars.sh
On a clean repo:

```sh
$ bin/compute-adam-jars.sh
You need to build ADAM before running this program
$ echo $?
0
```

That "You need to build…" [should have been preceded by another `echo`](https://github.com/bigdatagenomics/adam/blob/6cade81d97eccfbad4b9630a4c7673ff3f27973e/bin/compute-adam-classpath.sh#L36-L37); also: exit code `0`?

[This is the culprit](https://github.com/bigdatagenomics/adam/blob/6cade81d97eccfbad4b9630a4c7673ff3f27973e/bin/compute-adam-jars.sh#L23): unquoted sub-shell call.

ryan-williams:
fixes #772 

after:

```sh
$ bin/compute-adam-jars.sh
Failed to find appassembler scripts in /bin
You need to build ADAM before running this program
$ echo $?
1
```

### DIFF
--- a/bin/adam-submit
+++ b/bin/adam-submit
@@ -54,6 +54,8 @@ SCRIPT_DIR="$(cd `dirname $0`/..; pwd)"
 
 # Get list of required jars for ADAM
 ADAM_JARS=$("$SCRIPT_DIR"/bin/compute-adam-jars.sh)
+
+# Split out the CLI jar, since it will be passed to Spark as the "primary resource".
 ADAM_CLI_JAR=${ADAM_JARS##*,}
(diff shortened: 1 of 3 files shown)


fnothaft:
Thanks @ryan-williams! I've manually committed this as https://github.com/bigdatagenomics/adam/commit/e7bb9670fff41286f6198219b3ab41a2d1823a13.

PR #778, opened 2015-08-19
[ADAM-752] update spark compile dependency to version 1.4.1

heuermh:


ryan-williams:
Thanks @heuermh; how confident are we that jenkins passing means we don't think there are notable transitive-dep concerns? I'm inclined to trust it and try to add tests if we discover issues we hadn't anticipated.

heuermh:
I feel pretty good about the Jenkins tests, since it not only runs unit tests but calls ```adam-submit``` (commands ```transform```, ```transform -sort```, ```reads2ref```, ```print```, and ```flagstat```).

heuermh:
Woot!  Thank you, Mr Jenkins, sir.

fnothaft:
LGTM!

PR #782, opened 2015-08-20
Fix SAM Attribute parser for numeric array tags

jfeala:
Page 6 of the SAMv1 specs allow for numeric array tags (TYPE="B"), with values matching:
```
[cCsSiIf](,[-+]?[0-9]*\.?[0-9]+([eE][-+]?[0-9]+)?)+
```
This update of `org.bdgenomics.adam.util.AttributeUtils.typedStringToValue` skips the first character (a numeric type specifier) plus the leading comma, before trying to convert to a numeric array. This bug was breaking `loadAlignments` for BAM files generated by the STAR RNA-Seq aligner, which creates type "B" tags by default.


fnothaft:
Jenkins, add to whitelist.

fnothaft:
Thanks @jfeala! I have merged this manually.

Issue #760, opened 2015-08-07
Conversion to BAM requires post conversion reordering and sorting


beaunorgeot:
Reference chromosome ordering and read sorting information are lost when converting an ADAM file that has undergone MD,Indel, & BQSR back to BAM. 

This means that the resulting BAM must be Reordered and Sorted prior to use with another tool (like a variant caller), which seems unnecessary/wasteful.

massie:
Did you try this with the `-sort_reads` option in `transform`? That should result in ADAM emitting sorted/reordered files.

beaunorgeot:
The -sort_reads option in transform does not result in ADAM preserving bam ordering. 

The order of the bam file appears completely random:
samtools view -H ADAMcat.bam ##first couple of lines of output bam. These aren't ordered at all

@SQ	SN:Y	LN:59373566	M5:Y
@SQ	SN:14	LN:107349540	M5:14
@SQ	SN:9	LN:141213431	M5:9
@SQ	SN:7	LN:159138663	M5:7
@SQ	SN:17	LN:81195210	M5:17
@SQ	SN:18	LN:78077248	M5:18
@SQ	SN:8	LN:146364022	M5:8


samtools view -H input.bam ## first few lines. These are ordered correctly
@SQ	SN:1	LN:249250621
@SQ	SN:2	LN:243199373
@SQ	SN:3	LN:198022430
@SQ	SN:4	LN:191154276
@SQ	SN:5	LN:180915260
@SQ	SN:6	LN:171115067

ryan-williams:
Sounds like `-sort_reads` sorts the reads but @beaunorgeot is commenting on the header lines not being sorted?

fnothaft:
It looks like according to the [SAM spec](https://samtools.github.io/hts-specs/SAMv1.pdf) the order of the `@SQ` lines defines the sort order of the file. @beaunorgeot would it be correct to say that in a BAM file outputted from ADAM, the _reads_ are sorted, but the read sort order disagrees with the `@SQ` line sort order?

beaunorgeot:
As we discussed in the group call today, the reads presumably are sorted. However, when ADAM outputted BAMs are re-concatenated into a single file, the resulting file has an apparently random chromosome order and (for what is probably the same reason) the overall reads are unsorted. This outcome seems consistent w/partition numbers not correlating directly to coordinate ordering.

fnothaft:
Working on this.

fnothaft:
[...]

PR #784, opened 2015-08-20
[ADAM-783] Write @SQ header lines in sorted order.

### LINKED ISSUE #783: Sequence dictionary is written in incorrect order
As discussed on #760, one of the reasons that files that we save in SAM/BAM need to be resorted is because we write the sequence dictionary in arbitrary order. I'm breaking this out from the rest of that ticket, since this may just be part of the resolution of that issue.

fnothaft:
This change resolves #783 and #760. Specifically, now we write the SAM/BAM @SQ header
lines using the same lexicographic ordering that we use for sorting records, and we write the @HD line to note that we are sorted in coordinate order.

### DIFF
--- a/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
+++ b/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
@@ -158,7 +158,7 @@ class Transform(protected val args: TransformArgs) extends BDGSparkCommand[Trans
       } else {
         sc.loadAlignments(args.inputPath)
       }
-    }).adamSave(args)
+    }).adamSave(args, args.sortReads)
   }
 
   private def createKnownSnpsTable(sc: SparkContext): SnpTable = CreateKnownSnpsTable.time {
--- a/adam-core/src/main/scala/org/bdgenomics/adam/models/SequenceDictionary.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/models/SequenceDictionary.scala
@@ -109,11 +109,19 @@ class SequenceDictionary(val records: Vector[SequenceRecord]) extends Serializab
    * @return Returns a SAM formatted sequence dictionary.
    */
   def toSAMSequenceDictionary: SAMSequenceDictionary = {
-    new SAMSequenceDictionary(records.map(_ toSAMSequenceRecord).toList)
+    import SequenceRecord._
+    new SAMSequenceDictionary(records.sorted.map(_ toSAMSequenceRecord).toList)
   }
 
   override def toString: String = {
     records.map(_.toString).fold("SequenceDictionary{")(_ + "\n" + _) + "}"
+  }
+}
+
+object SequenceOrdering extends Ordering[SequenceRecord] {
+  def compare(a: SequenceRecord,
+              b: SequenceRecord): Int = {
+    a.name.compareTo(b.name)
   }
 }
 
@@ -182,6 +190,8 @@ class SequenceRecord(
 object SequenceRecord {
   val REFSEQ_TAG = "REFSEQ"
   val GENBANK_TAG = "GENBANK"
+
+  implicit def ordering = SequenceOrdering
 
   def apply(name: String,
             length: Long,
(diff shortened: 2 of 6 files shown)


ryan-williams:
lgtm from a quick pass

fnothaft:
Whoops! Forgot to add the test collateral...

fnothaft:
Rebased.

fnothaft:
Rerebased.

fnothaft:
This now resolves #760 as well. Can I get review/merge?

fnothaft:
Ping on review/merge... This is the last issue pending for the 0.17.1 release.

ryan-williams:
looking

fnothaft:
Thanks @ryan-williams!

ryan-williams:
This lgtm; one question: this optionally sorts the header lines to match the sort of the reads that can be optionally done with `adamSortReadsByReferencePosition`, right?

That seems reasonable, but feels a little backwards, since [the spec](https://samtools.github.io/hts-specs/SAMv1.pdf) says that the `@SQ` order should define the read order.

If ADAM will only ever output a lex-sort of the `@SQ`s, and the corresponding "coordinate" sort of the reads, that's fine. I have a few more thoughts about making `transform` round-trips identical byte-for-byte, but I'll put that in a separate issue.

ryan-williams:
Finally, I don't know what the preferred way of merging is these days :)

fnothaft:
Thanks for the review!

@ryan-williams for now, we're just emitting coordinate sorted order. The other orders are [defined here](https://samtools.github.io/htsjdk/javadoc/htsjdk/) and are `duplicate` and `queryname`. Neither of those seem to be defined anywhere (like, you know, the SAM spec), but a queryname imp'l is at https://github.com/samtools/htsjdk/blob/master/src/java/htsjdk/samtools/SAMRecordQueryNameComparator.java and a duplicate imp'l is at https://github.com/samtools/htsjdk/blob/master/src/java/htsjdk/samtools/SAMRecordDuplicateComparator.java. I believe queryname is just sorting lexicographically by read name.

I agree it is a bit backwards, but I think an equivalent way to read it is that you need to have the same lexicographic order for both the reads and the header.

We can continue using the merge button for now! After I've got scripts ready, we can cut over.

ryan-williams:
> The other orders are defined here and are duplicate and queryname. Neither of those seem to be defined anywhere (like, you know, the SAM spec)

FWIW, `queryname` is in the SAM spec, and is what you described:



`duplicate` is not, and there's [this note about it not being in the spec in htsjdk](https://github.com/samtools/htsjdk/blob/1.133/src/java/htsjdk/samtools/SAMFileHeader.java#L71).

My point, here and in #794, is that the SAM spec doesn't actually say that `@SQ` should be sorted anywhere, just that "The order of @SQ lines defines the alignment sorting order". `coordinate` sort of the reads is then defined not lexicographically based on the `@SQ` name, but by whatever order the `@SQ`s are already in, with `POS` as a secondary sort.

Arguably we should be able to leave `@SQ`s in whatever order they come to us in (and sort reads to conform to that) independently of whether we also want to lex-sort the `@SQ`s (and therefore also lex-sort the reads by `RNAME`).

Anyway, unless this sounds like such a good idea to you that you want to do it here, I'm find to merge this and then address the possibility of `@SQ`s in non-lex order (and sorting reads to match them) in #794.

fnothaft:
Rebased.

ryan-williams:
k, i'll merge when test passes

fnothaft:
@ryan-williams doesn't that snippet just specify what the `coordinate` sort order is?

I get what you're saying. I am personally OK with rewriting the `@SQ` lines to lexicographic order. We could have a different sorting approach that would take in the full sequence dictionary, but it would be a bit more difficult to implement and would probably be a bit slower. Let's look into this more as part of #794.

ryan-williams:
> doesn't that snippet just specify what the coordinate sort order is?
ah, you mean that they don't explicitly say what `queryname` sort is? that makes sense. I guess we're all just to assume that they mean lex-sort by `QNAME`.

Yea, we can discuss further on #794; always sorting `@SQ`s, and then also sorting reads, seems reasonable, but the reads' side of it should maybe be implemented in terms of the `@SQ` order. We sort of backed into this "always lex-sort `@SQ`s" approach after having been non-spec-compliant.

Issue #741, opened 2015-07-24
Allow transform to recursively search for BAMs in a directory and convert them all to ADAM


ryan-williams:
simply swapping `s/\.bam$/.adam/` on the file names.

I feel like I/we end up not converting stuff to ADAM because I have a mental block about doing it file by file

fnothaft:
Would you intend this to write out "one" parquet file (a directory of parquet files that was all of the samples), or to convert each .bam into an individual .adam parquet file? To rephrase, if you started with 100 BAM files, would you end with 100 ADAM files or 1 ADAM file?

ryan-williams:
ah, I was imagining 100 ADAM files, all dropped in place next to their respective BAMs

laserson:
This seems like a departure from the typical Hadoopy semantics.  If I give a Hadoop engine an input path in HDFS, I typically expect one output path to go along with it.

massie:
I think what Ryan is looking for here (correct me if I'm wrong) is something like a recursive `sync` option. A Hadoopy option here might be to specify a `bam_directory` and `adam_directory` and have a job that checks which BAM files in the `bam_directory` haven't been converted and convert them to ADAM using the `adam_directory` as the root of the output. Agree with @laserson here that mixing the file types together together would complicate things. We would, for example, need to filter out all the BAM files on input (which is possible but not ideal).

ryan-williams:
I'm open to peoples' preferences about what the directory structure should be, though a default of dropping an `.adam` next to any `.bam` feels really sane and useful to me.

The most common pattern at HammerLab is to do this, similarly to how other things in bioinformatics (e.g. `.bai`s) are often associated with their "parent"/"source" files by living next to them in a directory and sharing a basic `basename` structure, with a different extension.

So, most of the ADAM `transform` commands I run look like this:

```
bin/adam-submit transform /path/to/foo.{b,ad}am
```

I'm "this close" to writing my own crappy script to do this recursively in an HDFS directory, to all the directories or files ending in `.bam` that it finds, but I would much prefer ADAM to do it for me, not least of all because I imagine it would be able to benefit greatly from parallelizing N such `transform`ations in one Spark application instead of launching one Spark application per `.bam`.

[...]

PR #790, opened 2015-08-21
optionally add MDTags to reads with `transform`

### LINKED ISSUE #738: ADAM command for adding MDTags to AlignmentRecords
This would be a potentially useful function; @timodonnell was just lamenting that he doesn't know an easy/quick way of doing it.

ryan-williams:
fixes #738 

The entire reference is collected on to the driver and broadcast to workers.

I thought about doing some fancier region-join stuff, but then just decided to do a first pass this way.

I tested it on [`bqsr.sam` from this repo](https://github.com/bigdatagenomics/adam/blob/207eebaf3dbf6f01e5b8398b2af388ee70c862aa/adam-cli/src/test/resources/bqsr1.sam) as follows:

```
wget ftp://ftp.ncbi.nlm.nih.gov/genbank/genomes/Eukaryotes/vertebrates_mammals/Homo_sapiens/GRCh37/Primary_Assembly/assembled_chromosomes/FASTA/chr22.fa.gz
gunzip chr22.fa.gz

# rename the contig to "22" to match bqsr.sam
cat <(head -n 1 chr22.fa | pe 's/^>/>22 /') <(tail -n +2 chr22.fa) > chr22.fa.tmp
mv chr22.fa{.tmp,}

# Make a version of bqsr.sam with no MD tags
perl -pe 's/\tMD:Z:[^\s]+//' adam-cli/src/test/resources/bqsr1.sam > adam-cli/src/test/resources/bqsr1.stripped.sam

# Run transform! input: "stripped" .sam with no MDTags; output: "restored" .sam with MDTags.
bin/adam-submit --driver-memory 4g --conf spark.kryoserializer.buffer.max=256m -- transform adam-cli/src/test/resources/bqsr1.stripped.sam -add_md_tags chr22.fa -single adam-cli/src/test/resources/bqsr1.restored.sam

# This exits 0; the MD tags in the original and "restored" files are identical.
diff -q <(pcregrep -o 'MD:Z:[^\s]+' adam-cli/src/test/resources/bqsr1.sam) <(pcregrep -o 'MD:Z:[^\s]+' adam-cli/src/test/resources/bqsr1.restored.sam)
```


fnothaft:
Nice! Would you like to get this in to 0.17.1? I will make a review pass now.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala
```diff
@@ -19,7 +19,7 @@ package org.bdgenomics.adam.rdd.read
 
 import java.io.StringWriter
 
-import htsjdk.samtools.{ SAMFileHeader, SAMTextHeaderCodec, SAMTextWriter, ValidationStringency }
+import htsjdk.samtools.{ TextCigarCodec, SAMFileHeader, SAMTextHeaderCodec, ValidationStringency, SAMTextWriter }
```
fnothaft: Small nit: at this point, might it not make sense to just do `import htsjdk.samtools._`? I think the general guideline we've been using is to import the whole package if we are using more than 4 items from the package.

ryan-williams: haha funny you should ask; I [just complained to JetBrains](https://youtrack.jetbrains.com/issue/IDEA-144185) about IntelliJ reverting my disabling of the wildcard-imports when it exceeds 4, the preferences-search making it hard to find that setting, etc.

I'm fine to wildcard it if that's what you prefer.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala
```diff
@@ -285,6 +285,77 @@ class AlignmentRecordRDDFunctions(rdd: RDD[AlignmentRecord])
...
+      fragmentLengthOpt match {
+        case Some(fragmentLength) =>
+          sc.loadSequence(referenceFile, fragmentLength = fragmentLength)
+        case _ =>
+          sc.loadSequence(referenceFile)
+      }
+    )
+
+  def adamAddMDTags(referenceFragments: RDD[NucleotideContigFragment]): RDD[AlignmentRecord] = {
+    val collectedRefMap =
```
fnothaft: As future cleanup, we might want to make this object (the collectedRefMap) extend [ReferenceFile](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/util/ReferenceFile.scala), and then change `getRefSeq` to `ReferenceFile.extract`. Then we could use this with both 2Bit files and etc.

ryan-williams:
Thanks @fnothaft, no particular rush; this impl may be too inefficient for prime-time.

I updated the description with some info about a one-contig test I did locally, fjyi.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala
```diff
@@ -285,6 +285,77 @@ class AlignmentRecordRDDFunctions(rdd: RDD[AlignmentRecord])
...
+      (for {
+        contig <- Option(read.getContig)
+        contigName <- Option(contig.getContigName)
+        if read.getReadMapped
+        referenceStart = read.getStart
+        referenceLength = read.referenceLength
+        referenceEnd = referenceStart + referenceLength
+        refSeq = getRefSeq(contigName, referenceStart, referenceEnd)
+        cigar = TextCigarCodec.decode(read.getCigar)
+        mdTag = MdTag(read.getSequence, refSeq, cigar, referenceStart)
```
fnothaft: Nice! This makes it _way_ simpler than I'd originally thought it'd be.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala
fnothaft: I actually agree with you RE: not wildcard importing packages. My personal preference would be to newline it at the commas though, and to alphabetize the classes that are imported (I prefer to keep the imports alphabetized, because it generally keeps diffs smaller).

ryan-williams:
Hm:

```
[ERROR] /home/jenkins/workspace/ADAM-prb/HADOOP_VERSION/2.3.0/SCALAVER/2.10/SPARK_VERSION/1.2.1/label/centos/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala:310: error: could not find implicit value for parameter param: org.apache.spark.AccumulatorParam[Long]
[ERROR]     val mdTagsAdded = sc.accumulator(0L, "MDTags Added")
[ERROR]                                     ^
```

Seems to uniformly pass on Spark > 1.2 and fail on <= 1.2:

fnothaft:
What does the performance look like on the single contig? Personally, I would expect it to be pretty reasonable. The MDTag code is the same code that we rely on in the INDEL realigner, which is fairly zippy.

fnothaft:
The `AccumulatorParam[Long]` definitely exists in 1.2.0...: https://spark.apache.org/docs/1.2.0/api/scala/index.html#org.apache.spark.SparkContext$$LongAccumulatorParam$

ryan-williams:
I guess by "performance" I meant:
* the driver needs enough memory to collect the full reference
* we're broadcasting the full reference to all nodes
* I seemingly had to increase my `spark.kryoserializer.buffer.max` to run on chr22, which is ~50m as a FASTA (probably more as `NucleotideContigFragment`s.
  * It defaults to `64m`, I set it to `256m` as a guess and that worked.
  * I'm worried that I'd have to make that large enough for the whole reference...
  * Will we run into the "blocks must be <2GB" issue? [SPARK-6190](https://issues.apache.org/jira/browse/SPARK-6190)

ryan-williams:
Thoughts on how I should quantify/measure the performance? I can try to run it on a whole BAM on a cluster, and look at some of the Spark-reported numbers, but I'm not sure that will cover some of the things I'm really curious about.

fnothaft:
> I guess by "performance" I meant:
>
> * the driver needs enough memory to collect the full reference

Ah, gotcha. That's part of what I was thinking RE: refactoring around the `ReferenceFile` trait, since we could broadcast 2Bit files, which are fairly compact in memory (at the least, @laserson and I have both successfully used whole genome 2BIT files as broadcast objects; for hg19, the 2bit file is ~800MB), at the possible cost of increased region extraction costs.

FWIW, we have a few fairly large broadcast objects in the ADAM codebase (e.g., for BQSR) so adding another large broadcast doesn't trouble me too much.

> Thoughts on how I should quantify/measure the performance? I can try to run it on a whole BAM on a cluster, and look at some of the Spark-reported numbers, but I'm not sure that will cover some of the things I'm really curious about.

I was personally just looking for ballpark numbers. I do agree that the whole genome broadcast object is most likely to be the sticking point, so running that test seems reasonable.

Actually, I've got the high coverage NA12878 sitting on a cluster, along with the 1000G custom reference. I'll build and run this.

ryan-williams:
### Accumulators
I feel like it has to do with the `implicit` `LongAccumulatorParam` object, [which showed up in 1.3.0](https://github.com/apache/spark/blob/v1.3.0/core/src/main/scala/org/apache/spark/Accumulators.scala#L267-L270) but [was not in 1.2.1](https://github.com/apache/spark/blob/v1.2.1/core/src/main/scala/org/apache/spark/Accumulators.scala).

I'm not totally clear on why that is necessary, but I'll dig; let me know if you have thoughts about it.

### BroadcastRegionJoin

I just realized that I basically reimplemented `BroadcastRegionJoin` here; I'm going to look into reworking this to use that. lmk if that doesn't make sense.

I'm also going to look into an implementation that does a `ShuffleRegionJoin`, I have decided on an approach that should work for that.

### Testing on WGS
Is it easy for you to point me at the NA12878 and reference you speak of?

Should [eggo](https://github.com/bigdatagenomics/eggo) be useful to us here? cc @laserson

fnothaft:
> Testing on WGS
>
> Is it easy for you to point me at the NA12878 and reference you speak of?

For some _blasted_ reason, 1000G pulled down the high coverage WGS copy of NA12878. _Howaaaaaayver_, UCSC has a copy of it up at https://s3-us-west-2.amazonaws.com/bd2k-test-data/NA12878.mapped.ILLUMINA.bwa.CEU.high_coverage_pcr_free.20130906.bam. The reference that file is aligned to is ftp://ftp-trace.ncbi.nih.gov/1000genomes/ftp/technical/reference/human_g1k_v37.fasta.gz.

RE: region join, I wouldn't worry about that one too much for now. My thought is that it'd probably be preferable to reimplement vs. the `ReferenceFile` interface instead of the `RegionJoin` interface. cc @laserson for thoughts.

fnothaft:
It looks like this works on the WGS data! I did get an exception, but that was unrelated to the MD tag code; the reference file that I provided was missing the decoy contig...

Anywho, this took about 1 hr to run on 16 16 core machines on Spark 1.3.1 with the following Spark options:

```
--master yarn-cluster --num-executors 1024 --executor-memory 11g --executor-cores 1 --driver-memory 14g --conf spark.yarn.executor.memoryOverhead=3072 --conf spark.kryoserializer.buffer.max.mb=1024 --conf spark.driver.maxResultSize=0
```

I'm running again with the decoy contigs from [here](ftp://ftp.1000genomes.ebi.ac.uk/vol1/ftp/technical/reference/phase2_reference_assembly_sequence/hs37d5.fa.gz) and will report back in the AM.

ryan-williams:
Wow, cool, thanks Frank. How big is that BAM, if you know? I'll check tmrw
of not, nbd.

fnothaft:
BAM is 234GB --> approx. 185GB in ADAM.

fnothaft:
Completed in approx 2.1 hrs:

screen shot 2015-08-21 at 6 15 08 am

ryan-williams:
Just pushed a commit that should fix this for Spark 1.2.1.

ryan-williams:
I'm going to add a couple more things here:
* Option to overwrite existing MDTags (e.g. when you think they may have been incorrectly generated; @arahuja just ran into this).
* Option to use `BroadcastRegionJoin` vs. `ShuffleRegionJoin`, with cmdline flag.
* Option to control the fragment length from the cmdline.

fnothaft:
OOC, do you think it might make sense to break this into a separate command?

ryan-williams:
Another weird thing I saw when I tested it is that the accumulator values were exactly 3x what they should have been; I had 1000 reads in bqsr.sam, 992 mapped and 8 unmapped, and the accumulators showed up as 2976 "MDTags Added", 24 "Unmapped Reads". Can you spot-check the accumulators' values in your test runs @fnothaft? You should just be able to go to the "saveAsNewAPIHadoopFile at AlignmentRecordRDDFunctions.scala:179" stage page in Spark UI or Spree :)

ryan-williams:
re: separate command: maybe!

I was meaning to ask you, in reference to #741, whether you think we are moving toward or away from having commands for `bam2adam`, `adam2bam`, `vcf2adam`, `adam2vcf`, `fasta2adam`, `adam2fasta`. Not necessarily the same thing, but somewhat related to how much we want `transform` to be a kitchen-sink.

fnothaft:
Looks like we've got 100% concordance on the big BAM file, with the important aside that it doesn't seem to have MD tags on a large number of the reads.

ryan-williams:
How are you checking that OOC? In the description, I showed a pretty hacky thing I did in the shell to just compare the `MD:Z:` lines of two files.

fnothaft:
@ryan-williams yeah, I'm not sure. The `transform` kitchen sink is admittedly kind of nice...

fnothaft:
```
val rdd1 = sc.loadAlignments("hdfs://amp-bdg-master:8020/user/fnothaft/NA12878_high_coverage.adam")
val rdd2 = sc.loadAlignments("hdfs://amp-bdg-master:8020/user/fnothaft/NA12878_high_coverage.2.adam")
val zRdd = rdd1.zip(rdd2)
zRdd.map(p => { if (p._1.getMismatchingPositions != null && p._1.getMismatchingPositions == p._2.getMismatchingPositions) (1, 1) else (1, 0)}).reduce((v1: (Int, Int), v2: (Int, Int)) => (v1._1 + v2._1, v1._2 + v2._2))
```

ryan-williams:
Nice!

Any word on the accumulator values you're seeing?

fnothaft:
> Any word on the accumulator values you're seeing?

Ah, I didn't note those when I ran. Do you print those to the log? I'm in a weird network situation right now and can't get to my logs, but I could check those later...

ryan-williams:
They end up in the eventlog, if you're writing that; you can fire up a Spark history server that reads it and go to the stage page.

You could also just give me the event log and I'll look.

fnothaft:
I believe I should have that. I don't have access to the history server right now (aforementioned weird network connectivity issues...) but will check it when I am able to get access back.

ryan-williams:
Just pushed a big refactor. Still going to work on the `ReferenceFile` stuff, but here's what's new:

* can do a shuffle- or broadcast-based join.
  * haven't tested these head-to-head yet, but will do shortly
* pulled code out to `MDTagging.scala`
* changed the `-bqsr_strict` option to just `-stringency`, so that it can be reused across different `transform` code paths (e.g. for MDTagging here).
* added a `AlignmentRecord => Option[ReferenceRegion]` helper, `ReferenceRegion.opt`; used it in a few places.

ryan-williams:
jenkins, retest this please

ryan-williams:
I forget if I'm capable of telling Jenkins that, whether it will do it on its own in this case because I've pushed new code, etc.

ryan-williams:
jenkins, retest this please

ryan-williams:
Jenkins, retest this please

ryan-williams:
btw: I figured out why my accumulator values were 3x what they should have been by grepping for them in my event log: I was not caching the RDD where they were computed, and three different stages were computing that RDD. A properly-placed `.cache` fixes it.

ryan-williams:
jenkins, retest this please

ryan-williams:
jenkins, retest this please

ryan-williams:
OK, I took out the shuffle impl and shored up a broadcast impl around a `ReferenceFile` interface. I think it's a pretty good abstraction and we now have N=2 implementations of it so the API is starting to make sense.

I had to add kryo ser/de for `TwoBitFile`.

FWIW, here are Spree screengrabs of the previous broadcast- and shuffle impls on bqsr1.sam:

Broadcast:


Shuffle:


The shuffle impl does an extra ~400MB of shuffle read+write, which is interesting. Given that these data sets will basically always be small enough to broadcast, I'm happy with just leaving that impl in.

Here are /jobs pages screenshots of the above, as well:

Broadcast:


Shuffle:


Pretty amusing that we shuffle 400MB during the ShuffleRegionJoin; this test was on a 50MB FASTA file and the 1000-read bqsr1.sam.

Anyway, I think this PR is ready to go from my end.

ryan-williams:
Ah, one thing I should maybe still do is [let it read `.2bit`s from S3](https://github.com/bigdatagenomics/adam/commit/b5ab6ed19873c2ed3d61a1f8199b2f349ec7db52#diff-d36ea7d0742decd0b040a73a96af06e9R526); I don't know how to do that otomh, pointers welcome!

ryan-williams:
Retest this please

ryan-williams:
Jenkins, retest this please

ryan-williams:
jenkins, retest this please

ryan-williams:
jenkins, retest this please

massie:
This looks ready to merge to me. What you do you think @fnothaft? I'm happy to do this manually myself, if you think it's ready to go.

fnothaft:
This is good to go from my side. @ryan-williams is this ready for merge, or do you have remaining work?

ryan-williams:
Yea I just got epically blocked on the probably tiny task of figuring out whether I have to do something special to load files from s3 [here](https://github.com/ryan-williams/adam/commit/1f797ee92dada67ca27ee11b7c5936fbec18df04#diff-d36ea7d0742decd0b040a73a96af06e9R545); I'll defer to y'all on whether that's important for this PR or not!

fnothaft:
Oh, hah! I would say let's not worry about that too much for now. We can add it in a follow-on PR.

fnothaft:
@ryan-williams Can you rebase this to clean up the merge conflicts? I would like to get this into 0.18.0.

ryan-williams:
Working on this now

ryan-williams:
retest this please

fnothaft:
Thanks @ryan-williams! I've squashed this down and merged as https://github.com/bigdatagenomics/adam/commit/850acdeb5b81cdddd13c1edb06004eb98271823e.

PR #792, opened 2015-08-21
[ADAM-791] Add repartition parameter to Fasta2ADAM.

### LINKED ISSUE #791: FASTA2ADAM doesn't have repartition parameter
This is a simple thing that we should add... allows someone to set the number of partitions they want their contigs to be written out to.

fnothaft:
Resolves #791. This allows someone who is converting a FASTA file into ADAM to set the number of output partitions they would like.


ryan-williams:
lgtm!

PR #798, opened 2015-08-22
fix SAM-attr conversion bug; int[]'s not byte[]'s

### LINKED ISSUE #785: loadAlignments does not properly convert "TYPE=B" tags
The SAM specification allows numeric array tags (TYPE=B), which are not often used but are created by the STAR RNA-seq aligner.

I thought I resolves this with https://github.com/bigdatagenomics/adam/pull/782, but the problem is a bit more complicated. The code I fixed (https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/util/AttributeUtils.scala#L96) is not actually used by `AdamContext.loadBam`.

Instead, the problem is that `htsjdk.samtools.SAMRecord.SAMTagAndValue` correctly parses the tag string, but there is a problem casting the Java object to an array at https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/util/AttributeUtils.scala#L38 and it throws the following exception:

```
java.lang.ClassCastException: [I cannot be cast to [Ljava.lang.Byte;
	at org.bdgenomics.adam.util.AttributeUtils$.convertSAMTagAndValue(AttributeUtils.scala:38)
	at org.bdgenomics.adam.converters.SAMRecordConverter$$anonfun$convert$2.apply(SAMRecordConverter.scala:178)
	at org.bdgenomics.adam.converters.SAMRecordConverter$$anonfun$convert$2.apply(SAMRecordConverter.scala:174)
	at scala.collection.Iterator$class.foreach(Iterator.scala:727)
	at scala.collection.AbstractIterator.foreach(Iterator.scala:1157)
	at scala.collection.IterableLike$class.foreach(IterableLike.scala:72)
	at scala.collection.AbstractIterable.foreach(Iterable.scala:54)
	at org.bdgenomics.adam.converters.SAMRecordConverter.convert(SAMRecordConverter.scala:173)
	at org.bdgenomics.adam.rdd.ADAMContext$$anonfun$loadBam$1.apply(ADAMContext.scala:268)
```

I tried to fix it myself, but I'm not a Scala programmer, and casting a Java Object into a parameterized type was sending me down a StackOverflow rabbit hole. Here is an example alignment (it's the jM and jI tags that are causing the problem):
```
C404RANXX150222:1:1210:7660:76662	163	chr1	4259695	255	21S55M	=	4259874	246	TGGGTTGAGATCAGGCCCTCAATCTCCTCGATGGTATGGACGATGAACATGTCCTGGAGGTCCTCCATGGCGCTCT	BB//BF<//<<FFB<FFF//<FFF/FFF//////</</BFFFFF/B/FBF/FFF//FFF//</F/<<B<B<<//<B	NH:i:1	HI:i:1	AS:i:100	nM:i:10	NM:i:3	MD:Z:19T15C14A4	jM:B:c,-1	jI:B:i,-1
```
As a temporary fix, the file can be loaded if the jI and jM tags are removed, but STAR is becoming more and more popular as an RNA-Seq aligner so people might start running into this more often.

ryan-williams:
fixes #785 

[per discussion on email list](https://groups.google.com/forum/#!topic/adam-developers/xPtiVHgcFJ8)

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/util/AttributeUtils.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/util/AttributeUtils.scala
@@ -31,10 +31,11 @@ object AttributeUtils {
 
   def convertSAMTagAndValue(attr: SAMTagAndValue): Attribute = {
     attr.value match {
-      case x: java.lang.Integer   => Attribute(attr.tag, TagType.Integer, attr.value.asInstanceOf[Int])
-      case x: java.lang.Character => Attribute(attr.tag, TagType.Character, attr.value.asInstanceOf[Char])
-      case x: java.lang.Float     => Attribute(attr.tag, TagType.Float, attr.value.asInstanceOf[Float])
-      case x: java.lang.String    => Attribute(attr.tag, TagType.String, attr.value.asInstanceOf[String])
+      case x: java.lang.Integer   => Attribute(attr.tag, TagType.Integer, x)
+      case x: java.lang.Character => Attribute(attr.tag, TagType.Character, x)
+      case x: java.lang.Float     => Attribute(attr.tag, TagType.Float, x)
+      case x: java.lang.String    => Attribute(attr.tag, TagType.String, x)
+      case x: Array[Int]          => Attribute(attr.tag, TagType.NumericSequence, x)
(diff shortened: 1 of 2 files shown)


fnothaft:
LGTM, is there more work pending on this, or is it good to go?

ryan-williams:
I think it's gtg

fnothaft:
Thanks @ryan-williams!

PR #802, opened 2015-08-24
stop writing contig names as md5 sums

### LINKED ISSUE #801: Stop writing contig names in the MD5 field
per discussion on #794

ryan-williams:
fixes #801


heuermh:
LGTM

@fnothaft we haven't been very good about an [ADAM-issuenumber] prefix on commit messages lately, is that still something worth doing?

heuermh:
Thanks @ryan-williams

Issue #812, opened 2015-08-26
javadoc error fails build with distribution profile on JDK8


heuermh:
```bash
$ java -version
java version "1.8.0_51"
Java(TM) SE Runtime Environment (build 1.8.0_51-b16)
Java HotSpot(TM) 64-Bit Server VM (build 25.51-b03, mixed mode)

$ mvn -Pdistribution package
...
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO] 
[INFO] ADAM_2.10 .......................................... SUCCESS [  7.118 s]
[INFO] ADAM_2.10: Core .................................... FAILURE [  8.876 s]
[INFO] ADAM_2.10: APIs for Java ........................... SKIPPED
[INFO] ADAM_2.10: CLI ..................................... SKIPPED
[INFO] ADAM_2.10: Distribution ............................ SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 16.235 s
[INFO] Finished at: 2015-08-26T17:20:24-05:00
[INFO] Final Memory: 36M/640M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:2.9.1:jar (attach-javadocs) on project adam-core_2.10: MavenReportException: Error while creating archive:
[ERROR] Exit code: 1 - /Users/xxx/working/adam/adam-core/src/main/java/org/bdgenomics/adam/io/InterleavedFastqInputFormat.java:55: error: unknown tag: date
[ERROR] * @date Feb 2014
[ERROR] ^
[ERROR] /Users/xxx/working/adam/adam-core/src/main/java/org/bdgenomics/adam/io/InterleavedFastqInputFormat.java:206: warning: no @return
[ERROR] public Text createKey() {
[ERROR] ^
[ERROR] /Users/xxx/working/adam/adam-core/src/main/java/org/bdgenomics/adam/io/InterleavedFastqInputFormat.java:213: warning: no @return
[ERROR] public Text createValue() {
[ERROR] ^
[ERROR] /Users/xxx/working/adam/adam-core/src/main/java/org/bdgenomics/adam/io/InterleavedFastqInputFormat.java:220: warning: no @return
[ERROR] public long getPos() {
[ERROR] ^
[...]

Issue #808, opened 2015-08-25
adam appassembler script required but does not run


heuermh:
The appassembler script ```adam``` is required for the other scripts
```bash
$ brew install --verbose --debug adam.rb
==> Finishing up
ln -s ../Cellar/adam/0.17.1/bin/adam-pyspark adam-pyspark
ln -s ../Cellar/adam/0.17.1/bin/adam-shell adam-shell
ln -s ../Cellar/adam/0.17.1/bin/adam-submit adam-submit
ln -s ../Cellar/adam/0.17.1/bin/append_to_option.py append_to_option.py
ln -s ../Cellar/adam/0.17.1/bin/compute-adam-classpath.sh compute-adam-classpath.sh
ln -s ../Cellar/adam/0.17.1/bin/compute-adam-jars.sh compute-adam-jars.sh
==> Summary
/home/linuxbrew/.linuxbrew/Cellar/adam/0.17.1: 512 files, 64M, built in 14 seconds

$ adam-submit
Failed to find appassembler scripts in /bin
You need to build ADAM before running this program
```

I don't like the idea of installing something to the user's path that only throws an exception
```bash
$ brew install --verbose --debug adam.rb
==> Finishing up
ln -s ../Cellar/adam/0.17.1/bin/adam adam
ln -s ../Cellar/adam/0.17.1/bin/adam-pyspark adam-pyspark
ln -s ../Cellar/adam/0.17.1/bin/adam-shell adam-shell
ln -s ../Cellar/adam/0.17.1/bin/adam-submit adam-submit
ln -s ../Cellar/adam/0.17.1/bin/append_to_option.py append_to_option.py
ln -s ../Cellar/adam/0.17.1/bin/compute-adam-classpath.sh compute-adam-classpath.sh
ln -s ../Cellar/adam/0.17.1/bin/compute-adam-jars.sh compute-adam-jars.sh
==> Summary
/home/linuxbrew/.linuxbrew/Cellar/adam/0.17.1: 514 files, 64M, built in 49 seconds

$ adam
Exception in thread "main" java.lang.NoClassDefFoundError: org/apache/spark/Logging
	at java.lang.ClassLoader.defineClass1(Native Method)
	at java.lang.ClassLoader.defineClass(ClassLoader.java:800)
	at java.security.SecureClassLoader.defineClass(SecureClassLoader.java:142)
	at java.net.URLClassLoader.defineClass(URLClassLoader.java:449)
[...]

PR #813, opened 2015-08-26
[ADAM-808] build an assembly cli jar with maven shade plugin

heuermh:
Another approach to #808.

Still in progress (e.g. I'm not sure how ```adam-pyspark``` works), thought I'd should push it early for review/discussion.

Will depend on a fix for #812.


massie:
Thanks for this PR, Michael. 

We originally created an uber jar for distribution but ran into file limitations with zip/jar files (see http://stackoverflow.com/questions/9616250/what-is-the-maximum-number-of-files-per-jar). Now that we've have Spark and Hadoop as `provided` dependencies, the number of files in our jar file should have dropped significantly.

How much head room do we have? How many files are in this uber jar, e.g. what does `jar -tvf adam-uber.jar | wc -l` output?

fnothaft:
I think it'd make sense to move back to the uberjar if we've got enough head room.

adam-cli/pom.xml
```diff
@@ -14,22 +14,35 @@
...
+          <filters>
+            <filter>
+              <artifact>*:*</artifact>
+              <excludes>
+                <exclude>META-INF/*.SF</exclude>
+                <exclude>META-INF/*.DSA</exclude>
+                <exclude>META-INF/*.RSA</exclude>
+              </excludes>
+            </filter>
+          </filters>
```
fnothaft: OOC, I went back to the ol' days before we moved to appassembler and saw a few more exclusions (see at https://github.com/bigdatagenomics/adam/blob/adam-parent-0.6.0/adam-cli/pom.xml#L43). Might be worth doing a diff of the old `maven-shade-plugin` section versus the new one.

bin/adam-shell
```diff
@@ -28,8 +26,34 @@ if [[ -z $@ && -n "$ADAM_OPTS" ]]; then
     echo "Run adam-shell instead as adam-shell <spark-args>"
 fi
 
+# Find ADAM cli assembly jar
+ADAM_CLI_JAR=
+if [ -f "$SCRIPT_DIR/repo" ]; then
```
fnothaft: Why are we checking for a `/repo` directory here? Is this for backwards compatibility with appassembler?

bin/adam-submit
```diff
@@ -52,15 +52,29 @@ fi
 # Figure out where ADAM is installed
 SCRIPT_DIR="$(cd `dirname $0`/..; pwd)"
 
-# Get list of required jars for ADAM
-ADAM_JARS=$("$SCRIPT_DIR"/bin/compute-adam-jars.sh)
+# Find ADAM cli assembly jar
+ADAM_CLI_JAR=
+if [ -f "$SCRIPT_DIR/repo" ]; then
+  ASSEMBLY_DIR="$SCRIPT_DIR/repo"
```
fnothaft: Same question as above here.

bin/adam-submit
```diff
@@ -52,15 +52,29 @@ fi
...
+else
+  ASSEMBLY_DIR="$SCRIPT_DIR/adam-cli/target"
+fi
 
-# Split out the CLI jar, since it will be passed to Spark as the "primary resource".
-ADAM_CLI_JAR=${ADAM_JARS##*,}
-ADAM_JARS=$(echo "$ADAM_JARS" | rev | cut -d',' -f2- | rev)
+num_jars="$(ls -1 "$ASSEMBLY_DIR" | grep "^adam-cli.*\.jar$" | wc -l)"
```
fnothaft: I'd find it preferable to do:

```
jars=({ASSEMBLY_DIR}/adam-cli*.jar)
num_jars=${#jars[*]}
```

Or something of the sort.

adam-cli/pom.xml
heuermh: Yep, and there are quite a few other transformers that e.g. the Spark project uses, will have to look further

bin/adam-shell
heuermh: I updated the distribution assembly.xml to copy the adam-cli jar to repo, so this check is to see if the user is trying to run from the dist.  The Spark build uses a separate assembly module and has a more involved find (in the ```spark-class``` script), we may wish to crib from there.

fnothaft: Ah, OK! Got it.

heuermh:
```bash
$ jar -tvf adam-cli/target/adam-cli_2.10-0.17.2-SNAPSHOT.jar | wc -l
   37210
```
and the limitation was only with JDK6 and earlier, it shouldn't be a problem with JDK7 and later.

fnothaft:
If we are at half of the limit, then I would go with this.

> and the limitation was only with JDK6 and earlier, it shouldn't be a problem with JDK7 and later.

Is ZIP64 the default in JDK7 and onwards?

massie:
Agree. We have plenty of headroom. Zip64 is supported from JDK7 onward. It appears reading online that `zip` is default and, if there are >65535 files, that `zip64` is used. I guess they did that for backward compatibility (since pre-JDK7 can't read zip64).

adam-cli/pom.xml
massie: @heuermh Did you see any changes that needed to by made to the PR when you looked at the old `maven-shade-plugin` xml?

This RP looks good to me but, since it's a big change (and I've hit the merge button too quickly a few times lately), I wanted to check.

fnothaft: I'll build this and run on the cluster and see how it goes.

massie: Thanks, Frank.

heuermh: This is definitely not ready to go yet.

It still needs
 - removal of ```compute-adam-classpath.sh``` and ```compute-adam-jars.sh```; I haven't done this yet because I'm not sure how ```adam-pyspark``` is supposed to work
 - a careful look at maven-shade-plugin excludes/transformer configuration as mentioned above
 - decision on whether to continue using cli module as the assembly artifact or create a separate assembly module as in the Spark build
 - cleanup of changes to ```adam-shell.sh``` and ```adam-submit.sh``` per review notes
 - test assembly artifact on YARN cluster, to confirm issues such as #663 don't resurface
 - test assembly artifact in build for downstream applications

fnothaft: > removal of compute-adam-classpath.sh and compute-adam-jars.sh; I haven't done this yet because I'm not sure how adam-pyspark is supposed to work

I think it is OK to remove both of those, and possibly `adam-pyspark`? CC @laserson 

> decision on whether to continue using cli module as the assembly artifact or create a separate assembly module as in the Spark build

What exactly would the difference be here?

> test assembly artifact on YARN cluster, to confirm issues such as #663 don't resurface

This seems to be OK.

> test assembly artifact in build for downstream applications

I don't think this change should impact downstream applications, except possibly plugins. Are plugins your concern?

fnothaft: `adam-submit` and `adam-shell` are both OK on YARN (the issue seen in #663 does _not_ reproduce).

heuermh: >> test assembly artifact in build for downstream applications

> I don't think this change should impact downstream applications, except possibly plugins. Are plugins your concern?

avocado, rice, qc-metrics, etc. all use classpath-determining scripts similar to those being removed here in their ```bin``` directories.

fnothaft: > avocado, rice, qc-metrics, etc. all use classpath-determining scripts similar to those being removed here in their bin directories.

Oh, sure, but this change doesn't change the adam-{core,apis,cli} artifact that is published to Maven Central.
We can change the packaging in the downstream projects at a later date.

massie:
@heuermh You mentioned that the following needs to be done. How's it going?

- [ ] removal of compute-adam-classpath.sh and compute-adam-jars.sh; I haven't done this yet because I'm not sure how adam-pyspark is supposed to work
- [ ] a careful look at maven-shade-plugin excludes/transformer configuration as mentioned above decision on whether to continue using cli module as the assembly artifact or create a separate assembly module as in the Spark build
- [ ] cleanup of changes to adam-shell.sh and adam-submit.sh per review notes
- [ ] test assembly artifact on YARN cluster, to confirm issues such as #663 don't resurface
- [ ] test assembly artifact in build for downstream applications

@fnothaft Are you using this on the cluster with success?

bin/adam-submit
heuermh: This bit also needs to be more clever, avoiding other artifacts such as adam-cli_...-javadoc.jar
```bash
$ ./bin/adam-shell -h
Found multiple ADAM cli assembly jars in /Users/heuermh/working/adam/adam-cli/target:
adam-cli_2.10-0.17.2-SNAPSHOT-javadoc.jar
adam-cli_2.10-0.17.2-SNAPSHOT.jar
Please remove all but one jar.
```

adam-cli/pom.xml
heuermh: sorry, I've lost the threading for this comment.

I went back to the old shade plugin configuration and don't see those exclusions as being necessary any more.  I don't see ```reference.conf``` either, so I'm not sure what that transformer line was for.

heuermh:
Pushing after rebase with additional changes:  removed ```adam-pyspark``` and ```compute-*``` scripts; removed ```NOTICE.txt``` as cofoja is no longer a dependency; and cleaned up per review above.

I couldn't get this to do the right thing, even with some twiddling
```bash
jars=({ASSEMBLY_DIR}/adam-cli*.jar)
num_jars=${#jars[*]}
```

fnothaft:
> I couldn't get this to do the right thing, even with some twiddling

What (wrong thing) does it do instead?

heuermh:
> What (wrong thing) does it do instead?

If I change it to
```bash
jars=({$ASSEMBLY_DIR}/adam-cli*.jar)
```
then bash doesn't complain but
```bash
num_jars=0
```
when it shouldn't be.  Maybe an OSX thing?

fnothaft:
I will check this out on my end.

fnothaft:
Also, the above RE: `num_jars` looks fine by me as it is.

fnothaft:
@heuermh since this is a big build change, can you rebase this on ToT? Once it passes @AmplabJenkins, I will merge.

heuermh:
I might want to remove ```<mainClass>org.bdgenomics.adam.cli.ADAMMain</mainClass>``` as well.  Per issue #808 running the adam cli jar will fail.  WDYT?

fnothaft:
@heuermh that sounds like a good idea to me. Can you make that change and push it?

heuermh:
ok

fnothaft:
Thanks @heuermh! I've verified this on the cluster, and have merged it as https://github.com/bigdatagenomics/adam/commit/f39de5743e2d995db0eeb22cd324e9dd87596817.

heuermh:
Thanks!

PR #819, opened 2015-09-08
Implement custom ReferenceRegion hashcode

### LINKED ISSUE #817: ReferenceRegion joins fail on clusters
Tried out some simple code for `BroadcastRegionJoin` and saw some odd issues:

```scala
val features = sc.parallelize(Seq(
  Feature
  .newBuilder
  .setStart(14642).setEnd(14882)
  .setContig(Contig.newBuilder.setContigName("1").build)
  .build
)).keyBy(ReferenceRegion(_))

val bigRegions = sc.parallelize(Seq(
  Feature
  .newBuilder
  .setStart(14000).setEnd(14700)
  .setContig(Contig.newBuilder.setContigName("1").build)
  .build
)).keyBy(ReferenceRegion(_))


import org.bdgenomics.adam.rdd.BroadcastRegionJoin

val joined = BroadcastRegionJoin.partitionAndJoin(features, bigRegions)
println(joined.count())
```

Inside the `BroadcastRegionJoin` there are two keyeds RDDs created `smallerKeyed` and `largerKeyed` when testing a simple join those (which do have the correct `ReferenceRegion` of `ReferenceRegion(1,14642,14882,Independent)` - that is also empty.

This seems to have to do with the hashcode of `ReferenceRegion` and containing an enum.
See: http://dev.bizo.com/2014/02/beware-enums-in-spark.html and https://issues.apache.org/jira/browse/SPARK-3847

For example, using the RDDs above, when keyed:

```scala
largerKeyed.collect().foreach(kv => println(kv._1))
largerKeyed.map(_._1.hashCode).collect().foreach(println)
```
 gives

```
ReferenceRegion(1,14642,14882,Independent)
357733957
```

and 

```scala
smallerKeyed.collect().foreach(kv => println(kv._1))
smallerKeyed.map(_._1.hashCode).collect().foreach(println)
```

gives

```
ReferenceRegion(1,14642,14882,Independent)
748344672
```

arahuja:
Fixes #817

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
@@ -19,8 +19,8 @@ package org.bdgenomics.adam.models
 
 import com.esotericsoftware.kryo.io.{ Input, Output }
 import com.esotericsoftware.kryo.{ Kryo, Serializer }
-import org.bdgenomics.adam.rdd.ADAMContext._
 import org.bdgenomics.formats.avro._
+
 import scala.math.{ max, min }
 
 trait ReferenceOrdering[T <: ReferenceRegion] extends Ordering[T] {
@@ -228,6 +228,15 @@ case class ReferenceRegion(referenceName: String, start: Long, end: Long, orient
 
   def length(): Long = {
     end - start
+  }
+
+  override def hashCode: Int = {
+    var result = 37
+    result = 41 * result + (if (referenceName != null) referenceName.hashCode else 0)
+    result = 41 * result + (if (start != null) start.hashCode else 0)
+    result = 41 * result + (if (end != null) end.hashCode else 0)
+    result = 41 * result + (if (orientation != null) orientation.ordinal() else 0)
+    result
   }
 }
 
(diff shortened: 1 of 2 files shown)


arahuja:
Looks the fields can be null sometimes - adding a null check

adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
```diff
@@ -229,6 +229,15 @@ case class ReferenceRegion(referenceName: String, start: Long, end: Long, orient
   def length(): Long = {
     end - start
   }
+
+  override def hashCode: Int = {
+    var result = 37
+    result = 41 * result + (if (referenceName != null) referenceName.hashCode else 0)
+    result = 41 * result + (if (start != null) start.hashCode else 0)
```
laserson: Does this mean that two `ReferenceRegion` objects that differ only in that one has `start=0` and the other has `start=null` would have the same `hashCode`?

laserson: oh, sorry, I read that as `start` rather than `start.hashCode`.

fnothaft:
Thanks @arahuja! I've manually committed this as https://github.com/bigdatagenomics/adam/commit/6997508e1bc47246ce7ec363b2839081807b3df8.

Issue #823, opened 2015-09-17
SAM/BAM sorting should follow @SQ order


laserson:
Following up on discussion in #760 here:

https://github.com/bigdatagenomics/adam/issues/760#issuecomment-130498409

Not clear that we cannot correctly implement the spec.

ryan-williams:
**tl;dr**: possible way to actually fix this issue: add an optional `referenceIndex` to `SequenceRecord` and `Contig` classes, populated only when alignments are read from a SAM file whose `@SQ`s are not lex-sorted; when serializing a `SequenceDictionary`, sort by `referenceIndex` if present, otherwise lex-sort.

Many comments flying around various PRs and things atm but this seems like the best place to centralize discussion.

Here are two salient points that I see in this discussion:

1. The SAM spec seems to explicitly provide for the possibility that `@SQ`s will be deliberately sorted in a non-lexicographical order.
2. We have no way today to keep track of the ancestral SAM-`@SQ`-line ordering once we've `transform`ed into parquet format, so we've rallied around enforcing lexicographical ordering. #794 reached an impasse due to this and we basically gave up.
3. #824 is somewhat of a work-around to this loss of information, but it doesn't seem to me like it really fixes things; in particular, it leaves us unable to `transform` SAM -> ADAM -> SAM and have a non-lex-header-sorting be preserved, afaict (please correct me if I'm wrong).

However, I think there's a way we can actually fix this: add an (optional) `referenceIndex` to `SequenceRecord` and avro `Contig`s that is only populated when reading from a SAM whose `@SQ`s are not lex-sorted (and maybe then only if a flag is passed, if we're worried about space cost).

Then, whenever we build a `SequenceDictionary`, we can order `SequenceRecord`s by their `referenceIndex` fields (if present), and lex-sort them otherwise.

Curious to hear thoughts, lmk if this isn't clear.

fnothaft:
In full agreement with you @ryan-williams. Amusingly enough, back in the [old days](https://github.com/bigdatagenomics/adam/blob/adam-parent-0.6.0/adam-format/src/main/resources/avro/adam.avdl#L14) we did have `referenceIndex` as a field. We dropped it because of the difficulty of merging datasets* but I do agree that it would make sense to add those back as that is the most straightforward way to resolve the sort order problems (IMO, I've been having a similar mental conversation). I wouldn't necessarily make them optional; I might just give the option to renumber the indices if you want the full lexicographical sort.

[...]

Issue #794, opened 2015-08-21
Make `transform` round-trip emit the same BAM file


ryan-williams:
When testing #790, I was thinking that it would be nice if we made `transform`ing a BAM through ADAM and back return the same file, e.g.:

```
adam transform foo.bam foo.adam
adam transform foo.adam foo.2.bam
diff -q foo.bam foo.2.bam
```

Attempting this on toy `.sam` files found in the ADAM repo yields different `.sam` files that are non-trivial (and basically impossible programmatically) to verify are the same as their corresponding originals, which is disorienting!

I noticed four changes we make to BAMs today:

1. we drop `@SQ` lines that aren't mapped to in the file
1. I'm not sure that we preserve the same `@SQ` sort order, if they are not in lex order in the original BAM, nor do we support sorting reads according to an arbitrary `@SQ` order (per discussion on #760).
1. optional tags are not output in the same order.
1. we add an `M5` tag to `@SQ` lines that equals the `SN` field, which seems wrong; why add incorrect MD5 sum info?

I think that is roughly in increasing order of difficulty to fix.

This is by no means essential, but maybe worth working towards a bit, under the theme of making it easy/painless as possible for people to get data in and out of ADAM.

fnothaft:
I think this is reasonable for sorted output.

ryan-williams:
By "sorted output" do you mean "cases where the original BAM already has its `@SQ`s lex-sorted?"

fnothaft:
Ish... More so, I mean BAM files with `@HD order:coordinate`.

ryan-williams:
Interesting; any reason that it doesn't make sense to have identical round-trips with e.g. `@HD SO:queryname`?

btw, this happened:

![](http://cl.ly/image/2U250I1T3Z0K/Screen%20Shot%202015-08-21%20at%202.19.00%20PM.png)

probably just a gmail fluke

fnothaft:
Ah, sorry, I just generally meant !`@HD SO:undefined`. `SO:queryname` should be A-OK as well.

I am indeed hacking your gmail via GitHub... They've caught me! ;)

ryan-williams:
Hm, I just noticed a 4th thing we do differently: we add an `M5` tag to `@SQ` lines that equals the `SN` field, which seems wrong; why add incorrect MD5 sum info?

fnothaft:
[...]

PR #824, opened 2015-09-17
[ADAM-822][ADAM-823] SAM follows @SQ order

laserson:
Not sure this works yet.  Just ran into HadoopGenomics/Hadoop-BAM#21.  Going to sleep now.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/models/SequenceDictionary.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/models/SequenceDictionary.scala
@@ -109,8 +109,7 @@ class SequenceDictionary(val records: Vector[SequenceRecord]) extends Serializab
    * @return Returns a SAM formatted sequence dictionary.
    */
   def toSAMSequenceDictionary: SAMSequenceDictionary = {
-    import SequenceRecord._
-    new SAMSequenceDictionary(records.sorted.map(_ toSAMSequenceRecord).toList)
+    new SAMSequenceDictionary(records.map(_ toSAMSequenceRecord).toList)
   }
 
   override def toString: String = {
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala
@@ -127,10 +127,11 @@ class AlignmentRecordRDDFunctions(rdd: RDD[AlignmentRecord])
   def adamSAMSave(filePath: String,
                   asSam: Boolean = true,
                   asSingleFile: Boolean = false,
-                  isSorted: Boolean = false) = SAMSave.time {
+                  isSorted: Boolean = false,
+                  samHeader: SAMFileHeader = null) = SAMSave.time {
 
     // convert the records
-    val (convertRecords: RDD[SAMRecordWritable], header: SAMFileHeader) = rdd.adamConvertToSAM(isSorted)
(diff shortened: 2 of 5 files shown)


fnothaft:
I am -0 on this change, because it breaks #784 and is only a partial fix for #823. I think this change can be rewritten to still support #784. Right now, we can't completely fix #823, which is why #784 is necessary. Specifically, this change only fixes #823 if you have the original `SamFileHeader`. You'll have this if you started from a SAM file, but you won't if you started from a Parquet file.

adam-core/src/main/scala/org/bdgenomics/adam/models/SequenceDictionary.scala
```diff
@@ -109,8 +109,7 @@ class SequenceDictionary(val records: Vector[SequenceRecord]) extends Serializab
    * @return Returns a SAM formatted sequence dictionary.
    */
   def toSAMSequenceDictionary: SAMSequenceDictionary = {
-    import SequenceRecord._
```
fnothaft: There should be a parameter to choose if you wanted this sorted or not.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala
```diff
@@ -127,10 +127,11 @@ class AlignmentRecordRDDFunctions(rdd: RDD[AlignmentRecord])
   def adamSAMSave(filePath: String,
                   asSam: Boolean = true,
                   asSingleFile: Boolean = false,
-                  isSorted: Boolean = false) = SAMSave.time {
+                  isSorted: Boolean = false,
+                  samHeader: SAMFileHeader = null) = SAMSave.time {
```
fnothaft: I would prefer wrapping this as an option to passing a `null` as a parameter.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala
```diff
@@ -233,16 +234,19 @@ class AlignmentRecordRDDFunctions(rdd: RDD[AlignmentRecord])
    *
    * @return Returns a SAM/BAM formatted RDD of reads, as well as the file header.
    */
-  def adamConvertToSAM(isSorted: Boolean = false): (RDD[SAMRecordWritable], SAMFileHeader) = ConvertToSAM.time {
-    // collect global summary data
-    val sd = rdd.adamGetSequenceDictionary()
-    val rgd = rdd.adamGetReadGroupDictionary()
-
+  def adamConvertToSAM(isSorted: Boolean = false, samHeader: SAMFileHeader = null): (RDD[SAMRecordWritable], SAMFileHeader) = ConvertToSAM.time {
```
fnothaft: Ditto here about `null` as param.

adam-core/src/test/scala/org/bdgenomics/adam/models/SequenceDictionarySuite.scala
```diff
@@ -192,7 +192,7 @@ class SequenceDictionarySuite extends FunSuite {
     toSSD.assertSameDictionary(ssd)
   }
 
-  test("conversion to sam sequence dictionary has correct sort order") {
+  test("conversion to sam sequence dictionary maintains sequence order") {
```
fnothaft: I think you are subtly breaking this test. The sort order correction added by #794 is intended for when you have to regenerate the sequence dictionary via an aggregate, which is what you do when you have a Parquet file. It doesn't make sense to me to break (what I consider) the common case, hence my suggestion to toggle the sort order above.

fnothaft:
Superceded by https://github.com/bigdatagenomics/adam/commit/283ea9de6876e3b0add91ea50ca046d4a52dc7f6.

PR #830, opened 2015-09-24
better validation when saving paired fastqs

ryan-williams:
- `logError` for `LENIENT`
- `throw` for `STRICT

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala
@@ -419,22 +419,38 @@ class AlignmentRecordRDDFunctions(rdd: RDD[AlignmentRecord])
           record.getReadName.toString.dropRight(2)
       })
 
-    if (validationStringency == ValidationStringency.STRICT) {
-      val readIDsWithCounts: RDD[(String, Int)] = readsByID.mapValues(_.size)
-      val unpairedReadIDsWithCounts: RDD[(String, Int)] = readIDsWithCounts.filter(_._2 != 2)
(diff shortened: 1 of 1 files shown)


adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala
```diff
@@ -419,22 +419,35 @@ class AlignmentRecordRDDFunctions(rdd: RDD[AlignmentRecord])
...
+          val readNameOccurrencesMap: collection.Map[Int, Long] = unpairedReadIDsWithCounts.map(_._2).countByValue()
+
+          val msg =
+            List(
+              s"Found $numUnpairedReadIDsWithCounts read names that don't occur exactly twice:",
+
+              readNameOccurrencesMap.map({
+                case (numOccurrences, numReadNames) => s"${numOccurrences}x:\t$numReadNames"
+              }).take(100).mkString("\t", "\n\t", ""),
+              "",
```
fnothaft: Nit: if we have more than 100 improperly paired reads, then I would think we should print an ellipsis at the end of the message.

fnothaft:
Other than the small nit, LGTM!

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala
ryan-williams: ha yea I meant to add that and forgot, thanks, will add

ryan-williams:
nit should be fixed

fnothaft:
Thanks @ryan-williams! I'll leave this open for comments, but if we don't hear from anyone else tomorrow, I'll merge.

fnothaft:
Thanks @ryan-williams!

PR #831, opened 2015-09-24
Support loading paired fastqs

### LINKED ISSUE #771: Support loading paired-FASTQ files
Per [this thread](https://groups.google.com/forum/#!topic/adam-developers/aoGK-uav-Zc), we [support saving paired-FASTQ's](https://github.com/bigdatagenomics/adam/blob/0e3b5570e2e19f4d598f07f6f241f1408b3c6af4/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala#L387) but not reading them.

ryan-williams:
fixes #771

### DIFF
--- a/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
+++ b/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
@@ -86,6 +86,8 @@ class TransformArgs extends Args4jBase with ADAMSaveAnyArgs with ParquetArgs {
   var asSingleFile: Boolean = false
   @Args4jOption(required = false, name = "-paired_fastq", usage = "When converting two (paired) FASTQ files to ADAM, pass the path to the second file here.")
   var pairedFastqFile: String = null
+  @Args4jOption(required = false, name = "-record_group", usage = "Set converted FASTQs' record-group names to this value; if empty-string is passed, use the basename of the input file, minus the extension.")
+  var fastqRecordGroup: String = null
 }
 
 class Transform(protected val args: TransformArgs) extends BDGSparkCommand[TransformArgs] with Logging {
@@ -152,7 +154,7 @@ class Transform(protected val args: TransformArgs) extends BDGSparkCommand[Trans
       if (args.forceLoadBam) {
         sc.loadBam(args.inputPath)
       } else if (args.forceLoadFastq) {
-        sc.loadFastq(args.inputPath, Option(args.pairedFastqFile), stringency)
+        sc.loadFastq(args.inputPath, Option(args.pairedFastqFile), Option(args.fastqRecordGroup), stringency)
       } else if (args.forceLoadIFastq) {
         sc.loadInterleavedFastq(args.inputPath)
       } else if (args.forceLoadParquet) {
@@ -161,6 +163,7 @@ class Transform(protected val args: TransformArgs) extends BDGSparkCommand[Trans
         sc.loadAlignments(
           args.inputPath,
           filePath2Opt = Option(args.pairedFastqFile),
+          recordGroupOpt = Option(args.fastqRecordGroup),
           stringency = stringency
         )
       }
--- a/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
+++ b/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
@@ -88,6 +88,8 @@ class TransformArgs extends Args4jBase with ADAMSaveAnyArgs with ParquetArgs {
   var pairedFastqFile: String = null
   @Args4jOption(required = false, name = "-record_group", usage = "Set converted FASTQs' record-group names to this value; if empty-string is passed, use the basename of the input file, minus the extension.")
   var fastqRecordGroup: String = null
+  @Args4jOption(required = false, name = "-concat", usage = "Concatenate this file with <INPUT> and write the result to <OUTPUT>")
+  var concatFilename: String = null
 }
 
 class Transform(protected val args: TransformArgs) extends BDGSparkCommand[TransformArgs] with Logging {
@@ -150,7 +152,7 @@ class Transform(protected val args: TransformArgs) extends BDGSparkCommand[Trans
   }
 
   def run(sc: SparkContext) {
-    this.apply({
+    val rdd =
       if (args.forceLoadBam) {
         sc.loadBam(args.inputPath)
       } else if (args.forceLoadFastq) {
@@ -167,6 +169,29 @@ class Transform(protected val args: TransformArgs) extends BDGSparkCommand[Trans
           stringency = stringency
         )
       }
+
+    // Optionally load a second RDD and concatenate it with the first.
+    // Paired-FASTQ loading is avoided here because that wouldn't make sense
+    // given that it's already happening above.
+    val concatRddOpt =
+      Option(args.concatFilename).map(concatFilename =>
+        if (args.forceLoadBam) {
+          sc.loadBam(concatFilename)
+        } else if (args.forceLoadIFastq) {
+          sc.loadInterleavedFastq(concatFilename)
+        } else if (args.forceLoadParquet) {
+          sc.loadParquetAlignments(concatFilename)
+        } else {
+          sc.loadAlignments(
+            concatFilename,
+            recordGroupOpt = Option(args.fastqRecordGroup)
+          )
+        }
+      )
+
+    this.apply(concatRddOpt match {
+      case Some(concatRdd) => rdd ++ concatRdd
+      case None            => rdd
     }).adamSave(args, args.sortReads)
   }
 
--- a/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
+++ b/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
@@ -50,8 +50,8 @@ class TransformArgs extends Args4jBase with ADAMSaveAnyArgs with ParquetArgs {
   var markDuplicates: Boolean = false
   @Args4jOption(required = false, name = "-recalibrate_base_qualities", usage = "Recalibrate the base quality scores (ILLUMINA only)")
   var recalibrateBaseQualities: Boolean = false
-  @Args4jOption(required = false, name = "-strict_bqsr", usage = "Run BQSR with strict validation.")
-  var strictBQSR: Boolean = false
+  @Args4jOption(required = false, name = "-stringency", usage = "Stringency level for various checks; can be SILENT, LENIENT, or STRICT. Defaults to LENIENT")
+  var stringency: String = "LENIENT"
   @Args4jOption(required = false, name = "-dump_observations", usage = "Local path to dump BQSR observations to. Outputs CSV format.")
   var observationsPath: String = null
   @Args4jOption(required = false, name = "-known_snps", usage = "Sites-only VCF giving location of known SNPs")
@@ -91,6 +91,8 @@ class TransformArgs extends Args4jBase with ADAMSaveAnyArgs with ParquetArgs {
 class Transform(protected val args: TransformArgs) extends BDGSparkCommand[TransformArgs] with Logging {
   val companion = Transform
 
+  val stringency = ValidationStringency.valueOf(args.stringency)
+
   def apply(rdd: RDD[AlignmentRecord]): RDD[AlignmentRecord] = {
 
     var adamRecords = rdd
@@ -124,14 +126,11 @@ class Transform(protected val args: TransformArgs) extends BDGSparkCommand[Trans
     if (args.recalibrateBaseQualities) {
       log.info("Recalibrating base qualities")
       val knownSnps: SnpTable = createKnownSnpsTable(sc)
-      val stringency = if (args.strictBQSR) {
-        ValidationStringency.STRICT
-      } else {
-        ValidationStringency.LENIENT
-      }
-      adamRecords = adamRecords.adamBQSR(sc.broadcast(knownSnps),
+      adamRecords = adamRecords.adamBQSR(
+        sc.broadcast(knownSnps),
         Option(args.observationsPath),
-        stringency)
+        stringency
+      )
     }
 
     if (args.coalesce != -1) {
@@ -153,7 +152,7 @@ class Transform(protected val args: TransformArgs) extends BDGSparkCommand[Trans
       if (args.forceLoadBam) {
         sc.loadBam(args.inputPath)
       } else if (args.forceLoadFastq) {
-        sc.loadFastq(args.inputPath, Option(args.pairedFastqFile))
+        sc.loadFastq(args.inputPath, Option(args.pairedFastqFile), stringency)
       } else if (args.forceLoadIFastq) {
         sc.loadInterleavedFastq(args.inputPath)
       } else if (args.forceLoadParquet) {
--- a/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
+++ b/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
@@ -158,7 +158,11 @@ class Transform(protected val args: TransformArgs) extends BDGSparkCommand[Trans
       } else if (args.forceLoadParquet) {
         sc.loadParquetAlignments(args.inputPath)
       } else {
-        sc.loadAlignments(args.inputPath)
+        sc.loadAlignments(
+          args.inputPath,
+          filePath2Opt = Option(args.pairedFastqFile),
+          stringency = stringency
+        )
       }
     }).adamSave(args, args.sortReads)
   }
--- a/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
+++ b/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
@@ -84,6 +84,8 @@ class TransformArgs extends Args4jBase with ADAMSaveAnyArgs with ParquetArgs {
   var forceLoadParquet: Boolean = false
   @Args4jOption(required = false, name = "-single", usage = "Saves OUTPUT as single file")
   var asSingleFile: Boolean = false
+  @Args4jOption(required = false, name = "-paired_fastq", usage = "When converting two (paired) FASTQ files to ADAM, pass the path to the second file here.")
+  var pairedFastqFile: String = null
 }
 
 class Transform(protected val args: TransformArgs) extends BDGSparkCommand[TransformArgs] with Logging {
@@ -151,7 +153,7 @@ class Transform(protected val args: TransformArgs) extends BDGSparkCommand[Trans
       if (args.forceLoadBam) {
         sc.loadBam(args.inputPath)
       } else if (args.forceLoadFastq) {
-        sc.loadUnpairedFastq(args.inputPath)
+        sc.loadFastq(args.inputPath, Option(args.pairedFastqFile))
       } else if (args.forceLoadIFastq) {
         sc.loadInterleavedFastq(args.inputPath)
       } else if (args.forceLoadParquet) {
--- a/adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
@@ -114,6 +114,7 @@ class FastqRecordConverter extends Serializable with Logging {
   }
 
   def convertRead(element: (Void, Text),
+                  recordGroupOpt: Option[String] = None,
                   setFirstOfPair: Boolean = false,
                   setSecondOfPair: Boolean = false): AlignmentRecord = {
     val lines = element._2.toString.split('\n')
@@ -147,7 +148,7 @@ class FastqRecordConverter extends Serializable with Logging {
     require(readSequence.length == readQualities.length,
       "Read " + readName + " has different sequence and qual length.")
 
-    AlignmentRecord.newBuilder()
+    val builder = AlignmentRecord.newBuilder()
       .setReadName(readName)
       .setSequence(readSequence)
       .setQual(readQualities)
@@ -163,6 +164,9 @@ class FastqRecordConverter extends Serializable with Logging {
       .setPrimaryAlignment(null)
       .setSecondaryAlignment(null)
       .setSupplementaryAlignment(null)
-      .build()
+
+    recordGroupOpt.foreach(builder.setRecordGroupName)
+
+    builder.build()
   }
 }
--- a/adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
@@ -119,8 +119,28 @@ class FastqRecordConverter extends Serializable with Logging {
     val lines = element._2.toString.split('\n')
     require(lines.length == 4, "Record has wrong format:\n" + element._2.toString)
 
+    def trimTrailingReadNumber(readName: String): String = {
+      if (readName.endsWith("/1")) {
+        if (setSecondOfPair) {
+          throw new Exception(
+            s"Found read name $readName ending in '/1' despite second-of-pair flag being set"
+          )
+        }
+        readName.dropRight(2)
+      } else if (readName.endsWith("/2")) {
+        if (setFirstOfPair) {
+          throw new Exception(
+            s"Found read name $readName ending in '/2' despite first-of-pair flag being set"
+          )
+        }
+        readName.dropRight(2)
+      } else {
+        readName
+      }
+    }
+
     // get fields for first read in pair
-    val readName = lines(0).drop(1)
+    val readName = trimTrailingReadNumber(lines(0).drop(1))
     val readSequence = lines(1)
     val readQualities = lines(3)
 
--- a/adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
@@ -113,7 +113,9 @@ class FastqRecordConverter extends Serializable with Logging {
       .build()
   }
 
-  def convertRead(element: (Void, Text)): AlignmentRecord = {
+  def convertRead(element: (Void, Text),
+                  setFirstOfPair: Boolean = false,
+                  setSecondOfPair: Boolean = false): AlignmentRecord = {
     val lines = element._2.toString.split('\n')
     require(lines.length == 4, "Record has wrong format:\n" + element._2.toString)
 
@@ -129,9 +131,13 @@ class FastqRecordConverter extends Serializable with Logging {
       .setReadName(readName)
       .setSequence(readSequence)
       .setQual(readQualities)
-      .setReadPaired(false)
+      .setReadPaired(setFirstOfPair || setSecondOfPair)
       .setProperPair(null)
-      .setReadNum(0)
+      .setReadNum(
+        if (setFirstOfPair) 0
+        else if (setSecondOfPair) 1
+        else null
+      )
       .setReadNegativeStrand(null)
       .setMateNegativeStrand(null)
       .setPrimaryAlignment(null)
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
@@ -232,9 +232,7 @@ class ADAMContext(val sc: SparkContext) extends Serializable with Logging {
   }
 
   def loadBam(filePath: String): RDD[AlignmentRecord] = {
-
     val path = new Path(filePath)
-
     val fs =
       Option(
         FileSystem.get(path.toUri, sc.hadoopConfiguration)
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
@@ -19,15 +19,20 @@ package org.bdgenomics.adam.rdd
 
 import java.io.FileNotFoundException
 import java.util.regex.Pattern
+import htsjdk.samtools.{ IndexedBamInputFormat, SAMFileHeader, ValidationStringency }
 import htsjdk.samtools.{ ValidationStringency, SAMFileHeader, IndexedBamInputFormat }
 import org.apache.avro.Schema
 import org.apache.avro.generic.IndexedRecord
 import org.apache.avro.specific.SpecificRecord
-import org.apache.hadoop.fs.{ FileSystem, FileStatus, Path }
+import org.apache.hadoop.fs.{ FileStatus, FileSystem, Path }
 import org.apache.hadoop.io.{ LongWritable, Text }
 import org.apache.hadoop.mapreduce.lib.input.TextInputFormat
-import org.apache.spark.rdd.RDD
+import org.apache.parquet.avro.{ AvroParquetInputFormat, AvroReadSupport }
+import org.apache.parquet.filter2.predicate.FilterPredicate
+import org.apache.parquet.hadoop.ParquetInputFormat
+import org.apache.parquet.hadoop.util.ContextUtil
 import org.apache.spark.rdd.MetricsContext._
+import org.apache.spark.rdd.RDD
 import org.apache.spark.{ Logging, SparkContext }
 import org.bdgenomics.adam.converters._
 import org.bdgenomics.adam.instrumentation.Timers._
@@ -43,12 +48,8 @@ import org.bdgenomics.adam.rich.RichAlignmentRecord
 import org.bdgenomics.formats.avro._
 import org.bdgenomics.utils.instrumentation.Metrics
 import org.bdgenomics.utils.misc.HadoopUtil
-import org.seqdoop.hadoop_bam.util.SAMHeaderReader
 import org.seqdoop.hadoop_bam._
-import org.apache.parquet.avro.{ AvroParquetInputFormat, AvroReadSupport }
-import org.apache.parquet.filter2.predicate.FilterPredicate
-import org.apache.parquet.hadoop.ParquetInputFormat
-import org.apache.parquet.hadoop.util.ContextUtil
+import org.seqdoop.hadoop_bam.util.SAMHeaderReader
 import scala.collection.JavaConversions._
 import scala.collection.Map
 
@@ -103,7 +104,7 @@ object ADAMContext {
   implicit def setToJavaSet[A](set: Set[A]): java.util.Set[A] = setAsJavaSet(set)
 }
 
-import ADAMContext._
+import org.bdgenomics.adam.rdd.ADAMContext._
 
 class ADAMContext(val sc: SparkContext) extends Serializable with Logging {
 
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
@@ -639,7 +639,9 @@ class ADAMContext(val sc: SparkContext) extends Serializable with Logging {
 
   def loadAlignments(
     filePath: String,
-    projection: Option[Schema] = None): RDD[AlignmentRecord] = LoadAlignmentRecords.time {
+    projection: Option[Schema] = None,
+    filePath2Opt: Option[String] = None,
+    stringency: ValidationStringency = ValidationStringency.STRICT): RDD[AlignmentRecord] = LoadAlignmentRecords.time {
 
     if (filePath.endsWith(".sam") ||
       filePath.endsWith(".bam")) {
@@ -651,7 +653,7 @@ class ADAMContext(val sc: SparkContext) extends Serializable with Logging {
     } else if (filePath.endsWith(".fq") ||
       filePath.endsWith(".fastq")) {
       log.info("Loading " + filePath + " as unpaired FASTQ and converting to AlignmentRecords. Projection is ignored.")
-      loadUnpairedFastq(filePath)
+      loadFastq(filePath, filePath2Opt, stringency)
     } else if (filePath.endsWith(".fa") ||
       filePath.endsWith(".fasta")) {
       log.info("Loading " + filePath + " as FASTA and converting to AlignmentRecords. Projection is ignored.")
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
@@ -364,15 +364,17 @@ class ADAMContext(val sc: SparkContext) extends Serializable with Logging {
 
   def loadFastq(filePath1: String,
                 filePath2Opt: Option[String],
+                recordGroupOpt: Option[String] = None,
                 stringency: ValidationStringency = ValidationStringency.STRICT): RDD[AlignmentRecord] = {
     filePath2Opt match {
-      case Some(filePath2) => loadPairedFastq(filePath1, filePath2, stringency)
+      case Some(filePath2) => loadPairedFastq(filePath1, filePath2, recordGroupOpt, stringency)
       case None            => loadUnpairedFastq(filePath1)
     }
   }
 
   def loadPairedFastq(filePath1: String,
                       filePath2: String,
+                      recordGroupOpt: Option[String],
                       stringency: ValidationStringency): RDD[AlignmentRecord] = {
     val reads1 = loadUnpairedFastq(filePath1, setFirstOfPair = true)
     val reads2 = loadUnpairedFastq(filePath2, setSecondOfPair = true)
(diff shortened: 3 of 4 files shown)


adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
```diff
@@ -84,6 +84,8 @@ class TransformArgs extends Args4jBase with ADAMSaveAnyArgs with ParquetArgs {
   var forceLoadParquet: Boolean = false
   @Args4jOption(required = false, name = "-single", usage = "Saves OUTPUT as single file")
   var asSingleFile: Boolean = false
+  @Args4jOption(required = false, name = "-filename_record_group", usage = "Set converted FASTQs' record-group names to this value; if empty-string is passed, use the basename of the input file, minus the extension.")
```
fnothaft: I feel like this would be a bit cleaner as two switches:

* One that is "set record group ID from filename"
* One that is "set record group ID to passed in string"

They would be mutually exclusive. A "set sample ID" flag would be useful as well, IMO. I defer to you as to whether you'd like to implement that or not.

adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
```diff
@@ -86,19 +89,23 @@ class FastqRecordConverter extends Serializable with Logging {
...
-      .setProperPair(null)
-      .setFirstOfPair(null)
-      .setSecondOfPair(null)
-      .setReadNegativeStrand(null)
-      .setMateNegativeStrand(null)
-      .setPrimaryAlignment(null)
-      .setSecondaryAlignment(null)
-      .setSupplementaryAlignment(null)
-      .build()
+    val builder =
```
fnothaft: Nit: Would it be OK with you to put `val builder =` and `AlignmentRecord.newBuilder()` on the same line? Splitting them across two lines doesn't improve readability (IMO) and causes a large spurious diff (because the spacing of the following lines is changed).

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -19,15 +19,20 @@ package org.bdgenomics.adam.rdd
 
 import java.io.FileNotFoundException
 import java.util.regex.Pattern
-import htsjdk.samtools.SAMFileHeader
+
```
fnothaft: Nit: spacing.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -42,15 +47,11 @@ import org.bdgenomics.adam.rich.RichAlignmentRecord
...
 import org.bdgenomics.utils.instrumentation.Metrics
 import org.bdgenomics.utils.misc.HadoopUtil
-import org.seqdoop.hadoop_bam.util.SAMHeaderReader
 import org.seqdoop.hadoop_bam._
-import org.apache.parquet.avro.{ AvroParquetInputFormat, AvroReadSupport }
-import org.apache.parquet.filter2.predicate.FilterPredicate
-import org.apache.parquet.hadoop.ParquetInputFormat
-import org.apache.parquet.hadoop.util.ContextUtil
+import org.seqdoop.hadoop_bam.util.SAMHeaderReader
+
```
fnothaft: Nit: spacing

Thank you for sorting the imports, though!

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -578,9 +590,35 @@ class ADAMContext(val sc: SparkContext) extends Serializable with Logging {
     }
   }
 
+  def loadPairedFastQ(filePath1: String,
```
fnothaft: Nit: for consistency in capitalization, let's do `Fastq`, not `FastQ`.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -578,9 +590,35 @@ class ADAMContext(val sc: SparkContext) extends Serializable with Logging {
...
+                      filePath2: String,
+                      stringency: ValidationStringency,
+                      recordGroupOpt: Option[String]): RDD[(AlignmentRecord, AlignmentRecord)] = {
+    val reads1 = loadUnpairedFastq(filePath1, recordGroupOpt, setFirstOfPair = true)
+    val reads2 = loadUnpairedFastq(filePath2, recordGroupOpt, setSecondOfPair = true)
+
+    stringency match {
+      case ValidationStringency.STRICT | ValidationStringency.LENIENT =>
+        val count1 = reads1.count
```
fnothaft: Cache before count?

fnothaft: Or, is the presumption that loading should be pretty fast?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -578,9 +590,35 @@ class ADAMContext(val sc: SparkContext) extends Serializable with Logging {
...
+        val count2 = reads2.count
+
+        if (count1 != count2) {
+          val msg = s"Fastq 1 ($filePath1) has $count1 reads, fastq 2 ($filePath2) has $count2 reads"
+          if (stringency == ValidationStringency.STRICT)
+            throw new IllegalArgumentException(msg)
+          else if (stringency == ValidationStringency.LENIENT)
+            logError(msg)
```
fnothaft: If you hit this case, don't you need to `join` on readname instead of `zip`ing?

fnothaft: I suppose you don't _need to_, but doing the join would guarantee correct output. If I was calling this function with Lenient validation stringency and mismatched reads, that is the semantics I would expect. What are your thoughts? Does my view of the semantics seem reasonable or unreasonable?

fnothaft:
Overall LGTM! I've dropped a few comments inline RE: style nits and semantics.

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
ryan-williams: Good suggestions. I'm rethinking having this functionality in the first place, and it's not really related to what this PR's title says it's about, so I'm thinking of just splitting it out indefinitely.

adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
ryan-williams: sure, will do

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
ryan-williams: hm yea, intellij "optimized" these imports when I accidentally hit ⌘⇧O at some point, and it does blank lines between java stdlib, other, and scala stdlib imports, I guess. In any case I'll remove it.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
ryan-williams: accidental IJ ftw!

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
ryan-williams: eh I think `cache`ing makes sense, good call

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
ryan-williams: cool, logging an error and doing the join / dropping single reads sgtm.

ryan-williams:
Per my first comment response, I think I want to split out the filename->readGroup stuff. It's a little dirty and I don't know that I actually need it.

Also, I think having a way for `transform` to actually convert two fastq files to one ADAM file is a sensible thing and missing from this PR, so I'm going to add that, and respond to your other comments, thanks for the review, lmk if you have opinions on any of that plan.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
fnothaft: Ha, nice. "Optimized".

fnothaft:
That plan sounds great here; I would agree that 2 FASTQ --> 1 ADAM is definitely useful.

ryan-williams:
OK, much simpler PR now, and the functionality is exposed via `transform`

ryan-williams:
jenkins, retest this please

fnothaft:
Jenkins, test this please.

ryan-williams:
I've added some more functionality here, pretty well delineated in to individual commits if you want to browse them that way. This all addresses issues I encountered `transform`ing the illumina platinum fastqs to ADAM format:

* set recordGroupName via cmdline arg `-record_group`
* merge a set of alignments with another via cmdline arg `-concat <filename>`
* control validation stringency via cmdline arg `-stringency` (replaces `-strict_BQSR`)

With all of this, I can run a command like the following, which seems to me like the minimum functionality necessary to get the data we need out of paired `.fastq`s and in to output `.adam`s:

```
bin/adam-submit -- transform foo_1.fastq -paired_fastq foo_2.fastq -concat foo_unpaired.fastq -record_group foo foo.adam
```

Split out on to more lines, for maybe more clarity:
```
bin/adam-submit --           \
  transform                  \
  foo_1.fastq                \
  -paired_fastq foo_2.fastq  \
  -concat foo_unpaired.fastq \
  -record_group foo          \
  foo.adam
```

This takes a `_1.fastq` file, a `_2.fastq`, and a corresponding `_unpaired.fastq` file (present in [the 50x-coverage platinum data](http://www.ebi.ac.uk/ena/data/view/ERP001960)) and makes one `.adam` file / `RDD[AlignmentRecord]` out of them, setting the recordGroup name in the process.

I've run variations of this over the ~14TB of platinum data several times, and passing `-stringency SILENT` as well is a huge time-saver for skipping [two `.cache.count`s](https://github.com/bigdatagenomics/adam/pull/831/files#diff-d36ea7d0742decd0b040a73a96af06e9R387).

I'm interested in peoples' thoughts on this; there are several bits of work herein but they logically comprise supporting "loading paired-fastq data" in the minimum sane way, from my experimenting.

fnothaft:
OOC, why is the last commit (make flagstat safe...) rolled into this PR? All the others look OK, except possibly for 2df67c3 but I'll comment on that one inline.

Reviewing now.

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
```diff
@@ -84,11 +84,19 @@ class TransformArgs extends Args4jBase with ADAMSaveAnyArgs with ParquetArgs {
   var forceLoadParquet: Boolean = false
   @Args4jOption(required = false, name = "-single", usage = "Saves OUTPUT as single file")
   var asSingleFile: Boolean = false
+  @Args4jOption(required = false, name = "-paired_fastq", usage = "When converting two (paired) FASTQ files to ADAM, pass the path to the second file here.")
+  var pairedFastqFile: String = null
+  @Args4jOption(required = false, name = "-record_group", usage = "Set converted FASTQs' record-group names to this value; if empty-string is passed, use the basename of the input file, minus the extension.")
+  var fastqRecordGroup: String = null
+  @Args4jOption(required = false, name = "-concat", usage = "Concatenate this file with <INPUT> and write the result to <OUTPUT>")
+  var concatFilename: String = null
```
fnothaft: As I was commenting on the main page of the PR, I don't get the usage case for this? If you want to consolidate multiple files, why wouldn't you just use this command with the glob syntax to select multiple input files?

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
```diff
@@ -121,14 +130,11 @@ class Transform(protected val args: TransformArgs) extends BDGSparkCommand[Trans
     if (args.recalibrateBaseQualities) {
       log.info("Recalibrating base qualities")
       val knownSnps: SnpTable = createKnownSnpsTable(sc)
-      val stringency = if (args.strictBQSR) {
```
fnothaft: I'm not sure how I feel about collapsing import stringency and BQSR stringency down into a single flag. The two aren't logically tied together.

Is your thought here that the two stringencies (aside: what is the plural of stringency? I think that I am doing English wrong here...) are independent? I.e., you will only set the stringency flag on import if you are reading from FASTQ, which is unaligned data. If you're running BQSR, your data must be aligned, therefore the function of the stringency flag is context dependent?

Let me know if what I said wasn't clear...

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
fnothaft: Ping on this nit.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -580,7 +632,10 @@ class ADAMContext(val sc: SparkContext) extends Serializable with Logging {
 
   def loadAlignments(
     filePath: String,
-    projection: Option[Schema] = None): RDD[AlignmentRecord] = LoadAlignmentRecords.time {
+    projection: Option[Schema] = None,
+    filePath2Opt: Option[String] = None,
```
fnothaft: As a minor aside, we should throw an IAE if someone uses `filePath2Opt` and we detect a file that isn't a FASTQ.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/FlagStat.scala
```diff
@@ -93,15 +94,15 @@ object FlagStat {
...
-          b2i(p.getReadMapped),
-          b2i(p.getReadPaired),
-          b2i(p.getReadPaired && p.getFirstOfPair),
-          b2i(p.getReadPaired && p.getSecondOfPair),
-          b2i(p.getReadPaired && p.getProperPair),
-          b2i(p.getReadPaired && p.getReadMapped && p.getMateMapped),
-          b2i(p.getReadPaired && p.getReadMapped && !p.getMateMapped),
-          b2i(mateMappedToDiffChromosome),
-          b2i(mateMappedToDiffChromosome && p.getMapq >= 5),
+          b2i(b(p.getReadMapped)),
```
fnothaft: These changes look reasonable but I think we should pull these changes out to another PR.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/FlagStat.scala
```diff
@@ -93,15 +94,15 @@ object FlagStat {
...
+          b2i(b(p.getReadPaired)),
+          b2i(b(p.getReadPaired) && b(p.getFirstOfPair)),
+          b2i(b(p.getReadPaired) && b(p.getSecondOfPair)),
+          b2i(b(p.getReadPaired) && b(p.getProperPair)),
+          b2i(b(p.getReadPaired) && b(p.getReadMapped) && b(p.getMateMapped)),
+          b2i(b(p.getReadPaired) && b(p.getReadMapped) && b(!p.getMateMapped)),
+          b2i(b(mateMappedToDiffChromosome)),
+          b2i(b(mateMappedToDiffChromosome && p.getMapq >= 5)),
```
fnothaft: As an aside, should we parameterize the `mapq` here?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/FlagStat.scala
ryan-williams: Yea, the genesis of this was that I was running `flagstat` on my paired-fastq-converted `.adam`s and it was NPEing on several of these fields which were [left unset or set to `null` by `FastqRecordConverter.convert`](https://github.com/bigdatagenomics/adam/pull/831/files#diff-eb57d01d374e75a94641ec53a4b62fa5R117).

So given that my paired-fastq-`.adam`s were unable pass a basic read-counting verification step post-conversion, I thought this functionality should be included in this PR.

lmk if you still think it should be in a separate PR, I can go either way.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/FlagStat.scala
ryan-williams: you mean to guard against a NPE if `p.getMapq` is null? Sure, that makes sense, will do

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
ryan-williams: What kind of detecting do you think we should do? Inference based on filename extension? I'm not sure I like the sound of that but will defer to you

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/FlagStat.scala
fnothaft: Ah, I see. Thanks for the additional context. Flagstat is meant to be run on aligned data, not unaligned data, which explains some of the oddness. I don't think it is a bad idea to sanitize the fields though, esp. if the performance impact is small. Any chance you could do a quick head to head on an aligned BAM? Given the reasoning, I'd be OK with including the commit here, but let's not squash it down on merge.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/FlagStat.scala
fnothaft: Oh no, I meant the 5 in `p.getMapq >= 5`. But, sanitizing the p.getMapq is a good idea since not all aligners emit mapping qualities; I've gotten that as a bug report downstream in avocado.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
fnothaft: I'm thinking of a really small change; I'll shoot you a PR, because it feels somewhat cumbersome to type out not in code.

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
ryan-williams: Your question makes sense. I guess the short answer is that yes, I think the two notions of stringency are "independent" and unlikely to both need to be specified (let alone to different values) in the same `transform` run.

I'd go a step further though and say that it's worth formalizing the idea that, in the context of any one `transform` command, there might be a level of strictness of sanity/validation-checking that we should expose to the user via a cmdline flag. On the other hand, though, exposing multiple flags related to this basic idea feels redundant / like over-doing it, to me.

fnothaft: That sounds reasonable, but I got kinda lost here:

> I'd go a step further though and say that it's worth formalizing the idea that, in the context of any one `transform` command, there might be a level of strictness of sanity/validation-checking that we should expose to the user via a cmdline flag.

Just to check my understanding, are you saying that it'd be reasonable for each subpart of `transform` (e.g., BQSR, indel realignment, load from FASTQ, load from BAM, mark dups, etc) to have a validation strictness flag, but that we should only expose a single flag? If so, I think that makes sense in this case, and is probably not a bad default policy in general.

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
ryan-williams: well, in the paired-fastq case (cf. [the example cmdline on the main thread](https://github.com/bigdatagenomics/adam/pull/831#issuecomment-144555308)) I don't think using a glob as the `<INPUT>` argument helps me: in the presence of the `-paired_fastq` flag, the `<INPUT>` is interpreted as a FASTQ with the first-of-pair reads; the `-concat` flag is used to add some unpaired reads in to the same `.adam` as the paired reads.

The glob syntax would result in trying to `zip` up one `RDD` that is [first-of-pair and unpaired reads] with a second that is [second-of-pair] reads.

In any case, allowing for an explicit way to merge two files together in `transform`, which the `-concat` flag does here, seems worth doing, e.g. for cases where a glob is too coarse a mechanism.

I do have some desire to make an `ADAMCommand` that will merge two *or more* `RDD[AlignmentRecord]` files, since in the 200x Illumina Platinum data there are many FASTQs from each individual and I want to merge them into per-individual FASTQs, but I'd open a separate issue/PR for that.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
ryan-williams: sorry, just did it, will update the PR now

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
ryan-williams: same here, will update now

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/FlagStat.scala
ryan-williams: OK, will h2h on a BAM and let you know how it goes

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/FlagStat.scala
ryan-williams: ah I see.

I'll sanitize the `p.getMapq` and leave parameterizing the `5` as work for another PR

fnothaft: SGTM!

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
fnothaft: Just shot https://github.com/ryan-williams/adam/pull/1 your way. LMK your thoughts!

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
fnothaft: I think the paired-fastq case makes a lot of sense, but to me, that is a special case (hence the inclusion of the separate `-paired_fastq` flag and code path).

> The glob syntax would result in trying to `zip` up one `RDD` that is [first-of-pair and unpaired reads] with a second that is [second-of-pair] reads.

Glob should give the union, not the zip. For me, I would see the `merge` command that you described as duplicative of globbing. In the case you're describing, running `adam-submit merge RG1.adam RG2.adam RG3.adam ... merged.adam; adam-submit flagstat merged.adam` and `adam-submit flagstat RG*.adam` (RG1,2,3,...,n.adam are the converted FASTQs, one per read group) should yield equivalent results. Am I oversimplifying?

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
ryan-williams: Yea, across many things that `transform` does, it's reasonable to let the user say "be {strict,lenient,silent} with any applicable validation checks", and that that should happen via one flag that is not explicitly namespaced to any one task that the user may already separately be telling `transform` to do (e.g. BQSR).

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
ryan-williams: >> The glob syntax would result in trying to zip up one RDD … with a second

> Glob should give the union, not the zip.

Yea, but in the paired-fastq case I need something more like a `zip` (in [the validation code-path](https://github.com/bigdatagenomics/adam/pull/831/files#diff-d36ea7d0742decd0b040a73a96af06e9R387); really I just need to verify the two files' `count`s), not a `union`; my point was that, given that globbing does a `union`, it is not suitable for my use-case here.

> For me, I would see the merge command that you described as duplicative of globbing.

Only if the files being merged are all the files in a directory, right? What if I want to merge two specific files from a directory that has 10 similarly-named files in it?

The main point I'm trying to make is: we're exposing a logical "merge" capability but hiding it in the "glob" code path (which also restricts the kinds of filenames that can be merged), which feels indirect/confusing/obfuscated to me; I think it's good to support globbing but we should also expose the general capability that it represents (mergeing RDD[AR]s) more explicitly.

One semi-related thought is that maybe implying "concatenation" is not ideal, since what is really happening is a "union"? I could make this flag `-merge` instead of `-concat` if you like.

In fact, one possible approach I'd considered for exposing a more general "merge" capability was to have a `-merge` flag that takes a comma-delimited list of filenames to merge together (since args4j and the larger Scala ecosystem in 2015 have no solution for parsing a variable number of arguments, afaik).

Thanks, let me know your thoughts on all of this.

fnothaft: > Yea, but in the paired-fastq case I need something more like a zip (in the validation code-path; really I just need to verify the two files' counts), not a union; my point was that, given that globbing does a union, it is not suitable for my use-case here.

Oh, yeah, agreed there.

> Only if the files being merged are all the files in a directory, right? What if I want to merge two specific files from a directory that has 10 similarly-named files in it?

I don't think that restriction exists... Let me test this out.

> The main point I'm trying to make is: we're exposing a logical "merge" capability but hiding it in the "glob" code path (which also restricts the kinds of filenames that can be merged), which feels indirect/confusing/obfuscated to me; I think it's good to support globbing but we should also expose the general capability that it represents (mergeing RDD[AR]s) more explicitly.

My feeling here is that the "Hadoopey" way to "merge" files is with the glob syntax.

> (since args4j and the larger Scala ecosystem in 2015 have no solution for parsing a variable number of arguments, afaik).

<sarcasm> LOL, why would anyone ever want to parse a variable number of args, right? </sarcasm>

ryan-williams: > I don't think that restriction exists... Let me test this out.

Scenario: directory conatins files:
```
foo1.adam
foo2.adam
foo3.adam
foo4.adam
```

I want to merge `foo1.adam` and `foo3.adam`. How do I do that with a glob?

fnothaft: Here's how. So, I've got the [NCI-60 cell line VCFs](http://discover.nci.nih.gov/cellminer/home.do) on my cluster:

```
```

fnothaft: You can also do this across directories. Let's say I also wanted to load the chr20 VCF from 1000G, which is located here:

```
[fnothaft@amp-bdg-master data]$ hdfs dfs -ls /data/1kg-genotypes/
Found 24 items
-rw-r--r--   3 fnothaft bdg  34999123163 2015-09-11 20:23 /data/1kg-genotypes/0181860441fce741b1c34c27de0c7d95.ftp___ftp-trace.
ncbi.nih.gov_1000genomes_ftp_release_20110521_ALL.chr15.phase1_release_v3.20101123.snps_indels_svs.genotypes.vcf
-rw-r--r--   3 fnothaft bdg  85579386265 2015-09-11 21:38 /data/1kg-genotypes/0bdeb99215dfda2a7d2fe3c7fffbb8b1.ftp___ftp-trace.
ncbi.nih.gov_1000genomes_ftp_release_20110521_ALL.chr3.phase1_release_v3.20101123.snps_indels_svs.genotypes.vcf
-rw-r--r--   3 fnothaft bdg  75066370280 2015-09-11 21:25 /data/1kg-genotypes/2b67e1d968d9d90acfbd07a9a56d8985.ftp___ftp-trace.
ncbi.nih.gov_1000genomes_ftp_release_20110521_ALL.chr6.phase1_release_v3.20101123.snps_indels_svs.genotypes.vcf
-rw-r--r--   3 fnothaft bdg  56592766853 2015-09-11 21:05 /data/1kg-genotypes/386a681a3f413e426fcfd94a8f28d1ff.ftp___ftp-trace.
ncbi.nih.gov_1000genomes_ftp_release_20110521_ALL.chr12.phase1_release_v3.20101123.snps_indels_svs.genotypes.vcf
-rw-r--r--   3 fnothaft bdg  78364336068 2015-09-11 21:30 /data/1kg-genotypes/3cba98c57413edffa54335382723afbd.ftp___ftp-trace.
ncbi.nih.gov_1000genomes_ftp_release_20110521_ALL.chr5.phase1_release_v3.20101123.snps_indels_svs.genotypes.vcf
-rw-r--r--   3 fnothaft bdg  67654231596 2015-09-11 21:21 /data/1kg-genotypes/3ce9ec00721599910a8559f7a9b4b6b0.ftp___ftp-trace.
ncbi.nih.gov_1000genomes_ftp_release_20110521_ALL.chr8.phase1_release_v3.20101123.snps_indels_svs.genotypes.vcf
-rw-r--r--   3 fnothaft bdg  42522617578 2015-09-11 20:49 /data/1kg-genotypes/4332e865880a0499d32378cbdbad291a.ftp___ftp-trace.
ncbi.nih.gov_1000genomes_ftp_release_20110521_ALL.chr13.phase1_release_v3.20101123.snps_indels_svs.genotypes.vcf
-rw-r--r--   3 fnothaft bdg  33723323691 2015-09-11 20:22 /data/1kg-genotypes/5119cbb69a595dbc5fc33f729c534b8d.ftp___ftp-trace.
ncbi.nih.gov_1000genomes_ftp_release_20110521_ALL.chr18.phase1_release_v3.20101123.snps_indels_svs.genotypes.vcf
-rw-r--r--   3 fnothaft bdg  16064789455 2015-09-11 21:03 /data/1kg-genotypes/5c7ecbcb8396de30f00c389bfeeb2180.ftp___ftp-trace.
ncbi.nih.gov_1000genomes_ftp_release_20110521_ALL.chr21.phase1_release_v3.20101123.snps_indels_svs.genotypes.vcf
-rw-r--r--   3 fnothaft bdg  25236539814 2015-09-11 20:16 /data/1kg-genotypes/5d2fbed137fabddb7db1978088f2cb07.ftp___ftp-trace.
ncbi.nih.gov_1000genomes_ftp_release_20110521_ALL.chr19.phase1_release_v3.20101123.snps_indels_svs.genotypes.vcf
-rw-r--r--   3 fnothaft bdg  37491381512 2015-09-11 20:31 /data/1kg-genotypes/6dcc8433a2e0af7411a88408bd783686.ftp___ftp-trace.
ncbi.nih.gov_1000genomes_ftp_release_20110521_ALL.chr16.phase1_release_v3.20101123.snps_indels_svs.genotypes.vcf
-rw-r--r--   3 fnothaft bdg  15304146564 2015-09-11 21:12 /data/1kg-genotypes/727230f6a9d4dcd581d00c67dce2ac41.ftp___ftp-trace.
ncbi.nih.gov_1000genomes_ftp_release_20110521_ALL.chr22.phase1_release_v3.20101123.snps_indels_svs.genotypes.vcf
-rw-r--r--   3 fnothaft bdg  38957455789 2015-09-11 20:33 /data/1kg-genotypes/7cb1c0c0b7f0e744b9515f2793d3473b.ftp___ftp-trace.
ncbi.nih.gov_1000genomes_ftp_release_20110521_ALL.chr14.phase1_release_v3.20101123.snps_indels_svs.genotypes.vcf
-rw-r--r--   3 fnothaft bdg  93086828627 2015-09-11 21:38 /data/1kg-genotypes/83904d42bf1aeb708ee3dcb92457cd34.ftp___ftp-trace.
ncbi.nih.gov_1000genomes_ftp_release_20110521_ALL.chr1.phase1_release_v3.20101123.snps_indels_svs.genotypes.vcf
-rw-r--r--   3 fnothaft bdg 102433657522 2015-09-11 21:41 /data/1kg-genotypes/89a49cd0c6da043ee67783a2e656089e.ftp___ftp-trace.
ncbi.nih.gov_1000genomes_ftp_release_20110521_ALL.chr2.phase1_release_v3.20101123.snps_indels_svs.genotypes.vcf
-rw-r--r--   3 fnothaft bdg  39585660568 2015-09-11 20:35 /data/1kg-genotypes/8ec5d5d5e2bafbcb34834a7350abaa7c.ftp___ftp-trace.
ncbi.nih.gov_1000genomes_ftp_release_20110521_ALL.chrX.phase1_release_v3.20101123.snps_indels_svs.genotypes.vcf
-rw-r--r--   3 fnothaft bdg  58670776704 2015-09-11 21:11 /data/1kg-genotypes/9f4fc764f6252797b0cee1ebc2d2fa0c.ftp___ftp-trace.ncbi.nih.gov_1000genomes_ftp_release_20110521_ALL.chr11.phase1_release_v3.20101123.snps_indels_svs.genotypes.vcf
hr11.phase1_release_v3.20101123.snps_indels_svs.genotypes.vcf
-rw-r--r--   3 fnothaft bdg            0 2015-09-11 21:11 /data/1kg-genotypes/_SUCCESS
-rw-r--r--   3 fnothaft bdg  68602158250 2015-09-11 21:23 /data/1kg-genotypes/c076f1c67ed5105c787b27ca82cacc19.ftp___ftp-trace.ncbi.nih.gov_1000genomes_ftp_release_20110521_ALL.chr7.phase1_release_v3.20101123.snps_indels_svs.genotypes.vcf
-rw-r--r--   3 fnothaft bdg  26476036862 2015-09-11 20:01 /data/1kg-genotypes/d47964c28806a97a1f656c7802237e5c.ftp___ftp-trace.ncbi.nih.gov_1000genomes_ftp_release_20110521_ALL.chr20.phase1_release_v3.20101123.snps_indels_svs.genotypes.vcf
-rw-r--r--   3 fnothaft bdg  51166771761 2015-09-11 20:57 /data/1kg-genotypes/deecfc73a58a1270467835b94c01adaf.ftp___ftp-trace.ncbi.nih.gov_1000genomes_ftp_release_20110521_ALL.chr9.phase1_release_v3.20101123.snps_indels_svs.genotypes.vcf
-rw-r--r--   3 fnothaft bdg  58300480951 2015-09-11 21:10 /data/1kg-genotypes/e88431af8b2e8623836d85664fadf97c.ftp___ftp-trace.ncbi.nih.gov_1000genomes_ftp_release_20110521_ALL.chr10.phase1_release_v3.20101123.snps_indels_svs.genotypes.vcf
-rw-r--r--   3 fnothaft bdg  84767980059 2015-09-11 21:34 /data/1kg-genotypes/f282e6bd9e5f733f12224cc0a572cb39.ftp___ftp-trace.ncbi.nih.gov_1000genomes_ftp_release_20110521_ALL.chr4.phase1_release_v3.20101123.snps_indels_svs.genotypes.vcf
-rw-r--r--   3 fnothaft bdg  32386267135 2015-09-11 20:23 /data/1kg-genotypes/f2b4e316e8228a4827442a43857b8dca.ftp___ftp-trace.ncbi.nih.gov_1000genomes_ftp_release_20110521_ALL.
```

Then I run:

```
./bin/adam-submit ${ADAM_OPTS} -- vcf2adam "{/data/1kg-genotypes,nci-60}/*{OVCAR-3,OVCAR-8,chr20}*.vcf" nci-60/OVCAR-38_and_1kg_chr20.adam
```

I am not proud of that last glob, but it gets the job done.

ryan-williams: Ah, right, we just pass the `<INPUT>` arg to Hadoop and let it expand things, so we don't just get `*` expansion but `{}` and a few other things, thanks, I knew that at one point but forgot.

In any case, I don't think making that the only way to access merging-rdd functionality is ideal; relying on hadoop's globbing, making sure your shell doesn't expand it for you first, remembering subtle differences between hadoop's glob expansion and most shells', etc. make it a little tricky. Plus, there's the use here where I actually need to distinguish between the unpaired file that I am merging and the file in the `<INPUT>` argument.

So tl;dr: good that we can do this with "globs" but I still feel like exposing it explicitly makes sense.

fnothaft: SGTM.

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
fnothaft: Yeah, that sounds reasonable.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/FlagStat.scala
ryan-williams: Maybe trivial but I validated it on `bqsr1.sam`, which has a few types of reads represented:

```
1000 + 0 in total (QC-passed reads + QC-failed reads)
0 + 0 primary duplicates
0 + 0 primary duplicates - both read and mate mapped
0 + 0 primary duplicates - only read mapped
0 + 0 primary duplicates - cross chromosome
0 + 0 secondary duplicates
0 + 0 secondary duplicates - both read and mate mapped
0 + 0 secondary duplicates - only read mapped
0 + 0 secondary duplicates - cross chromosome
992 + 0 mapped (99.20%:0.00%)
1000 + 0 paired in sequencing
498 + 0 read1
502 + 0 read2
980 + 0 properly paired (98.00%:0.00%)
984 + 0 with itself and mate mapped
8 + 0 singletons (0.80%:0.00%)
4 + 0 with mate mapped to a different chr
0 + 0 with mate mapped to a different chr (mapQ>=5)
```

lmk if you want me to find a trickier example to test on

fnothaft: SGTM. I think this was the last issue pending on this PR? Are we good to merge from your end?

ryan-williams: yea I think this is good to go from my end, thanks for the reviews!

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
fnothaft: We will resolve this in #841.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/FlagStat.scala
fnothaft: np! thanks for the good changes. I will squash and merge.

fnothaft:
Jenkins, test this please.

fnothaft:
Jenkins, test this please.

fnothaft:
Jenkins, test this please.

fnothaft:
Thanks @ryan-williams ! I've merged this as https://github.com/bigdatagenomics/adam/commit/4c9654ee1fe3b9e8fff92059d3686f5fd4a1f6de and https://github.com/bigdatagenomics/adam/commit/9db4108b324154fae0404f0dd5f46ab86bbad7ec.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -365,8 +363,47 @@ class ADAMContext(val sc: SparkContext) extends Serializable with Logging {
...
+            throw new IllegalArgumentException(msg)
+          else {
+            // ValidationStringency.LENIENT
+            logError(msg)
+          }
+        }
+      case ValidationStringency.SILENT =>
+    }
+
+    reads1 ++ reads2
```
allenday: If I understand correctly, this puts the _1 reads first, followed by the _2 reads.

If so, is this really what's wanted?  Doesn't it make more sense to interleave?

fnothaft: > If so, is this really what's wanted? Doesn't it make more sense to interleave?
>
> How about reads1.zip(reads2)

That's a fair point, but I generally try to write any Spark transformations without assuming the relative order of records (or more precisely, I only make assumptions about the relative order of records after sorting).

> Ideally they will be preserved as pairs in a single record so that if we need to process the pair as a pair or round trip the .adam back out to .fastq (what I'm doing now) it's trivial to do. As is now, again if I understand correctly ++ operator, read pairs are broken up.

We've added the [`Fragment`](https://github.com/bigdatagenomics/bdg-formats/blob/master/src/main/resources/avro/bdg.avdl#L286) schema recently to cover this use case. However, I don't think we've connected the FASTQ loading code to load FASTQ data as `Fragment`s.

shibuvp:
im trying to convert fastq files into adam .bt it is not working         
my command------------------
./adam-submit -- transform hdfs://master.hdp:8020/Sample/NewFile1.fastq -paired_fastq hdfs://master.hdp:8020/Sample/NewFile2.fastq -record_group hdfs://master.hdp:8020/Sample/adamresult001.adam

PR #832, opened 2015-09-24
add ability to recursively transform files

### LINKED ISSUE #741: Allow transform to recursively search for BAMs in a directory and convert them all to ADAM

ryan-williams:
fixes #741

### DIFF
--- a/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
+++ b/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
@@ -18,6 +18,7 @@
 package org.bdgenomics.adam.cli
 
 import htsjdk.samtools.ValidationStringency
+import org.apache.hadoop.fs.{ FileStatus, Path, FileSystem }
 import org.apache.spark.rdd.RDD
 import org.apache.spark.{ Logging, SparkContext }
 import org.bdgenomics.adam.algorithms.consensus._
@@ -84,6 +85,10 @@ class TransformArgs extends Args4jBase with ADAMSaveAnyArgs with ParquetArgs {
   var forceLoadParquet: Boolean = false
   @Args4jOption(required = false, name = "-single", usage = "Saves OUTPUT as single file")
   var asSingleFile: Boolean = false
+  @Args4jOption(required = false, name = "-recursive", usage = "Recursively search <INPUT> directory for files with extension matching the argument to -input_extension, and convert them to extension <OUTPUT>")
+  var recursive: Boolean = false
+  @Args4jOption(required = false, name = "-input_extension", usage = "When -recursive is enabled, look in the <INPUT> directory for files with this extension. Can include leading '.' or not.")
+  var inputExtension: String = null
 }
 
 class Transform(protected val args: TransformArgs) extends BDGSparkCommand[TransformArgs] with Logging {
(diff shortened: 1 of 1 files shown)


ryan-williams:
jenkins, retest this please

ryan-williams:
jenkins, retest this please

ryan-williams:
Looks like I'm using a hadoop 2 API and breaking the hadoop 1 build.

I'll look for a workaround. Is it time to drop hadoop 1 yet?

laserson:
I thought we decided to drop Hadoop 1 support a while ago.

fnothaft:
We dropped Hadoop 1 temporarily because Spark 1.4.x broke Hadoop 1 support. This is fixed in Spark 1.5.0. We've been holding on to Hadoop 1 because `spark-ec2`'s Hadoop 2 config is irredeemably broken, and because `spark-ec2` is _arguably_ the "easiest way for someone to launch a Spark cluster for testing on AWS". I would be glad to drop Hadoop 1 support if we can recommend a better alternative to `spark-ec2`.

What is it that uses a Hadoop 2 API in a breaking way? The other reason we've kept Hadoop 1 support is that most of the time, the Hadoop 2 vs. 1 API issues have reasonably straightforward WARs.

ryan-williams:
> What is it that uses a Hadoop 2 API in a breaking way?

[this bit](https://github.com/bigdatagenomics/adam/pull/832/files#diff-1ee985bcfcfad885cfc143944371af07R161) is the best SO had to offer for recursively traversing a directory in Hadoop, but it's Hadoop2-only

fnothaft:
> this bit is the best SO had to offer for recursively traversing a directory in Hadoop, but it's Hadoop2-only

I'm not 100% sure, but I think you can get similar semantics using `fs.globStatus`. We're using `fs.globStatus` successfully [elsewhere in ADAM](https://github.com/bigdatagenomics/adam/blob/9bf6942bc5c34245292a780896c7fbe7ef408f2e/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala#L302), so it should be good across both Hadoop 1 and 2.

ryan-williams:
Ah yea I'd noticed `globStatus` around; from what I can tell from [the docs](https://github.com/apache/hadoop/blob/release-1.0.4/src/core/org/apache/hadoop/fs/FileSystem.java#L923), it only supports a `*` wildcard for within one path segment, not a `**` for recursive traversal.

I'll look into manually generating a list of absolute paths in a hadoop1-compatible way.

ryan-williams:
OK just pushed a version that hopefully works with Hadoop 1 and 2

ryan-williams:
jenkins, retest this please

fnothaft:
Jenkins, test this please.

ryan-williams:
does jenkins not abide by commands via email?

ryan-williams:
I added some docstrings and a CYA check so this should be good to go when tests pass unless anyone wants me to add tests here.

ryan-williams:
wat, that wasn't even a force-push. Any leads on why jenkins does this @fnothaft ?

jenkins, retest this please

ryan-williams:
seems like jenkins isn't listening to me when i ask for retests at all, anyway I force-pushed a new change here with a removal of an unneeded import

fnothaft:
>> Any leads on why jenkins does this

There's a bug in the Jenkins git plugin, IIRC. @shaneknapp ?

>> seems like jenkins isn't listening to me when i ask for retests at all

I think it looks for a single comment, and I think the capitalization needs to match exactly what Jenkins is looking for (see [commands here](https://wiki.jenkins-ci.org/display/JENKINS/GitHub+pull+request+builder+plugin)).

ryan-williams:
retest this please

ryan-williams:
hm, any protips figuring out why these tests failed above?

fnothaft:
The ADAM build has been flapping for a while now...

fnothaft:
Jenkins, test this please.

fnothaft:
Jenkins, retest this please.

ryan-williams:
Closing per [my comment on #833](https://github.com/bigdatagenomics/adam/issues/833#issuecomment-143789541)

PR #834, opened 2015-09-25
add "file" cmd for running multiple cmds at once

### LINKED ISSUE #833: Add "file" command for reading multiple ADAM command-lines from a file and running them in sequence
I had this idea yesterday while thinking about ways to extend #832, and I think it will be a simple and powerful abstraction to expose in the ADAM CLI.

My immediate impetus is that I have a directory with many pairs of `.fastq` files that I want converted to ADAM.

There's not an obvious way to tell e.g. `transform`: "find all pairs of files of that end with `_1.fastq` vs. `_2.fastq`, and send them in pairs through the `transform` cmd a la #831", nor should there be, probably. Instead, I can generate the commands I want via some shell commands, and then pass them as a file to ADAM.

ryan-williams:
fixes #833 

Example use:

```sh
$ cat cmds
transform /Users/ryan/c/adam/adam-core/src/test/resources/reads12.sam /Users/ryan/c/adam/adam-core/src/test/resources/reads12.adam
transform /Users/ryan/c/adam/adam-core/src/test/resources/bqsr1.sam /Users/ryan/c/adam/adam-core/src/test/resources/bqsr1.adam

# "find-suffix" bash helper of mine, equivalent to: find . -name "*$1"
$ fns .adam  

$ bin/adam-submit file cmds

# The two .adam files exist where there were none previously
$ fns .adam
adam-core/src/test/resources/bqsr1.adam
adam-core/src/test/resources/reads12.adam
```

### DIFF
--- a/adam-cli/src/main/scala/org/bdgenomics/adam/cli/ADAMMain.scala
+++ b/adam-cli/src/main/scala/org/bdgenomics/adam/cli/ADAMMain.scala
@@ -66,8 +66,24 @@ object ADAMMain extends Logging {
           BuildInformation,
           View
         )
+      ),
+      CommandGroup(
+        "META",
+        List(
+          ExecCommandsFromFile
+        )
       )
     )
+
+  val commands =
+    for {
+      grp <- commandGroups
+      cmd <- grp.commands
+    } yield cmd
+
+  def getCommand(commandName: String): Option[BDGCommandCompanion] = {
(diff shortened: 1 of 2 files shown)


ryan-williams:
Closing per [my comment on #833](https://github.com/bigdatagenomics/adam/issues/833#issuecomment-143789541)

PR #839, opened 2015-09-28
let flagstat write to file

### LINKED ISSUE #838: Let `flagstat` write output to a file
Currently I get `flagstat` output mixed in with my driver's stdout; I'm interested in running it over many files and programmatically verifying that the results match some expected values. I think that optionally writing the output to a file is a good way to accomplish this.

ryan-williams:
fixes #838


fnothaft:
Thanks @ryan-williams! Merged as https://github.com/bigdatagenomics/adam/commit/c2b0844a0bb8d2c2ee8d8ca78d297fbf205b3af8.

PR #840, opened 2015-09-28
fix flagstat output file encoding

ryan-williams:
Lesson: don't pick [random encodings from `DataOutputStream`](http://docs.oracle.com/javase/7/docs/api/java/io/DataOutputStream.html?is-external=true#writeUTF(java.lang.String)) to output your files in!

Also, cleans up some extra whitespace.

### DIFF
--- a/adam-cli/src/main/scala/org/bdgenomics/adam/cli/FlagStat.scala
+++ b/adam-cli/src/main/scala/org/bdgenomics/adam/cli/FlagStat.scala
@@ -89,7 +89,7 @@ class FlagStat(protected val args: FlagStatArgs) extends BDGSparkCommand[FlagSta
                |%d + %d singletons (%.2f%%:%.2f%%)
                |%d + %d with mate mapped to a different chr
                |%d + %d with mate mapped to a different chr (mapQ>=5)
-             """.stripMargin('|').format(
+             """.stripMargin('|').trim.format(
       passedVendorQuality.total, failedVendorQuality.total,
       passedVendorQuality.duplicatesPrimary.total, failedVendorQuality.duplicatesPrimary.total,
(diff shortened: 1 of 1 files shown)


adam-cli/src/main/scala/org/bdgenomics/adam/cli/FlagStat.scala
```diff
@@ -121,7 +121,8 @@ class FlagStat(protected val args: FlagStatArgs) extends BDGSparkCommand[FlagSta
         val fs = FileSystem.get(conf)
         val path = new Path(outputPath)
         val outputStream = fs.create(path, true)
-        outputStream.writeUTF(output)
+        outputStream.writeBytes(output)
```
fnothaft: Why `writeBytes` instead of `writeChars`?

ryan-williams: if all output chars are < 128 (which should always be true of this output), then `writeBytes` should actually be UTF8, whereas `writeChars` will output two bytes per char, and the former seems preferable to me? I was thinking that the existing `println` behavior is just doing one byte per char but I don't actually know, this is not my area of expertise. lmk what you think.

fnothaft: Oh, I see. I am OK either way.

ryan-williams: fwiw I just tested with `writeChars` and my shell displays the results just like it did before, which was my only litmus test, so I'm fine either way

fnothaft:
Thanks @ryan-williams!

PR #846, opened 2015-10-05
Updating maven build plugin dependency versions

heuermh:


fnothaft:
LGTM.

pom.xml
```diff
@@ -23,7 +23,7 @@
     <parquet.version>1.8.1</parquet.version>
     <!-- Edit the following line to configure the Hadoop (HDFS) version. -->
     <hadoop.version>2.2.0</hadoop.version>
-    <scoverage.version>0.99.2</scoverage.version>
+    <scoverage.version>1.1.1</scoverage.version>
```
fnothaft: As an aside, have you checked the output of the `coverage` profile with this? I don't think the `coverage` profile works now, would be interesting to see if this fixed it. If not, we should open an issue to tackle later.

heuermh: Will check it out shortly

heuermh: No it does not, let me try with ```0.99.10``` . . .

heuermh: I see as of ```0.99.2``` support for scala 2.10 was dropped.
https://github.com/scoverage/scalac-scoverage-plugin/blob/master/2.10.md

Updating to ```1.1.1``` as this pull request does results in a maven transitive dependency error.

Versions ```0.99.2``` and ```0.99.10``` fail with
```adam-core/target/scoverage.coverage.xml (No such file or directory)```

Downgrading to ```0.98.2``` fails with
```adam-core/target/scala-2.10.4/classes/scoverage.coverage (No such file or directory)```

fnothaft: I've opened #849 for this.

ryan-williams:
this lgtm, thanks @heuermh. I love the feeling of all of these tools improving under us :)

fnothaft:
Thanks @heuermh! Rebased and merged as https://github.com/bigdatagenomics/adam/commit/bbb0c32266250057a95e2cdab647d38832729609.

PR #852, opened 2015-10-06
[ADAM-851] Slienced Parquet logging.

### LINKED ISSUE #851: Parquet logger has reawaken from its slumber
After years (a year and a half? two years?) of quiet, the logging inside of Parquet has reawaken and is spamming all of our logs.

This is because when we [upgraded to Parquet 1.7.0](https://github.com/bigdatagenomics/adam/commit/a70b4ca12299a9639275671cb20f7887c63ecb09), the root package of Parquet changed from `parquet` to `org.apache.parquet` and we didn't update [this line](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/util/ParquetLogger.scala#L25) in the `ParquetLogger` utility.

fnothaft:
Fixed the ParquetLogger utility so that it silences `org.apache.parquet.hadoop`
instead of `parquet.hadoop`. All is quiet again. Resolves #851.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/util/ParquetLogger.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/util/ParquetLogger.scala
@@ -22,7 +22,7 @@ import java.util.logging.{ Level, Logger }
 object ParquetLogger {
 
   val hadoopLoggerLevel = (level: Level) => {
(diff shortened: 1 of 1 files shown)


fnothaft:
Rebased.

heuermh:
Thanks!

PR #858, opened 2015-10-08
Add Avro LRU cache for strings to reduce memory use

massie:
This commit adds a new Avro Binary Decoder that uses an LRU cache
of Utf8 (string) objects to minimize the creation of Uft8 objects.
If a Utf8 exists in the cache, its reference is used instead of
creating a new object. Moreover, the decoder readString() method
reuses a single Utf8 object when reading strings instead of
creating a new Utf8 object on each read.

The read() method in AvroSerializer was synchronized which isn't
necessary since Spark ensures that only a single thread will
call read. That sychronization is removed with this commit.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/serialization/BinaryDecoderWithLRUCache.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/serialization/BinaryDecoderWithLRUCache.scala
@@ -1,0 +1,120 @@
+/**
+ * Licensed to Big Data Genomics (BDG) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The BDG licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.bdgenomics.adam.serialization
+
+import java.io.{ IOException, InputStream }
+import java.nio.ByteBuffer
+import java.util
+import java.util.Map.Entry
+
+import org.apache.avro.io.{ BinaryDecoder, Decoder, DecoderFactory }
+import org.apache.avro.util.Utf8
+
+/**
+ * Note: A LinkedHashMap takes 40 * SIZE + 4 * CAPACITY bytes. If the size == capacity, then the memory
+ * footprint is CAPACITY * 44 bytes.
+ */
+private[serialization] class LRUCache[K, V](maxCapacity: Int, loadFactor: Double = 0.75f)
+    extends util.LinkedHashMap[K, V](maxCapacity, 0.75f, true) {
+  assert(maxCapacity > 0, "LRU cache maxCapacity should be greater than 0")
+  assert(0 < loadFactor && loadFactor <= 1.0, "LRU load factor must be (0,1.0]")
+  override def removeEldestEntry(eldest: Entry[K, V]): Boolean = {
+    size() > maxCapacity
+  }
+}
+
+/**
+ * An Avro binary decoder that keep a LRU cache of strings to minimize memory pressure.
+ *
+ * WARNING: This class is not thread-safe! This isn't an issue since Spark guarantees that only
+ * a single thread will access it.
+ *
+ * A amount of memory strings take in Java is 8 * (int) ((((#chars) * 2) + 45) / 8) bytes, e.g. a string that is
+ * 128 characters long will take, 8 * ((((128) * 2) + 45) / 8) = 296 bytes. This BinaryDecoder is not for general
+ * use but is optimized for ADAM -omics data, where strings are typically ~128 characters long.
+ *
+ * The default cache capacity size of 64K elements, equates to a maximum memory use of (44*64K) for the cache and
+ * (296*64K) for the strings, a total of ~22 MB when full.
+ *
+ * @param in The input stream of Avro data
+ */
+private[serialization] class BinaryDecoderWithLRUCache(in: InputStream, cache: LRUCache[Utf8, Utf8]) extends Decoder {
+
+  def this(in: InputStream, cacheCapacity: Int = 65535) = this(in, new LRUCache[Utf8, Utf8](cacheCapacity))
+
+  val wrappedDecoder = DecoderFactory.get().directBinaryDecoder(in, null.asInstanceOf[BinaryDecoder])
+
+  private val scratchUtf8: Utf8 = new Utf8
+
+  @throws(classOf[IOException])
+  override def readString: String = {
+    readString(scratchUtf8).toString
+  }
+
+  @throws(classOf[IOException])
+  override def readString(old: Utf8): Utf8 = {
+    // NOTE: The 'old' parameter is intentionally being ignored. We always reuse 'scratchUtf8' to keep
+    // from creating a new Utf8 object every time we read a string. This isn't thread-safe, of course,
+    // but Spark ensures that each thread has its own serialization infrastructure.
+    val readValue = wrappedDecoder.readString(scratchUtf8)
+    if (!cache.containsKey(readValue)) {
+      val valueCopy = new Utf8(readValue)
+      cache.put(valueCopy, valueCopy)
+    }
+    cache.get(readValue)
+  }
+
+  override def readLong(): Long = wrappedDecoder.readLong()
+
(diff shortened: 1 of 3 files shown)


adam-core/src/main/scala/org/bdgenomics/adam/serialization/BinaryDecoderWithLRUCache.scala
```diff
@@ -0,0 +1,120 @@
...
+ * use but is optimized for ADAM -omics data, where strings are typically ~128 characters long.
+ *
+ * The default cache capacity size of 64K elements, equates to a maximum memory use of (44*64K) for the cache and
+ * (296*64K) for the strings, a total of ~22 MB when full.
+ *
+ * @param in The input stream of Avro data
+ */
+private[serialization] class BinaryDecoderWithLRUCache(in: InputStream, cache: LRUCache[Utf8, Utf8]) extends Decoder {
+
+  def this(in: InputStream, cacheCapacity: Int = 65535) = this(in, new LRUCache[Utf8, Utf8](cacheCapacity))
```
heuermh: Sorry, likely scala ignorance here, where is the ctr (InputStream, LRUCache) defined?

massie: Two line above.

ryan-williams: you have a shadowed `in` here, is that weird?

adam-core/src/main/scala/org/bdgenomics/adam/serialization/BinaryDecoderWithLRUCache.scala
```diff
@@ -0,0 +1,120 @@
...
+    readString(scratchUtf8).toString
+  }
+
+  @throws(classOf[IOException])
+  override def readString(old: Utf8): Utf8 = {
+    // NOTE: The 'old' parameter is intentionally being ignored. We always reuse 'scratchUtf8' to keep
+    // from creating a new Utf8 object every time we read a string. This isn't thread-safe, of course,
+    // but Spark ensures that each thread has its own serialization infrastructure.
+    val readValue = wrappedDecoder.readString(scratchUtf8)
+    if (!cache.containsKey(readValue)) {
```
heuermh: I might prefer using Guava's ```Cache<K,V>``` instead of LinkedHashMap.  The properties & behavior of the cache implementation is more configurable (see [CacheBuilder](http://docs.guava-libraries.googlecode.com/git-history/release/javadoc/com/google/common/cache/CacheBuilder.html)) and I think the [get method](http://docs.guava-libraries.googlecode.com/git-history/release/javadoc/com/google/common/cache/Cache.html#get(K,%20java.util.concurrent.Callable)) semantics are more clear than the "if cached, return; otherwise create, cache and return" pattern.

adam-core/src/main/scala/org/bdgenomics/adam/serialization/BinaryDecoderWithLRUCache.scala
massie: I have two ctors here so that I can pass in the cache for testing. Do you have recommendations on a better way to do this?

adam-core/src/main/scala/org/bdgenomics/adam/serialization/BinaryDecoderWithLRUCache.scala
massie: I'll take a look at Guava now. Thanks for the suggestion, @heuermh .

adam-core/src/main/scala/org/bdgenomics/adam/serialization/BinaryDecoderWithLRUCache.scala
fnothaft: I think @ryan-williams was saying that the `in` in `def this(in: InputStream,` shadows the `in` in `private[serialization] class BinaryDecoderWithLRUCache(in: InputStream`.

adam-core/src/main/scala/org/bdgenomics/adam/serialization/BinaryDecoderWithLRUCache.scala
heuermh: I believe for a LRU cache implementation all you need is
```java
Cache<Utf8, Utf8> cache = CacheBuilder.newBuilder().maximumSize(65535L);
```
and then the get would be
```java
cache.get(readValue, new Callable() {
  @Override
  public Utf8 call() throws Exception {
    return new Utf8(readValue);
  }
});
```
and something to handle or rethrow the checked Exception, after java &rarr; scala of course.  :)

adam-core/src/main/scala/org/bdgenomics/adam/serialization/BinaryDecoderWithLRUCache.scala
```diff
@@ -0,0 +1,120 @@
...
+ * A amount of memory strings take in Java is 8 * (int) ((((#chars) * 2) + 45) / 8) bytes, e.g. a string that is
+ * 128 characters long will take, 8 * ((((128) * 2) + 45) / 8) = 296 bytes. This BinaryDecoder is not for general
+ * use but is optimized for ADAM -omics data, where strings are typically ~128 characters long.
+ *
+ * The default cache capacity size of 64K elements, equates to a maximum memory use of (44*64K) for the cache and
+ * (296*64K) for the strings, a total of ~22 MB when full.
+ *
+ * @param in The input stream of Avro data
+ */
+private[serialization] class BinaryDecoderWithLRUCache(in: InputStream, cache: LRUCache[Utf8, Utf8]) extends Decoder {
```
fnothaft: OOC, why are we caching strings as `Utf8`? We moved the schemas from `Utf8` to `String` in https://github.com/bigdatagenomics/bdg-formats/commit/2e22217c97e2dbc2062e11bc9e7ab829528bf776.

adam-core/src/main/scala/org/bdgenomics/adam/serialization/BinaryDecoderWithLRUCache.scala
```diff
@@ -0,0 +1,120 @@
...
+
+/**
+ * An Avro binary decoder that keep a LRU cache of strings to minimize memory pressure.
+ *
+ * WARNING: This class is not thread-safe! This isn't an issue since Spark guarantees that only
+ * a single thread will access it.
+ *
+ * A amount of memory strings take in Java is 8 * (int) ((((#chars) * 2) + 45) / 8) bytes, e.g. a string that is
+ * 128 characters long will take, 8 * ((((128) * 2) + 45) / 8) = 296 bytes. This BinaryDecoder is not for general
+ * use but is optimized for ADAM -omics data, where strings are typically ~128 characters long.
```
fnothaft: I don't think you want the cache to look at the read data (sequence and quality). This data is long (increases cache size) and unique (low hit rate). I think we would ideally have a tuning parameter that sets the max string length that we look into the cache for.

adam-core/src/main/scala/org/bdgenomics/adam/serialization/BinaryDecoderWithLRUCache.scala
massie: Utf8 is an Avro internal object that allow Strings to be mutable for performance. I believe, setting the string type to "String" tells Avro how to pass String data back to the user.. it doesn't alter the internal representation of Strings.

fnothaft: Ah, OK, gotcha. Thanks for the clarification!

adam-core/src/main/scala/org/bdgenomics/adam/serialization/BinaryDecoderWithLRUCache.scala
massie: So you're recommending a string length cutoff of say 64 bytes so that referenceName, readName, cigar, etc strings are cached but sequence and quality are likely not? I like that idea.

fnothaft: Yeah, I think that would be reasonable. I might go down to 50. Would it be easy to make this a property that we can check against at runtime?

adam-core/src/main/scala/org/bdgenomics/adam/serialization/BinaryDecoderWithLRUCache.scala
massie: There are some issues when using Guava cache here.

1. The `cache.getIfPresent()` method has a key type of `AnyRef`. This allows you to miss data in the cache if you use a different type (e.g. String vs Utf8).
2. I like the `Callable` getter but it doesn't work for us here. We are reusing an `Utf8` object and, if the key doesn't exist in the cache, we make a copy and store it. This can't be done with the `Callable` get unless we use `getIfPresent` first.

heuermh:
@massie would you like this to be merged before 0.18.2 or 0.19.0, or will it be superseded by other performance changes you have in the works?

massie:
This will be superseded by other changes in the works... I'm going to go ahead and close it.

heuermh:
Thanks!

PR #859, opened 2015-10-08
[ADAM-857] Corrected handling of env vars in bin scripts

### LINKED ISSUE #857: adam-submit fails with no Spark installation
If I don't set `SPARK_HOME` and I don't have `spark-submit` on my `PATH`, then `adam-submit` errors out early:

```
$ bin/adam-submit 
Using ADAM_MAIN=org.bdgenomics.adam.cli.ADAMMain
```

I believe it's because the subshell call to `which spark-submit` returns an error code and `adam-submit` calls `set -e`.

laserson:
Because of `set -e`, the `which spark-submit` subshell command was causing the script to exit early.

Fixes #857.


heuermh:
Works for me, thank you @laserson

PR #866, opened 2015-10-13
[ADAM-864] Don't force shuffle if reducing partition count.

### LINKED ISSUE #864: Coalesce option should not force shuffle if reducing partition count
In several of the CLI commands, we have a `-coalesce` option that allows the user to choose the number of output partitions. We run `coalesce(_, true)` which forces a shuffle. Instead, we should check whether or not the `coalesce` will increase or decrease the partition count. If it will decrease the partition count, we shouldn't pass the `true` flag unless the user wants to force a shuffle.

fnothaft:
Resolves #864. In Spark, coalescing will reduce the number of partitions in an
RDD without performing a shuffle, but coalescing will only increase the number
of partitions if a shuffle is performed. This PR modifies Transform and Vcf2ADAM
to check whether the coalesce option will increase or decrease the partition
count. Additionally, it adds a flag that allows the user to force a shuffle;
this may be desirable as this causes a HashPartitioned shuffle, which may
improve the balance of records across partitions.

### DIFF
--- a/adam-cli/src/main/scala/org/bdgenomics/adam/cli/ADAM2Fasta.scala
+++ b/adam-cli/src/main/scala/org/bdgenomics/adam/cli/ADAM2Fasta.scala
@@ -27,10 +27,10 @@ import org.bdgenomics.utils.cli._
 import org.kohsuke.args4j.{ Argument, Option => Args4JOption }
 
 class ADAM2FastaArgs extends ParquetLoadSaveArgs {
-  @Args4JOption(required = false, name = "-partitions", usage = "Number of partitions, default 1")
-  var partitions: Int = 1
-  @Args4JOption(required = false, name = "-shuffle", usage = "Shuffle while partitioning, default false")
-  var shuffle: Boolean = false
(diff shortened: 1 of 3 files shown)


heuermh:
Could you add ```ADAM2Fasta``` in this pull request?  It uses ```-partitions``` and ```-shuffle``` as argument names instead of ```-coalesce``` and ```-force_shuffle_coalesce```, respectively.

fnothaft:
@heuermh Done!

heuermh:
LGTM, thanks.

heuermh:
Can I merge this?  If so, is the Merge github button still ok?  If I remember correctly, the ```commit-pr.sh``` script won't work for me because I'm using git over https.

fnothaft:
I thought I'd pushed a fix for the commit-pr.sh issue? Let me rebase this first... or, do you just want me to merge it?

heuermh:
```commit-pr.sh``` hasn't been revised yet.

Rebase and I'll hit the button, no need to bend the don't-merge-your-own-pull-requests rule.  :)

fnothaft:
Rebased!

heuermh:
Thanks!

PR #877, opened 2015-11-04
Minor fix to commit script to support https.

fnothaft:
@heuermh this should fix the issue you were seeing.


heuermh:
LGTM, thanks

PR #884, opened 2015-11-18
[ADAM-883] Add caching to Transform pipeline.

### LINKED ISSUE #883: Transform DAG causes stages to recompute
We need to add caching before Indel Realignment and BQSR when running multiple stages in a pipeline.

fnothaft:
The Transform pipeline in the CLI has several stages (e.g., sort, indel
realignment, BQSR) that trigger recomputation. If you are running a single
stage off of local storage/HDFS/Tachyon, this is OK. However, if you're running
multiple stages, or you are loading data from S3/etc, this can lead to serious
performance degradation. To address this, I've added the proper caching
statements. Additionally, I've added a hook so that the user can specify the
storage level to use for caching. Resolves #883.

### DIFF
--- a/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
+++ b/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
@@ -19,8 +19,9 @@ package org.bdgenomics.adam.cli
 
 import htsjdk.samtools.ValidationStringency
 import org.apache.parquet.filter2.dsl.Dsl._
-import org.apache.spark.rdd.RDD
 import org.apache.spark.{ Logging, SparkContext }
+import org.apache.spark.rdd.RDD
+import org.apache.spark.storage.StorageLevel
 import org.bdgenomics.adam.algorithms.consensus._
 import org.bdgenomics.adam.instrumentation.Timers._
 import org.bdgenomics.adam.models.SnpTable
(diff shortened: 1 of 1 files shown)


adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
```diff
@@ -105,6 +106,10 @@ class TransformArgs extends Args4jBase with ADAMSaveAnyArgs with ParquetArgs {
   var mdTagsFragmentSize: Long = 1000000L
   @Args4jOption(required = false, name = "-md_tag_overwrite", usage = "When adding MD tags to reads, overwrite existing incorrect tags.")
   var mdTagsOverwrite: Boolean = false
+  @Args4jOption(required = false, name = "-cache", usage = "Cache data to avoid recomputing between stages.")
+  var cache: Boolean = false
+  @Args4jOption(required = false, name = "-storageLevel", usage = "Set the storage level to use for caching.")
```
heuermh: ```-storageLevel``` &rarr; ```-storage_level```?

fnothaft: Good catch; will fix. Thanks!

fnothaft:
Fixed nit and rebased.

heuermh:
Thanks!

PR #885, opened 2015-11-18
Fix tightlist bug in Pandoc

massie:


fnothaft:
Merged! Thanks @massie.

PR #901, opened 2015-12-22
Single file save from #733, rebased

fnothaft:
This is a rebase of the code contributed by @antonstamov as PR #733, but rebased onto the latest changes, and with unit tests added for the single file save code.


fnothaft:
I just tested this on the NA12878 high coverage BAM from 1000 Genomes on the cluster, and it looks great. I validated the output with FlagStat; not necessarily a comprehensive test, but should be close enough. That being said, since we need to run an aggregate before converting to BAM, performance is really bad if you sort then save. I'll include this in an update shortly.

The build issue is because `fullOuterJoin` was not in Spark 1.2.1. There's a simple workaround for this.

heuermh:
ok to merge now, or wait for performance update?

fnothaft:
The perf fix is in #906 --> you don't need to run the aggregate to reconstruct the `SequenceDictionary` and `RecordGroupDictionary`.

PR #906, opened 2015-12-27
ADAM-599, 905: Move to bdg-formats:0.7.0 and migrate metadata

### LINKED ISSUE #599: Fix record oriented shuffle
Due to our shuffle being record oriented, we experience an approximately 8-10x increase in data volume when we shuffle. This is because our data is stored on disk in a columnar representation, but is shuffled in a row oriented format.

fnothaft:
Resolves #599 and #905:

* Moves to bdg-formats:0.7.0, where the recordGroup metadata fields have been eliminated from the `AlignmentRecord` schema.
* Adds code so that `loadAlignments` always returns Sequence and RecordGroup dictionaries.
* Supports the loading/storage of Sequence/RecordGroup dictionaries by writing them to disk as Avro files using the `Contig` and `RecordGroupMetadata` records from bdg-formats.
* Scala/javadoc cleanup related to the above changes.


adam-cli/src/main/scala/org/bdgenomics/adam/cli/Adam2Fastq.scala
```diff
@@ -70,7 +70,7 @@ class Adam2Fastq(val args: Adam2FastqArgs) extends BDGSparkCommand[Adam2FastqArg
       else
         None
 
-    var reads: RDD[AlignmentRecord] = sc.loadAlignments(args.inputPath, projection = projectionOpt)
+    var reads: RDD[AlignmentRecord] = sc.loadAlignments(args.inputPath, projection = projectionOpt)._1
```
heuermh: I would rather this primary API entry point didn't return a tuple.  How about a new load method that returns the tuple, and keep this one returning the RDD?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -235,7 +235,23 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+   * @param filePath Path to the file on disk.
+   *
+   * @return Returns a Tuple3 containing (an RDD of reads, the sequence
+   *   dictionary representing the contigs these reads are aligned to
+   *   if the reads are aligned, the record group dictionary for the reads
+   *   if one is available).
+   *
+   * @see loadAlignments
+   */
+  def loadBam(filePath: String): (RDD[AlignmentRecord], SequenceDictionary, RecordGroupDictionary) = {
```
heuermh: same here...

fnothaft: I'd prefer to funnel users towards loading/saving the tuple. My reasoning is that the tuple is the full set of information you need to reconstruct all the alignment data (the reads themselves + information about the sequencing process that generated the reads + information about the sequences the reads were aligned to). There _are_ places where we don't use all of the tuple, but the cost of loading the second and third elements of the tuple should be small relative to the cost of loading the RDD.

I'd like to understand why you'd like the primary entrypoint to be loading just the RDD. What are the pros/cons from your side?

heuermh: Evidenced by the number of places ```._1``` had to be added, this is a rather breaking change, even though we aren't committed to binary compatibility at this point.

My concern though is more from the story telling point -- the main abstraction in Spark is the RDD and ADAM supports loading those from native bioinformatics file formats, e.g. you load a BAM file and get an RDD of AlignmentRecords.  Except now that's not the method signature because there's some other stuff too.

How is that for a weak argument?  :)

fnothaft: > Evidenced by the number of places ._1 had to be added, this is a rather breaking change, even though we aren't committed to binary compatibility at this point.

IIRC, most of the places `._1` had to be added were in tests. I did add it to a few of the CLI commands, but I think that's because most of the commands that we have aren't _transformations_. If you're transforming the data, then you frequently need the sequence/record group dictionaries. If you're just running an aggregation or you're viewing a file, then you may not need the dictionaries.

> My concern though is more from the story telling point -- the main abstraction in Spark is the RDD and ADAM supports loading those from native bioinformatics file formats, e.g. you load a BAM file and get an RDD of AlignmentRecords. Except now that's not the method signature because there's some other stuff too.

I don't think it changes the story really. The sequence/record group dictionaries are just secondary tables of metadata that you frequently want to go with your reads. We could represent them as RDDs if we wanted to, but there's not really a good reason to do that because they're very small (for a single sample) and you're typically going to want to manipulate them on the driver (for sequence dictionaries) or do broadcast joins (with record group dictionaries).

heuermh: I share the opinion from the Google Guava folks that [tuple types are awful obfuscators](https://github.com/google/guava/wiki/IdeaGraveyard#tuples-for-n--2).  It is off-putting for non-Scala-natives and makes for a weak API.  What if the number or order of values in the tuple changes over time?  Could the tuple be instead something with a name?

In either case I'd still like to keep the RDD API, e.g.
```scala
var alignmentRecords: RDD[AlignmentRecord]
  = sc.loadAlignments(path, projection)

var (alignmentRecords, sequenceDictionary, readGroupDictionary)
  = sc.loadAlignmentsAndOtherStuff(path, projection)

// or

var alignmentRecordsAndOtherStuff: AlignmentRecordsAndOtherStuff
  = sc.loadAlignmentsAndOtherStuff(path, projection)
```

fnothaft: > I share the opinion from the Google Guava folks that tuple types are awful obfuscators. It is off-putting for non-Scala-natives and makes for a weak API. What if the number or order of values in the tuple changes over time? Could the tuple be instead something with a name?

Thanks for suggesting this. I think you've got a good point, and I think it'd be a straightforward refactor to support. I'll add a commit that proposes a better interface.

> In either case I'd still like to keep the RDD API, e.g.

I'm opposed to your proposal for the reasons I gave above (not strongly, I'm something around a -0 to -0.5) but I think the two of us are kind of going in circles here. Perhaps let's wait and get more comments from folks (CC @laserson @massie @tdanford @ryan-williams) and see what the consensus is. If

laserson:
I may be a little out of context, since I've been offline for a month.  I tend to be on the side of doing less magic for the user.  Will returning such a tuple turn out to be a hassle if a user does some more exotic things other than just vanilla genome sequencing?  Will there be situations where the additional non-read objects will be dummies?  I would also support avoiding a tuple and replacing with a named type in case we go down that path.

fnothaft:
OK, rebased and added code to address the tuple method signature in a new commit (999f48f44d4b82cef9c9692f7343b62d2326023b).

heuermh:
@fnothaft thank you for the update!  I don't see the source for ```GenomicRDD``` in the diff, did I miss that?

fnothaft:
Added the missing file and squashed down to two commits (one per issue).

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -238,7 +245,23 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+  /**
+   * Loads a SAM/BAM file.
+   *
+   * This reads the sequence and record group dictionaries from the SAM/BAM file
+   * header. SAMRecords are read from the file and converted to the
+   * AlignmentRecord schema.
+   *
+   * @param filePath Path to the file on disk.
+   *
+   * @return Returns a Tuple3 containing (an RDD of reads, the sequence
```
heuermh: Tuple3 &rarr; an aligned read RDD (a tuple of ...

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -346,15 +371,126 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+  }
+
+  /**
+   * Loads alignment data from a Parquet file.
+   *
+   * @param filePath The path of the file to load.
+   * @param predicate An optional predicate to push down into the file.
+   * @param projection An optional schema designating the fields to project.
+   *
+   * @return Returns a Tuple3 containing (an RDD of reads, the sequence
```
heuermh: ...same here

adam-core/src/test/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctionsSuite.scala
```diff
@@ -122,12 +123,18 @@ class AlignmentRecordRDDFunctionsSuite extends ADAMFunSuite {
...
 
     val tempFile = Files.createTempDirectory("reads12")
-    rdd12A.adamSAMSave(tempFile.toAbsolutePath.toString + "/reads12.sam", asSam = true)
+    rdd12A.adamSAMSave(tempFile.toAbsolutePath.toString + "/reads12.sam",
+      sd,
+      rgd,
+      asSam = true)
 
-    val rdd12B: RDD[AlignmentRecord] = sc.loadBam(tempFile.toAbsolutePath.toString + "/reads12.sam/part-r-00000")
+    val rdd12B = sc.loadBam(tempFile.toAbsolutePath.toString + "/reads12.sam/part-r-00000").rdd
```
heuermh: Both of these still work for me
```scala
val rdd12B = sc.loadBam(tempFile.toAbsolutePath.toString + "/reads12.sam/part-r-00000")                                                                                        
val rdd12B: RDD[AlignmentRecord] = sc.loadBam(tempFile.toAbsolutePath.toString + "/reads12.sam/part-r-00000") 
```
I assume this is the implicit at work, and the ```.rdd``` isn't actually necessary here?

fnothaft: Either way should be fine. I think I changed `rdd12B` to be consistent with `rdd12A`.

heuermh:
Even though this makes a lot of changes, some of which might be binary-incompatible (when it comes to the implicit stuff with scala I'm not too sure), my downstream projects still work.

+1 from me after minor doc fixes mentioned above.

fnothaft:
@heuermh fixed the doc issues.

heuermh:
Thank you, @fnothaft!

Issue #721, opened 2015-07-01
Writing to a BAM file with adamSAMSave consistently fails


danvk:
I'm running [this code][1] on a yarn cluster. It's trying to filter a BAM file to just those alignments which are either on chr22 or have a mate on chr22.

```scala
override def run(args: Arguments, sc: SparkContext): Unit = {
  val filterContig = args.filterContig
  val alignments = sc.loadAlignments(args.reads)
  val matchingAlignments = alignments.filter(matchesContig(_, filterContig))
  matchingAlignments.persist()
  println("Found " + matchingAlignments.count() + " alignments with   one pair in " + filterContig)
  matchingAlignments.coalesce(10).adamSAMSave(args.outputPath, asSam = false)
}
```

I'm consistently getting this error:

> Exception in thread "main" org.apache.spark.SparkException: Job aborted due to stage failure: Task 5 in stage 5.0 failed 4 times, most recent failure: Lost task 5.3 in stage 5.0 (TID 706, demeter-csmaz08-10.demeter.hpc.mssm.edu): java.lang.AssertionError: assertion failed: Cannot return header if not attached.

My command line is this:

> spark-submit --master yarn --deploy-mode client --executor-memory 16g --driver-memory 10g --num-executors 1000 --executor-cores 1 --driver-java-options "-Dyarn.resourcemanager.am.max-attempts=1 -Dlog4j.configuration=scripts/log4j.properties" --class org.hammerlab.guacamole.Guacamole --verbose target/guacamole-with-dependencies-0.0.1-SNAPSHOT.jar structural-variant --reads hdfs:///datasets/dream/data/synthetic-challenge-4/synthetic.challenge.set4.tumour.bam --filter-contig 22 --out hdfs:///user/vanded03/synth4.tumor.chr22+mate.bam

(the input is from the dream challenge)

Would this be expected to work? cc @ryan-williams 

[1]: https://gist.github.com/danvk/d37fa75be381553b6d70

arahuja:
I was seeing the same issue in #676 - which was apparently fixed, but I haven't checked since.

danvk:
@arahuja I believe that issue was specifically when you used `.coalesce(1)`. I ran out of memory when I tried that, so I'm using `.coalesce(10)` and running into this issue.

ryan-williams:
Closing as a ~dupe of #676

PR #917, opened 2016-01-12
[ADAM-916] New strategy for writing header.

### LINKED ISSUE #916: BAM header is not getting set on partition 0 with headerless BAM output format
The bug that will not die... Reported by @almussel. See #676, #691, #711, #712, #721...

```
WARNING:toil.leader:8ecfbeab-88ba-45e6-a3a4-ff1270e13051:       16/01/11 22:40:30 WARN scheduler.TaskSetManager: Lost task 0.0 in stage 4.0 (TID 4, 172.31.16.110): java.lang.AssertionError: assertion failed: Cannot return header if not attached.
WARNING:toil.leader:8ecfbeab-88ba-45e6-a3a4-ff1270e13051:               at scala.Predef$.assert(Predef.scala:179)
WARNING:toil.leader:8ecfbeab-88ba-45e6-a3a4-ff1270e13051:               at org.bdgenomics.adam.rdd.read.ADAMBAMOutputFormat$.getHeader(ADAMBAMOutputFormat.scala:60)
WARNING:toil.leader:8ecfbeab-88ba-45e6-a3a4-ff1270e13051:               at org.bdgenomics.adam.rdd.read.ADAMBAMOutputFormat.<init>(ADAMBAMOutputFormat.scala:68)
WARNING:toil.leader:8ecfbeab-88ba-45e6-a3a4-ff1270e13051:               at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
WARNING:toil.leader:8ecfbeab-88ba-45e6-a3a4-ff1270e13051:               at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:57)
WARNING:toil.leader:8ecfbeab-88ba-45e6-a3a4-ff1270e13051:               at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
WARNING:toil.leader:8ecfbeab-88ba-45e6-a3a4-ff1270e13051:               at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
WARNING:toil.leader:8ecfbeab-88ba-45e6-a3a4-ff1270e13051:               at java.lang.Class.newInstance(Class.java:383)
WARNING:toil.leader:8ecfbeab-88ba-45e6-a3a4-ff1270e13051:               at org.apache.spark.rdd.InstrumentedOutputFormat.<init>(InstrumentedOutputFormat.scala:33)
WARNING:toil.leader:8ecfbeab-88ba-45e6-a3a4-ff1270e13051:               at org.bdgenomics.adam.rdd.read.InstrumentedADAMBAMOutputFormat.<init>(ADAMBAMOutputFormat.scala:71)
WARNING:toil.leader:8ecfbeab-88ba-45e6-a3a4-ff1270e13051:               at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
WARNING:toil.leader:8ecfbeab-88ba-45e6-a3a4-ff1270e13051:               at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:57)
WARNING:toil.leader:8ecfbeab-88ba-45e6-a3a4-ff1270e13051:               at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
WARNING:toil.leader:8ecfbeab-88ba-45e6-a3a4-ff1270e13051:               at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
WARNING:toil.leader:8ecfbeab-88ba-45e6-a3a4-ff1270e13051:               at java.lang.Class.newInstance(Class.java:383)
WARNING:toil.leader:8ecfbeab-88ba-45e6-a3a4-ff1270e13051:               at org.apache.spark.rdd.PairRDDFunctions$anonfun$saveAsNewAPIHadoopDataset$1$anonfun$12.apply(PairRDDFunctions.scala:1020)
WARNING:toil.leader:8ecfbeab-88ba-45e6-a3a4-ff1270e13051:               at org.apache.spark.rdd.PairRDDFunctions$anonfun$saveAsNewAPIHadoopDataset$1$anonfun$12.apply(PairRDDFunctions.scala:1014)
WARNING:toil.leader:8ecfbeab-88ba-45e6-a3a4-ff1270e13051:               at org.apache.spark.scheduler.ResultTask.runTask(ResultTask.scala:66)
WARNING:toil.leader:8ecfbeab-88ba-45e6-a3a4-ff1270e13051:               at org.apache.spark.scheduler.Task.run(Task.scala:88)
WARNING:toil.leader:8ecfbeab-88ba-45e6-a3a4-ff1270e13051:               at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:214)
WARNING:toil.leader:8ecfbeab-88ba-45e6-a3a4-ff1270e13051:               at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
WARNING:toil.leader:8ecfbeab-88ba-45e6-a3a4-ff1270e13051:               at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
WARNING:toil.leader:8ecfbeab-88ba-45e6-a3a4-ff1270e13051:               at java.lang.Thread.run(Thread.java:745)
```

@almussel will get us Spark logs tomorrow.

fnothaft:
Resolves #916. Not fully done yet; seems to work for BAM but not SAM. The SAM file header doesn't write correctly... Still debugging that. Needs more testing.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala
@@ -17,8 +17,15 @@
  */
 package org.bdgenomics.adam.rdd.read
 
-import java.io.{ OutputStream, StringWriter }
-import htsjdk.samtools.{ SAMFileHeader, SAMTextHeaderCodec, SAMTextWriter, ValidationStringency }
+import java.io.{
+  InputStream,
+  OutputStream,
+  StringWriter,
+  Writer
+}
+import htsjdk.samtools._
+import htsjdk.samtools.util.{ BinaryCodec, BlockCompressedOutputStream }
+import java.lang.reflect.InvocationTargetException
 import org.apache.avro.Schema
 import org.apache.avro.file.DataFileWriter
 import org.apache.avro.specific.{ SpecificDatumWriter, SpecificRecordBase }
@@ -28,7 +35,7 @@ import org.apache.hadoop.io.LongWritable
 import org.apache.spark.SparkContext
 import org.apache.spark.broadcast.Broadcast
 import org.apache.spark.rdd.MetricsContext._
-import org.apache.spark.rdd.{ PartitionPruningRDD, RDD }
+import org.apache.spark.rdd.RDD
 import org.apache.spark.storage.StorageLevel
 import org.bdgenomics.adam.algorithms.consensus.{ ConsensusGenerator, ConsensusGeneratorFromReads }
 import org.bdgenomics.adam.converters.AlignmentRecordConverter
@@ -42,9 +49,9 @@ import org.bdgenomics.adam.rich.RichAlignmentRecord
 import org.bdgenomics.adam.util.MapTools
 import org.bdgenomics.formats.avro._
 import org.seqdoop.hadoop_bam.SAMRecordWritable
-import scala.reflect.ClassTag
-
+import scala.annotation.tailrec
 import scala.language.implicitConversions
+import scala.reflect.ClassTag
 
 class AlignmentRecordRDDFunctions(rdd: RDD[AlignmentRecord])
     extends ADAMSequenceDictionaryRDDAggregator[AlignmentRecord](rdd) {
@@ -297,7 +304,7 @@ class AlignmentRecordRDDFunctions(rdd: RDD[AlignmentRecord])
     rgd: RecordGroupDictionary,
     asSam: Boolean = true,
     asSingleFile: Boolean = false,
-    isSorted: Boolean = false) = SAMSave.time {
+    isSorted: Boolean = false): Unit = SAMSave.time {
 
     // if the file is sorted, make sure the sequence dictionary is sorted
     val sdFinal = if (isSorted) {
@@ -314,51 +321,51 @@ class AlignmentRecordRDDFunctions(rdd: RDD[AlignmentRecord])
     // add keys to our records
     val withKey = convertRecords.keyBy(v => new LongWritable(v.get.getAlignmentStart))
 
-    val bcastHeader = rdd.context.broadcast(header)
-    val mp = rdd.mapPartitionsWithIndex((idx, iter) => {
-      log.info(s"Setting ${if (asSam) "SAM" else "BAM"} header for partition $idx")
-      val header = bcastHeader.value
-      synchronized {
-        // perform map partition call to ensure that the SAM/BAM header is set on all
-        // nodes in the cluster; see:
-        // https://github.com/bigdatagenomics/adam/issues/353,
-        // https://github.com/bigdatagenomics/adam/issues/676
-
-        asSam match {
-          case true =>
-            ADAMSAMOutputFormat.clearHeader()
-            ADAMSAMOutputFormat.addHeader(header)
-            log.info(s"Set SAM header for partition $idx")
-          case false =>
-            ADAMBAMOutputFormat.clearHeader()
-            ADAMBAMOutputFormat.addHeader(header)
-            log.info(s"Set BAM header for partition $idx")
-        }
-      }
-      Iterator[Int]()
-    }).count()
-
-    // force value check, ensure that computation happens
-    if (mp != 0) {
-      log.error("Had more than 0 elements after map partitions call to set VCF header across cluster.")
-    }
-
-    // attach header to output format
-    asSam match {
-      case true =>
-        ADAMSAMOutputFormat.clearHeader()
-        ADAMSAMOutputFormat.addHeader(header)
-        log.info("Set SAM header on driver")
-      case false =>
-        ADAMBAMOutputFormat.clearHeader()
-        ADAMBAMOutputFormat.addHeader(header)
-        log.info("Set BAM header on driver")
-    }
-
     // write file to disk
     val conf = rdd.context.hadoopConfiguration
 
     if (!asSingleFile) {
+      val bcastHeader = rdd.context.broadcast(header)
+      val mp = rdd.mapPartitionsWithIndex((idx, iter) => {
+        log.info(s"Setting ${if (asSam) "SAM" else "BAM"} header for partition $idx")
+        val header = bcastHeader.value
+        synchronized {
+          // perform map partition call to ensure that the SAM/BAM header is set on all
+          // nodes in the cluster; see:
+          // https://github.com/bigdatagenomics/adam/issues/353,
+          // https://github.com/bigdatagenomics/adam/issues/676
+
+          asSam match {
+            case true =>
+              ADAMSAMOutputFormat.clearHeader()
+              ADAMSAMOutputFormat.addHeader(header)
+              log.info(s"Set SAM header for partition $idx")
+            case false =>
+              ADAMBAMOutputFormat.clearHeader()
+              ADAMBAMOutputFormat.addHeader(header)
+              log.info(s"Set BAM header for partition $idx")
+          }
+        }
+        Iterator[Int]()
+      }).count()
+
+      // force value check, ensure that computation happens
+      if (mp != 0) {
+        log.error("Had more than 0 elements after map partitions call to set VCF header across cluster.")
+      }
+
+      // attach header to output format
+      asSam match {
+        case true =>
+          ADAMSAMOutputFormat.clearHeader()
+          ADAMSAMOutputFormat.addHeader(header)
+          log.info("Set SAM header on driver")
+        case false =>
+          ADAMBAMOutputFormat.clearHeader()
+          ADAMBAMOutputFormat.addHeader(header)
+          log.info("Set BAM header on driver")
+      }
+
       asSam match {
         case true =>
           withKey.saveAsNewAPIHadoopFile(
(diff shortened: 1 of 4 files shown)


fnothaft:
This is good to go now. Ping for review and merge.

fnothaft:
OK, it seems like on Spark 1.2 we got a merging order that is inconsistent with the ordering we're expecting. I was worried about that... Let me look into this more and see if there is a way we can ensure a consistent merging order.

fnothaft:
Rebased and somewhat fixed.

heuermh:
Single unit test failure
```none
Couldn't find any files matching /tmp/bqsr1442238865989322732/bqsr1.bam
```

fnothaft:
Jenkins, retest this please.

fnothaft:
Jenkins, test this please.

fnothaft:
Jenkins, test this please.

fnothaft:
Pushed a fix; tests should pass on this run. From my side, this PR is good to go (assuming tests pass on @AmplabJenkins).

fnothaft:
Good news: the `concat` method doesn't exist in `org.apache.hadoop.fs.FileSystem` in the Hadoop 1.x stream. I've added code that calls `concat` via reflection. I'm testing this on the cluster now.

fnothaft:
This all works, except for the `concat` operation, which fails on the cluster because the header doesn't take up an exact, full HDFS block. Anyways, we fall back on the single threaded writing code, which is not so hot from a performance perspective, but works correctly.

fnothaft:
Jenkins, retest this please.

fnothaft:
Also, even when we fall back on the single threaded copy code, it takes ~1sec per file to merge the files. E.g., if we have ~2000 partitions, then it takes about 2000 seconds to merge the whole file together. This is unfortunate, but is much better than the ~30hrs latency @beaunorgeot had been seeing when doing something like `samtools cat` on a single machine.

ryan-williams:
Thanks for tackling this @fnothaft; I'm missing state on this, can you answer a few high-level questions:
* do you know how the missing header issue is currently happening? Last I looked at this (many months ago), we had what we thought was a high-confidence fix, and yet I still saw it happen once (but not reproducibly).
* in this fix, you're writing out the header to a path that is specified in the hadoop job config, and having each partition read the header in from that path?
* it seems like all the existing {add,clear}Header code is still present, but has been moved around; how does that interact with the new writing/reading-header-file stuff?

fnothaft:
OK, this succeeded on the cluster. Save time is slower than I'd like (~40 min to cat NA12878 together) but it worked. I'm running another test now that sorts then saves, then I'll pull the file out of HDFS and run samtools index locally. If that succeeds, then we are good to go. I'll update when that's done; I'm guessing it'll probably finish sometime late tonight.

fnothaft:
`samtools index` worked on the cluster, so we are both writing a concatenated file _and_ keeping the sort order. Yippee!

@ryan-williams 

> do you know how the missing header issue is currently happening? Last I looked at this (many months ago), we had what we thought was a high-confidence fix, and yet I still saw it happen once (but not reproducibly).

God, I still have no idea what's causing it. I agree—I thought our fix was high confidence, but the log files are pretty unambiguous that we're getting a partition that tries to write to disk but the header is not set.

> in this fix, you're writing out the header to a path that is specified in the hadoop job config, and having each partition read the header in from that path?

Exactly. It turns out that this is how Hadoop-BAM's CLI uses the Hadoop-BAM output formats. This isn't obvious though... I got the idea for this just by reading the Hadoop-BAM CLI source code.

> it seems like all the existing {add,clear}Header code is still present, but has been moved around; how does that interact with the new writing/reading-header-file stuff?

I only touched the headerless output formats, so they're still vulnerable to the transient header disappearance issue. I'll need to propagate the fix to all output formats later. We'll probably have to do the same thing for VCF as well.

fnothaft:
Anyways, this is good to merge now.

heuermh:
@ryan-williams can you review?

I'll take a look and run a few extra tests this morning.  We would like to get this merged as soon as possible.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala
```diff
@@ -380,53 +387,210 @@ class AlignmentRecordRDDFunctions(rdd: RDD[AlignmentRecord])
...
+        .sortBy(_.getName)
+        .toArray
+
+      // try to merge this via the fs api, which should guarantee ordering...?
+      // however! this function is not implemented on all platforms, hence the try.
+      try {
+
+        // concat files together
+        // we need to do this through reflection because the concat method is
+        // NOT in hadoop 1.x
```
heuermh: after this is merged, we should create an issue to revisit it once we drop Hadoop 1.x support

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala
```diff
@@ -380,53 +387,210 @@ class AlignmentRecordRDDFunctions(rdd: RDD[AlignmentRecord])
...
+            // increment file copy count
+            filesCopied += 1
+          })
+
+          // flush and close the output stream
+          os.flush()
+          os.close()
+
+          // delete temp files
+          //fs.delete(headPath, true)
```
heuermh: I see temp files in ```tailPath``` are deleted above but don't see any call to delete ```headPath```

ryan-williams:
Unfortunately I don't have much time to get in to the weeds on this one, but my main thought is:

if we don't know why the previous fix didn't fix the issue, even though we all still believe it should have from everything we know about the code, then I don't like complicating matters in this way.

Ultimately, I'll defer to y'all since it sounds like the goal here is more to unblock specific user(s) than to have a fix that we understand, but obviously that doesn't seem like a great state of affairs longer-term, especially if this introduces a slow-down to boot.

Can we deterministically repro the crash? I agree that something is still wrong, @danvk (on #721) and I (undocumented?) have both seen the bug since the last "fix", IIRC.

fnothaft:
The perf is only bad when saving BAM as a single file, and "bad" is relative to the FileSystem 'concat' method. This is a correctness fix (block ordering + header attach) relative to the old 'FileUtil.copyMerge' approach. I agree that we need to harmonize the single file save/many file save approaches, but that's for later.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDFunctions.scala
fnothaft: Both are removed here... Although, I commented those lines out when debugging something; let me fix that real quick.

Are you referring to the code several lines above at https://github.com/bigdatagenomics/adam/pull/917/files#diff-a0df6432a6d340425e3510af17cb8bd2R516 ? The headPath file there doesn't need to be removed because the `concat` merges all of the tail files _into_ the head file. We then move the head file to the desired output path.

fnothaft:
This is good to go from my side. @heuermh were you able to complete your testing?

Just to follow up more on @ryan-williams' comments from earlier (sorry, I was on the run so I kept my comments brief), this fix _only_ impacts the single file save code, and does not lead to a performance decrease over the _current_ single file save code. This fix does three things:

1. Writes the file header to a file in HDFS (or whichever underlying file system you are using; I'll use HDFS as a placeholder here for convenience).
2. Adds code to the headerless output formats that ensures that they read the header from HDFS. This code will never fail unless writing the header to HDFS failed in the first place, in which case you've got bigger problems and I can't help you ;)
3. Modifies the file merging code. This introduces a fast path to merging using the `concat` method, and a second path to merging that manually merges the files together. The second path (manual merge) is necessary anyways, because the prior method using `FileUtil.copyMerge` did not guarantee the ordering of files that were being merged, which meant that we'd occasionally get files where the header was at the end, etc.

The manual merge code path is much slower than the `concat` method, but the `concat` method has harsh, undocumented preconditions (specifically, all files being merged—except for the last one—must take up a full block in HDFS) and seems to throw a NotImplementedError on file systems that aren't HDFS, which is dandy. The manual merge code path is _not_ slower than the `FileUtil.copyMerge` approach we were using previously.

As for the million dollar question: why does the header detach issue keep on happening, despite our best efforts to fix it? Frankly, I'm not entirely sure why it's failing, but the root cause is something along the lines of "when we set the header in the singleton object that accompanies the output format, it doesn't get consistently set across all of the JVMs where an instance of the actual output format will be created and executed." The new approach uses the Hadoop job configuration to pass the path to the header file. This conf is passed to the OutputFormat by Spark itself, so we can guarantee that the configuration seen by the OutputFormat will be consistent across the cluster. Since the header is written to the underlying file system (which again, if your underlying file system is _not_ accessible across the whole cluster, then you're running some horrible frankencluster and I can't help you), we can guarantee that reading the header file from the path passed in the config will succeed.

I think a TL;DR for why this failed is that the obvious way of attaching the header (I call it obvious because the method is named `setHeader`) to Hadoop-BAM's OutputFormats is not guaranteed to work consistently in a distributed setting. I think that the Hadoop-BAM team knows this, because the Hadoop-BAM CLI wraps their own output formats to do this exact "write header to HDFS, set path in config, read header from config path when creating record reader" pattern.

fnothaft:
Jenkins, retest this please.

heuermh:
Tested ok for me, merged from the ski lift.

Thank you, @fnothaft!

PR #921, opened 2016-01-15
[ADAM-920] Limit tag/orig qual flags in Transform.

### LINKED ISSUE #920: Add back limit_projection on Transform
I'd removed this in #906. We should still not project the tags and original qualities fields.

fnothaft:
Resolves #920.


heuermh:
LGTM, thank you @fnothaft

PR #924, opened 2016-01-20
Changed fisher strand bias field for VCF2Adam from SB to FS

### LINKED ISSUE #923: Vcf2Adam uses SB field instead of FS field for fisher exact test for strand bias
https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantAnnotationConverter.scala#L86 is where the conversion attribute is defined.

Couldn't find any hard documentation on this field but I'm guessing that this SB is just the difference of the number of reads from the positive strand versus the negative strand.
I have a VCF from UnifiedGenotyper and the field that they use is for fisher exact test strand bias is
``##INFO=<ID=FS,Number=1,Type=Float,Description="Phred-scaled p-value using Fisher's exact test to detect strand bias">"``

andrewmchen:
Resolves #923. Tested on the cluster and works with VCF file generated from UnifiedGenotyper.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantAnnotationConverter.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantAnnotationConverter.scala
@@ -83,13 +83,13 @@ object VariantAnnotationConverter extends Serializable {
   )
 
   val INFO_KEYS: Seq[AttrKey] = Seq(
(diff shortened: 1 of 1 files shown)


heuermh:
Would both ```SB``` and ```FS``` be present in the same file?
Perhaps we should have both ```strandBias``` and ```fisherStrandBiasPValue``` as keys.

fnothaft:
+1

@heuermh you could have both, but you should be able to recompute strand bias from the Fisher score and the allelic depth.

fnothaft:
Thanks @andrewmchen! Committed as 0151b2d12facdb9de61b86888b2fe96ac886c8c9.

PR #926, opened 2016-01-25
Replacing symbolic test resource links with file copies

heuermh:
See
https://github.com/BD2KGenomics/cgl-docker-lib/pull/72#issuecomment-174602132


fnothaft:
Actually, do we even need to copy the files? I think if we include `adam-core`'s test jar in the test scope for the other two modules, we should be able to eliminate these files from the resource directories. I'd like to avoid having duplicate copies of files, esp. because with the SAM files, the headers sometimes need to be updated when htsjdk versions change.

heuermh:
That is possibly true, I don't know how we're reading files in test cases.  Wouldn't take much to read resources off the classpath though.

fnothaft:
Off of the top of my head, we load them from the classpath, which is why I think the test-jar approach would work.

heuermh:
Yep, just got there.  ```val path = resourcePath("small.sam")```.  Should only take a sec to use the test-jar.

heuermh:
Spits out these warnings
```
- unordered sam to unordered sam
2016-01-25 12:38:31 WARN  AlignmentRecordRDDFunctions:520 - Caught exception when merging via Hadoop FileSystem API:
java.lang.UnsupportedOperationException: Not implemented by the RawLocalFileSystem FileSystem implementation
2016-01-25 12:38:31 WARN  AlignmentRecordRDDFunctions:521 - Retrying as manual copy from the driver which will degrade performance.
```
but works fine.  Closing in favor of #927.

fnothaft:
Those warnings were added as part of https://github.com/bigdatagenomics/adam/pull/917/files. Specifically, the `concat` method in the Hadoop `FileSystem` class is much more efficient, but it's really fragile (only works on HDFS, requires all files to exactly fill blocks). If the `concat` path fails, we throw the warning message and retry via the slow but reliable path.

PR #927, opened 2016-01-25
Removing symbolic test resource links, read from test classpath instead

heuermh:
See
https://github.com/BD2KGenomics/cgl-docker-lib/pull/72#issuecomment-174602132

### DIFF
--- a/adam-apis/src/test/resources/small.sam
+++ b/adam-apis/src/test/resources/small.sam
@@ -1,1 +1,0 @@
-../../../../adam-core/src/test/resources/small.sam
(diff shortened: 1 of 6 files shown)


fnothaft:
LGTM once build is passing.

heuermh:
Sorry, I was foiled by classpath caching locally.  Thank you, Jenkins.

It appears that ```ADAMContext.loadBam``` can't handle classpath jar file paths.  Should I just copy the resource to a temporary file first?

fnothaft:
> It appears that ADAMContext.loadBam can't handle classpath jar file paths. Should I just copy the resource to a temporary file first?

Ah, yeah, I think I've run into that before... I've done exactly what you're talking about, although that was for a special case. If we'd need to do something hacky like that just to pass unit tests, this approach may not be worth doing, alas... I defer to you. Sorry—I thought this approach would work better...

fnothaft:
Thanks @heuermh ! Committed as efe4dfa0d7ba1bf6a4c08d435e88691c947b41b0.

PR #931, opened 2016-02-04
added start and end values to Interval Trait. Used for IntervalRDD

akmorrow13:


fnothaft:
Jenkins, add to whitelist.

fnothaft:
Jenkins, test this please.

fnothaft:
Also, +1 from me.

heuermh:
Do we still need width if we have start and end?  Do we need or want to support open ended intervals?  Intervals around the origin in circular genomes?  Zero-based or one-based coordinate systems?  etc.

fnothaft:
@heuermh 

> Do we still need width if we have start and end?

Would probably make sense to define `width = end - start`

> Do we need or want to support open ended intervals? Zero-based or one-based coordinate systems? etc.

We decided early on that we were going to be consistently 0-based, open ended inside of ADAM.

> Intervals around the origin in circular genomes? 

I wouldn't worry about this for now since no one's asked for it and it adds a fair bit of complexity, but it is something to consider later.

PR #935, opened 2016-02-09
[ADAM-934] Properly handle unset library name during duplicate marking

### LINKED ISSUE #934: MarkDuplicates fails if library name is not set
Reported by @jpdna. See https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/MarkDuplicates.scala#L73. If the library name is not set for a read group, we will call `.get` on a `None`, which throws an exception. The "correct" behavior is to treat the library ID as `null` in this case. This isn't ideal, but this is equivalent to our MarkDuplicates implementation pre-#906, and works correctly in the degenerate case where all reads are sequenced from a single library.

fnothaft:
Resolves #934. If the library name for a read group is not set, we will use a null string during the groupBy. This is equivalent to our pre-#906 implementation. Additionally, this commit adds logging that prints a warning message for the user if there are read groups whose library ID is not set.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/MarkDuplicates.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/MarkDuplicates.scala
@@ -17,6 +17,7 @@
  */
 package org.bdgenomics.adam.rdd.read
 
+import org.apache.spark.Logging
 import org.apache.spark.rdd.RDD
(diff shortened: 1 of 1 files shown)


jpdna:
+1

fnothaft:
Ping for merge.

heuermh:
Fixed via 883ac4d7474ed43c673707c0465096fe9972b7ae, thanks!

PR #939, opened 2016-02-12
adam2vcf doesn't have info fields

andrewmchen:
I added info fields only if the vcf is from a single sample.

What's the desired behavior if the vcf has multiple samples per site?

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantContextConverter.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantContextConverter.scala
@@ -33,6 +33,7 @@ import org.bdgenomics.adam.models.{
 import org.bdgenomics.adam.util.PhredUtils
 import org.bdgenomics.formats.avro._
 import scala.collection.JavaConversions._
+import scala.collection.mutable.HashMap
 
 object VariantContextConverter {
   private val NON_REF_ALLELE = Allele.create("<NON_REF>", false /* !Reference */ )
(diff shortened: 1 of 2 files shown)


fnothaft:
Whoops! You should have an info field per site, as well as sample specific annotations (per sample, per site), see https://samtools.github.io/hts-specs/VCFv4.2.pdf. Essentially, the info field behavior should be the same in the single and multi sample cases.

andrewmchen:
Hm I read through the VCF spec and I don't think I completely understand still. 

First, my understanding was that before this PR we wouldn't have any per site info field. Second, when convert a VCF to a RDD of adam Genotypes, are the VariantCallingAnnotations statistics not per sample but instead statistics aggregated across samples from that site?

heuermh:
@andrewmchen after we get 0.19.0 out, I plan to focus on VCF INFO handling, for supporting the VCF ANN spec and other related issues.  Please let me know how I might help in a week or so.

andrewmchen:
@heuermh Sure! Would love to work with you on some of the VCF features. Right now I'm just playing around with it for the purposes of avocado. 

Maybe this isn't the desired behavior, but I think it's kind of strange that adam2vcf returns a much different vcf than the original that was passed into vcf2adam. Are the related issues you mentioned related to this?

fnothaft:
I will open a follow on PR that addresses some issues with this (specifically, stores sample info in genotype, aggregate in INFO).

fnothaft:
@heuermh is this good to go on your side? I just looked in more detail and think it is fine (doesn't need any more patches). If it is good from your side, then I will merge.

heuermh:
LGTM

fnothaft:
Merged! Thanks @andrewmchen!

PR #943, opened 2016-02-13
[ADAM-942] Replace first/secondInRead with readInFragment

### LINKED ISSUE #942: readInFragment value is not properly read from parquet file into RDD[AlignmentRecord]
`readInFragment` field is not properly loaded from adam parquet file into an RDD when running `adam flagstat` due to a bug in defining the list of projected fields in FlagStat.scala [ here ](https://github.com/bigdatagenomics/adam/blob/master/adam-cli/src/main/scala/org/bdgenomics/adam/cli/FlagStat.scala#L61)
There are no longer `firstOfPair` / `secondOfPair` fields as these are replaced by values 0,1 in the `readInFragment` field.  Currently `readInFragment` field is ignored in input  and assumes its default value of 0 in the RDD.

We should check if this same type of error is repeated outside of FlagStat, and for any other fields.

To prevent this problem in future @fnothaft suggests in #941  autogeneration for field names enum used  in projections

heuermh:
Fixes #942


fnothaft:
LGTM!

jpdna:
Thanks for the commit @heuermh

PR #953, opened 2016-02-21
Use hadoop-bam BAMInputFormat to do loadIndexedBam

andrewmchen:
I changed the loadIndexedBam function to use the new InputFormat released in hadoop-bam 7.4.0 that filters using an index file. We used to use a InputFormat that @erictu wrote.

In order to make this change, I had to upgrade our htsjdk library to version [2.1.0](https://github.com/bigdatagenomics/adam/compare/master...andrewmchen:loadIndexedBam-hadoop-bam?expand=1#diff-600376dffeb79835ede4a0b285078036R30). Hope this doesn't break anything.

I haven't tested this on the cluster but I wrote a small test [here](https://github.com/bigdatagenomics/adam/compare/master...andrewmchen:loadIndexedBam-hadoop-bam?expand=1#diff-306c6dde689c4cbcb6d3e982dc401221R342). When I get a chance I'll try testing this on the cluster as well.


fnothaft:
Thanks for opening this, @andrewmchen! I will review soon. Moving to the new Hadoop-BAM/HTSJDK releases sounds good, but I would like to defer merging this to after the 0.19.0 release, as the version bump requires a move to Java 8. Since 0.19.0 has a lot of fixes in it, I'd like to keep that as widely available as possible.

andrewmchen:
OK great thanks! I'm guessing then that the compilation problems on Jenkins are because of the mismatch between Java 7 and 8?

fnothaft:
The unit test failure is showing a Java version mismatch, but I'd have to look into it more closely to say something more intelligent.

heuermh:
+1 to waiting until 0.20 or later

adam-core/src/main/scala/org/bdgenomics/adam/models/Interval.scala
```diff
@@ -41,4 +42,19 @@ trait Interval {
    */
   def width: Long = end - start
 
+  /**
+   * Need to implement getStart function from Locatable. (1-based start position. closed)
```
heuermh: Mixing 0- and 1- based coordinate systems in the same class gives me the hives.  Can we hide this as an implementation detail in an adapter somewhere in the i/o code?

adam-core/pom.xml
```diff
@@ -181,6 +181,7 @@
     <dependency>
       <groupId>org.seqdoop</groupId>
       <artifactId>hadoop-bam</artifactId>
+      <version>7.4.0</version>
```
heuermh: We'll probably do the update to Hadoop-BAM and HTSJDK in a separate pull request after taking a little time to solicit feedback on the move to JDK8.

pom.xml
```diff
@@ -511,7 +511,7 @@
       <dependency>
         <groupId>com.google.guava</groupId>
         <artifactId>guava</artifactId>
-        <version>16.0.1</version> <!-- note: version 17.0 breaks hadoop 2.6+ at runtime -->
+        <version>19.0</version> <!-- note: version 17.0 breaks hadoop 2.6+ at runtime -->
```
heuermh: I don't think we can make this change until Hadoop fixes the problem upstream.  Will try to find the relevant issue; I had a reference to it before.

fnothaft:
Jenkins, test this please.

fnothaft:
Superceded by #1036.

PR #970, opened 2016-03-17
adamGetReferenceString reduces pairs correctly, fixes #967

### LINKED ISSUE #967: adamGetReferenceString doesn't reduce pairs correctly
Throws the error `Regions being joined must be adjacent.` when reducing using the `reducePairs` function. 

This issue manifests when trying to reduce multiple pairs of NucleotideContigFragments. Spark's reduce operation requires the function to be commutative and associative when applied to the data, which isn't necessarily the case all the time. If we have 10 fragments, each of ReferenceRegion 10kb, sometimes we will end up reducing nonadjacent regions. 

To solve this issue, I implemented a solution that first collects the data and applies scala Array's reduceLeft operator so that the merged regions are always adjacent (shown below in NucleotideContigFragmentRDDFunctions). I'll submit a PR for it soon.

```
      val refPairRDD: RDD[(ReferenceRegion, String)] = rdd.keyBy(ReferenceRegion(_))
        .filter(kv => kv._1.isDefined)
        .map(kv => (kv._1.get, kv._2))
        .filter(kv => kv._1.overlaps(region))
        .sortByKey()
        .map(kv => getString(kv))

      val pair: (ReferenceRegion, String) = refPairRDD.collect.reduceLeft(reducePairs)
```

erictu:
fixes #967

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/contig/NucleotideContigFragmentRDDFunctions.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/contig/NucleotideContigFragmentRDDFunctions.scala
@@ -142,14 +142,14 @@ class NucleotideContigFragmentRDDFunctions(rdd: RDD[NucleotideContigFragment]) e
     }
 
     try {
-      val pair: (ReferenceRegion, String) = rdd.keyBy(ReferenceRegion(_))
+      val refPairRDD: RDD[(ReferenceRegion, String)] = rdd.keyBy(ReferenceRegion(_))
         .filter(kv => kv._1.isDefined)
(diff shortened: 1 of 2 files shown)


heuermh:
Thank you, @erictu.  Could you add a unit test demonstrating the issue?

erictu:
Done. Just to note, because the bug is nondeterministic (based on how Spark optimizes its reduce function), the unit test without the fixes doesn't always throw an error. With the new fixes, the test passes all the time.

heuermh:
Thanks, I understand now.  LGTM

heuermh:
Thank you, @erictu, merged as 3f21d8370e6891ad0ca053dc0c147c69eb255939

PR #976, opened 2016-03-24
add validation stringency to bam parsing, flagstat

### LINKED ISSUE #973: Add stringency parameter to flagstat

ryan-williams:
fixes #973

### DIFF
--- a/adam-cli/src/main/scala/org/bdgenomics/adam/cli/FlagStat.scala
+++ b/adam-cli/src/main/scala/org/bdgenomics/adam/cli/FlagStat.scala
@@ -17,15 +17,15 @@
  */
 package org.bdgenomics.adam.cli
 
-import org.apache.hadoop.fs.{ Path, FileSystem }
-import org.apache.hadoop.mapreduce.Job
+import htsjdk.samtools.ValidationStringency
+import org.apache.hadoop.fs.{ FileSystem, Path }
 import org.apache.spark.SparkContext
 import org.apache.spark.rdd.RDD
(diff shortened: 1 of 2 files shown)


adam-cli/src/main/scala/org/bdgenomics/adam/cli/FlagStat.scala
```diff
@@ -41,6 +41,8 @@ class FlagStatArgs extends Args4jBase with ParquetArgs {
   val inputPath: String = null
   @Argument(required = false, metaVar = "OUTPUT", usage = "Optionally write the stats to this file.", index = 1)
   val outputPath: String = null
+  @Args4jOption(required = false, name = "-stringency", usage = "Set the parsing stringency: SILENT, LENIENT, STRICT.")
+  val stringency: String = "SILENT"
```
heuermh: what is the default if SAMHeaderReader.VALIDATION_STRINGENCY_PROPERTY is not set?

ryan-williams: [STRICT](https://github.com/samtools/htsjdk/blob/1.139/src/java/htsjdk/samtools/ValidationStringency.java#L20). `flagstat` is a simple enough operation that i thought `SILENT` was more often than not what a user would want.

heuermh: ok, I agree with you on flagstat defaulting to SILENT, and I see the default in loadBam is STRICT.  +1

heuermh:
Thanks @ryan-williams!

PR #977, opened 2016-03-24
Remove unused file VcfHeaderUtils.scala

### LINKED ISSUE #428: VcfStringUtils unused?
I can't find any usages of [VcfStringUtils](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/util/VcfStringUtils.scala)

Also seems to be true for 

[VcfHeaderBuilder](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/util/VcfHeaderUtils.scala)

Was there a purpose for the StringUtils?  Seems like we could the VcfHeaderBuilder to generate a header specific to fields being written which does not happen now in `adamVCFSave`

heuermh:
Fixes #428


fnothaft:
Thanks @heuermh! Merged as https://github.com/bigdatagenomics/adam/commit/2e843de7d1c0a3185fd0e7a9a3c65eef07ae4ad3.

Issue #980, opened 2016-03-25
Adam2Fastq should output reverse complement when 0x10 flag is set for read


jpdna:
The expected output of bam2fastq conversion is to reverse complement the output sequence when the alignment record was marked as reverse complemented with the 0x10 flag.in the SAM/BAM file.  

Currently this does not happen, as  the same seq string as is found in the BAM file is output in fastq regardless of 0x10 flag.  Qualities need to be reversed as well.

Samtools follows this standard and recent discussion with Shop from Simons project brought this issue up as a common mistake in bam2fastq utilities.

arahuja:
@jpdna Looks like the seq and quality scores are reverse complemented/reversed under some conditions - are these not correct? 

https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/converters/AlignmentRecordConverter.scala#L78-L85

I did some comparisons with adam and picard's bam2fastq to see the same results, I think one of these tests is checked in

jpdna:
@arahuja, thanks - you are correct that in most cases the rev comp is happening as intended, but as described below I think the case of reads  with flags set as both reverse and unmapped is currently handled incorrectly in the code and tests.  

It happens that the first case I looked at was one where the reverse `0x10` flag was set and also the read was unmapped `0x4`.   Both samtools and picard currently reverse complement the fastq output in this case.   The [BAM spec](https://samtools.github.io/hts-specs/SAMv1.pdf) says:
```
• Bit 0x10 indicates whether SEQ has been reverse complemented and QUAL reversed. 
When bit 0x4 is unset, this corresponds to the strand to which the segment has been 
mapped.  When 0x4 is set,this indicates whether the unmapped read is stored in its
original orientation as it came off the sequencing machine
``` 
I interpret the above to to mean that seeing flags `0x10` && `0x4` indicates that when creating the BAM file the aligner chose to represent the SEQ string as reverse complemented despite the fact the read was unmapped, and thus to produce a correct pre-alignment fastq dump true to the original, we need to undo that by reverse complementing again when we make fastq dump.  

So, I think that samtools and Picard are correct and the current behavior in ADAM should be changed to match, and the fastq currently generated by ADAM is actually incorrect in these cases and not suitable to be used as input to a new alignment because they will appear to map to the wrong strand.
PR on the way.

fnothaft:
[...]

PR #982, opened 2016-03-26
Fix Adam2fastq in case of read with both reverse and unmapped flags

jpdna:
Fixes the bug in fastq dump discussed in  #980
specifically for case of a read  that has both reverse and unmapped flags set in the input bam file.

### DIFF
--- a/adam-core/src/test/scala/org/bdgenomics/adam/converters/AlignmentRecordConverterSuite.scala
+++ b/adam-core/src/test/scala/org/bdgenomics/adam/converters/AlignmentRecordConverterSuite.scala
@@ -219,16 +219,16 @@ class AlignmentRecordConverterSuite extends FunSuite {
     assert(secondRecordFastq(1) === "AATTCAAAACCAGCCTGGCCAATATGGTGAAACCTCATCTCTACTAAAAATACAAAAATTAGCCAGGCATGGTGGTGCGTGCGTGTAGTCCCAGCTACTT")
     assert(secondRecordFastq(2) === "+")
     assert(secondRecordFastq(3) === "?-DDBEEB=EEEDDEDEEEA:D?5?E?CEBE5ED?D:AEDEDEDED-B,BC0AC,BB6@CDBDEC?BCBAA@5,=8CA-?A>?2:&048<BB5BE#####")
-
   }
 
-  test("converting to fastq with unmapped reads when read reverse strand flag was set") {
-    //SRR062634.10448889      117     22      16079761        0       *       =       16079761        0       
-    // TTTCTTTCTTTTATATATATATACACACACACACACACACACACACATATATGTATATATACACGTATATGTATGTATATATGTATATATACACGTATAT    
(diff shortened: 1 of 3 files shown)


adam-core/src/test/scala/org/bdgenomics/adam/converters/AlignmentRecordConverterSuite.scala
```diff
@@ -222,7 +222,7 @@ class AlignmentRecordConverterSuite extends FunSuite {
 
   }
 
-  test("converting to fastq with unmapped reads") {
+  test("converting to fastq with unmapped reads when read reverse strand flag was set") {
```
heuermh: could you create a new unit test instead of modifying this existing one?  it is better to always be increasing unit test coverage

jpdna: The test as it is was previously written handled the case which it was then testing, which was the case of reverse stand flag with negative flag, incorrectly and thus failed with the changes introduced by this PR.   I modified the test to be correct, and the  change in the message above to `converting to fastq with unmapped reads when read reverse strand flag was set` just describes the  test more specifically.

fnothaft: I would agree with @heuermh; specifically, I would fix the current test, and add a new test that tests the "unmapped + reverse strand flag" case.

fnothaft:
Other than the test case nit, LGTM!

fnothaft:
Thanks for the test fix! Can you squash down and rebase? @heuermh if it looks good to you, I will merge.

heuermh:
+1

jpdna:
I updated PR so that tests cover correctly both umapped with Ox10 set and unset.  The existing test did need to be revised as it was incorrect before.    I had to add a the new test using different different records because actually the case which was not covered previously was the case of umapped without Ox10 set, and required a different record to test.  

I also revised to be consistent what was previously a confusing re-ordering of `(firstRecord,SecondRecord)` [here](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/test/scala/org/bdgenomics/adam/converters/AlignmentRecordConverterSuite.scala#L231)  when return tuple values are assigned in these tests.  As far as I can tell `first` and `second` here only refer to the order in which they appear in the input SAM file, and the assertion of
```
assert(firstRecord.getReadInFragment === 1)
assert(secondRecord.getReadInFragment === 0)
```
or
```
assert(firstRecord.getReadInFragment === 0)
assert(secondRecord.getReadInFragment === 1)
```
is more clear if the tuple order remains the same and the val names are not switched when parsing the tuple.

fnothaft:
Actually, I can take care of the rebase + squash when merging. I'll merge this now...

fnothaft:
Thanks @jpdna! Squashed and merged as https://github.com/bigdatagenomics/adam/commit/b8e36b27dab1e53a4fe80bcad754d7eff455223a.

PR #998, opened 2016-04-13
[ADAM-710] Add saveAs methods for feature formats GTF, BED, IntervalList, and NarrowPeak

### LINKED ISSUE #710: Support writing RDD[Feature] to various file formats
- [ ] GTF
- [ ] BED
- [ ] .interval_list (cf. #703)
- [ ] NarrowPeak

I'm imagining methods in `FeatureRDDFunctions` analogous to the {Sam,Bam,Adam,Fastq} methods in `AlignmentRecordRDDFunctions`.

heuermh:
Fixes #710.

Pushing this early for review.

Still todo is writing IntervalList header and ```-single``` file support, if that might be desirable.  Currently this uses Spark ```saveAsTextFile``` which writes to as many ```part-00...``` text files as there are partitions.


heuermh:
@tdanford Some specific review questions for you, if you have the time:

 - Is there a bug here? https://github.com/heuermh/adam/blob/feature-writers/adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala#L188
 - Can we add ```frame``` as a optional field to the schema?  Currently I'm pushing it in and out of attributes https://github.com/heuermh/adam/blob/feature-writers/adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureRDDFunctions.scala#L162
 - Why the UUID for ```featureId```?  https://github.com/heuermh/adam/blob/feature-writers/adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala#L226  It makes round-tripping not possible
 - Should we add Dbxref support to all the formats?  Currently only IntervalListParser parses them out https://github.com/heuermh/adam/blob/feature-writers/adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala#L226

heuermh:
@tdanford bump for feedback, see last comment

fnothaft:
@heuermh what `needs work` on this PR? Do you think we can get this in to 0.20.0 (end of month)?

heuermh:
The four items in the comment above that need review and single-file SAM/BAM-style header support for IntervalList similar to that described in #1009.

fnothaft:
OK. I'll get going on the review.

> single-file SAM/BAM-style header support for IntervalList similar to that described in #1009.

What does the `IntervalList` header look like? I'll work on #1009 and we can then slot this on top.

heuermh:
> What does the `IntervalList` header look like?

"The file format is relatively simple and reflects the SAM alignment format to a degree. A SAM style header must be present in the file which lists the sequence records against which the intervals are described."

https://samtools.github.io/htsjdk/javadoc/htsjdk/htsjdk/samtools/util/IntervalList.html

Thus we might be able to reuse the SAM header stuff directly; the current IntervalList [parsing code](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala#L133) does not.

fnothaft:
> Thus we might be able to reuse the SAM header stuff directly; the current IntervalList parsing code does not. You had mentioned refactoring the SAM/BAM header code although maybe not as part of #1009; I can pick this up based on what you do there.

I would prefer using the SAM header code to rolling our own.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
```diff
@@ -118,7 +122,11 @@ class GTFParser extends FeatureParser {
     _id.foreach(f.setFeatureId)
     _parentId.foreach(parentId => f.setParentIds(List[String](parentId)))
 
-    f.setAttributes(attrs)
+    if (frame != ".") {
+      f.setAttributes(collection.mutable.Map(attrs.toSeq: _*) += ("frame" -> frame))
```
fnothaft: `collection.mutable.Map(attrs.toSeq: _*) += ("frame" -> frame)` to `attrs.map(v -> ("frame", v)).toMap)`?

fnothaft: Also, could we get some docs here? What's this conditional for? It looks like we're special casing some formatting thing. Why's that? Etc.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
```diff
@@ -96,6 +96,10 @@ class GTFParser extends FeatureParser {
       .setFeatureType(feature)
       .setSource(source)
 
+    if (score != ".") {
```
fnothaft: Ditto here RE: adding docs for special case code.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
```diff
@@ -177,6 +185,7 @@ class IntervalListParser extends Serializable {
             Feature.newBuilder()
               .setContig(Contig.newBuilder().setContigName(fields(0)).build())
               .setStart(fields(1).toLong - 1)
+              // todo: IntervalList ranges are closed, is this correct?
```
fnothaft: This is correct. Drop `todo` but keep comment as documentation? E.g., `// NOTE: IntervalList ranges are closed`

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
```diff
@@ -196,8 +205,16 @@ class IntervalListParser extends Serializable {
 
 class BEDParser extends FeatureParser {
 
+  def isHeader(line: String): Boolean = {
+    return line.startsWith("#") || line.startsWith("browser") || line.startsWith("track")
```
fnothaft: Nit: Drop `return`

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
```diff
@@ -196,8 +205,16 @@ class IntervalListParser extends Serializable {
...
+  }
+
   override def parse(line: String): Seq[Feature] = {
 
+    if (isHeader(line)) {
+      return Seq()
```
fnothaft: Ditto here. I'd rather refactor the function so that we eliminate all `return` statements.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
```diff
@@ -230,22 +247,22 @@ class BEDParser extends FeatureParser {
       })
     }
     val attributes = new ArrayBuffer[(String, String)]()
-    if (fields.length > 6) {
+    if (fields.length > 6 && fields(6) != ".") {
```
fnothaft: It seems like there's a lot of repeated code here. Could we refactor? E.g.:

```
def attrExists(attrIdx: Int, attr: String) {
  if (fields.length > attrIdx && fields(attrIdx) != ".") {
    attributes += ("attr" -> fields(attrIdx)
  }
}

attrExists(6, "thickStart")
attrExists(7, "thickEnd")
...
```

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureRDDFunctions.scala
```diff
@@ -140,4 +141,96 @@ class FeatureRDDFunctions(featureRDD: RDD[Feature]) extends Serializable with Lo
...
+        case Strand.Forward     => "+"
+        case Strand.Reverse     => "-"
+        case Strand.Independent => ""
+      }
+      if (!feature.getAttributes.containsKey("thickStart")) {
+        // write BED6 format
+        Joiner.on('\t').join(chrom, start, end, name, score, strand)
+      } else {
+        // write BED12 format
+        val thickStart = feature.getAttributes.getOrElse("thickStart", ".")
```
fnothaft: I would factor this out --> `feature.getAttributes.getOrElse(attr, ".")` to:

```
def getAttr(attr: String): String = {
  feature.getAttributes.getOrElse(attr, ".")
}

val thickStart = getAttr("thickStart")
...
```

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureRDDFunctions.scala
```diff
@@ -140,4 +141,96 @@ class FeatureRDDFunctions(featureRDD: RDD[Feature]) extends Serializable with Lo
...
+      val start = feature.getStart
+      val end = feature.getEnd
+      val name = Option(feature.getFeatureType).getOrElse(".")
+      val score = Option(feature.getValue).getOrElse(".")
+      val strand = feature.getStrand match {
+        case Strand.Forward     => "+"
+        case Strand.Reverse     => "-"
+        case Strand.Independent => "."
+      }
+      val signalValue = feature.getAttributes.getOrElse("signalValue", "0")
```
fnothaft: Might factor these out too.

adam-core/src/test/scala/org/bdgenomics/adam/rdd/features/FeatureRDDFunctionsSuite.scala
```diff
@@ -0,0 +1,256 @@
...
+
+import java.io.File
+import org.apache.spark.rdd.RDD
+import org.bdgenomics.adam.rdd.ADAMContext._
+import org.bdgenomics.adam.rdd.features.FeatureRDDFunctionsSuiteHelper._
+import org.bdgenomics.adam.util.ADAMFunSuite
+import org.bdgenomics.formats.avro._
+import org.scalactic.{ Equivalence, TypeCheckedTripleEquals }
+
+object FeatureRDDFunctionsSuiteHelper {
```
fnothaft: The functions in this object seem like they'd be useful to surface outside of the `test` scope.

adam-core/src/test/scala/org/bdgenomics/adam/rdd/features/FeatureRDDFunctionsSuite.scala
```diff
@@ -0,0 +1,256 @@
...
+        f.getContig.getContigName,
+        f.getStart,
+        f.getEnd
+      )
+    features.coalesce(1).sortBy(reference)
+  }
+}
+
+class FeatureRDDFunctionsSuiteHelper extends Serializable {
+  // empty
```
fnothaft: What's this here for?

adam-core/src/test/scala/org/bdgenomics/adam/rdd/features/FeatureRDDFunctionsSuite.scala
```diff
@@ -0,0 +1,256 @@
...
+
+  sparkTest("save IntervalList as NarrowPeak format") {
+    val inputPath = resourcePath("features/SeqCap_EZ_Exome_v3.hg19.interval_list")
+    val features = sc.loadIntervalList(inputPath)
+    val outputPath = tempLocation(".narrowPeak")
+    features.saveAsNarrowPeak(outputPath)
+  }
+
+  /*
+  sparkTest("round trip IntervalList format") {
```
fnothaft: Eliminate dead test code, or change to an ignored test --> `ignore("round trip..."){`

fnothaft:
LGTM! I just made a pass through with a variety of nits.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
heuermh: Yeah I don't like it, I will propose adding frame as an attribute to Feature in an upcoming pull request.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
heuermh: What, you want my Scala not to look like Java?  :)

fnothaft: ;)

adam-core/src/test/scala/org/bdgenomics/adam/rdd/features/FeatureRDDFunctionsSuite.scala
heuermh: Adds the Serializable marker, otherwise tests will error

fnothaft: You can just add that to the singleton object, no? E.g., `object FeatureRDDFunctionsSuiteHelper extends Serializable`

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
heuermh: If IntervalList intervals are 1-based and closed, shouldn't the next line be
```scala
.setEnd(fields(2).toLong)
```
E.g. `[7-9]` in IntervalList should be `[6-9)` in our 0-based, closed open system.

fnothaft: Agreed, my `correct` was RE: the comment.

fnothaft:
@heuermh when do you think we can wrap this by?

heuermh:
Updated with GFF3 support and addressed some comments above.

Still left to do:
 * Continue refactoring `FeatureParser`to reuse functionality where appropriate and improve consistency between formats
 * Refactor `toGenes` to use new proposed feature fields (`geneId`, `transcriptId`, and `exonId`) instead of attributes
 * `-single` file support, if that might be desirable
 * Writing header for IntervalList and GFF3 formats

heuermh:
Ready for another round of review.  I'd be ok pushing this and adding `-single` and header support later.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
```diff
@@ -44,86 +46,120 @@ class FeatureFile(parser: FeatureParser) extends Serializable {
...
-      case _   => Strand.Independent
+    // assign values for various feature fields from attributes
+    Features.assignAttributes(GTFParser.parseAttributes(attributes), f)
+
+    Seq(f.build())
+  }
+}
+
+object GFF3Parser {
+  def parseAttributes(attributes: String): Seq[(String, String)] =
```
fnothaft: Add docs? This is a weird line. I don't love the pattern match on the array, but that's just me. I would probably do something like:

```
attributes.split(";")
  .map(s => {
    val eqIdx = s.indexOf("=")
    (s.take(eqIdx), s.drop(eqIdx + 1))
  }).toSeq
```

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
```diff
@@ -44,86 +46,120 @@ class FeatureFile(parser: FeatureParser) extends Serializable {
...
+    if (line.startsWith("#")) {
+      return Seq()
     }
-    f.setStrand(_strand)
 
-    val exonId: Option[String] = attrs.get("exon_id").orElse {
-      attrs.get("transcript_id").flatMap(t => attrs.get("exon_number").map(e => t + "_" + e))
+    val fields = line.split("\t")
+    // skip empty or invalid lines
+    if (fields.length < 9) {
```
fnothaft: Should log invalid lines.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
```diff
@@ -44,86 +46,120 @@ class FeatureFile(parser: FeatureParser) extends Serializable {
...
+    val f = Feature.newBuilder()
+      .setSource(source)
+      .setFeatureType(featureType)
+      .setContigName(seqid)
+      .setStart(start.toLong - 1L) // GFF3 coordinate system is 1-based
+      .setEnd(end.toLong) // GFF3 ranges are closed
+
+    // set score if specified
+    if (score != ".") {
+      f.setScore(source.toDouble)
```
fnothaft: s/source/score/g?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
fnothaft: Ping.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
```diff
@@ -194,58 +230,65 @@ class IntervalListParser extends Serializable {
...
     fb.setStart(fields(1).toLong)
     fb.setEnd(fields(2).toLong)
 
     if (fields.length > 3) {
-      fb.setFeatureType(fields(3))
+      fb.setName(fields(3))
     }
     if (fields.length > 4) {
-      fb.setValue(fields(4) match {
+      fb.setScore(fields(4) match {
```
fnothaft: I don't love this pattern match. I would just:

```
if (fields(4) != ".") {
  fb.setScore(fields(4).toDouble)
}
```

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
```diff
@@ -194,58 +230,65 @@ class IntervalListParser extends Serializable {
...
-      fb.setStrand(fields(5) match {
-        case "+" => Strand.Forward
-        case "-" => Strand.Reverse
-        case _   => Strand.Independent
-      })
+      Features.toStrand(fields(5)).foreach(fb.setStrand(_))
     }
     val attributes = new ArrayBuffer[(String, String)]()
-    if (fields.length > 6) {
+    if (fields.length > 6 && fields(6) != ".") {
```
fnothaft: Factor out `fields.length > n && fields(n) != "."` into:

```
def validIdx(n: Int): Boolean = {
  fields.length > n && fields(n) != "."
}
```

Or something like that.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureRDDFunctions.scala
```diff
@@ -133,11 +132,167 @@ class FeatureRDDFunctions(featureRDD: RDD[Feature]) extends Serializable with Lo
...
+      val featureId: Option[String] = Option(f.getFeatureId)
+      val geneId: Option[String] = Option(f.getGeneId)
+      val transcriptId: Option[String] = Option(f.getTranscriptId)
+      val exonNumber: Option[String] = Option(f.getAttributes.get("exon_number"))
+      val exonId: Option[String] = Option(f.getExonId) match {
+        case Some(e) => Option(e)
+        case None    => transcriptId.flatMap(t => exonNumber.map(e => t + "_" + e))
+      }
+
+      val (_featureId, _parentId) =
```
fnothaft: We don't use `_` prefixes elsewhere in this codebase, so I'd prefer not to introduce them here. I would suffix `featureId: Option[String]` to `featureIdOpt`, which is fairly common across Scala projects.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureRDDFunctions.scala
```diff
@@ -133,11 +132,167 @@ class FeatureRDDFunctions(featureRDD: RDD[Feature]) extends Serializable with Lo
...
+        val blockSizes = feature.getAttributes.getOrElse("blockSizes", ".")
+        val blockStarts = feature.getAttributes.getOrElse("blockStarts", ".")
+        List(chrom, start, end, name, score, strand, thickStart, thickEnd, itemRgb, blockCount, blockSizes, blockStarts).mkString("\t")
+      }
+    }
+    featureRDD.map(toBed).saveAsTextFile(fileName)
+  }
+
+  def saveAsIntervalList(fileName: String) = {
+    def escape(entry: (Any, Any)): String = {
```
fnothaft: Where does `escape` get used? Perhaps I'm just missing things but I don't see it get called anywhere.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureRDDFunctions.scala
```diff
@@ -133,11 +132,167 @@ class FeatureRDDFunctions(featureRDD: RDD[Feature]) extends Serializable with Lo
...
+
+  def sortByExon(): RDD[Feature] = {
+    def exon(f: Feature) =
+      (
+        f.getContigName,
+        f.getStart,
+        f.getEnd,
+        f.getFeatureType,
+        // try exon_id first
+        Option(f.getExonId).getOrElse("~"),
```
fnothaft: Is there any significance to `~`? I would document.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureRDDFunctions.scala
```diff
@@ -133,11 +132,167 @@ class FeatureRDDFunctions(featureRDD: RDD[Feature]) extends Serializable with Lo
...
+        // then try transcript_id and exon_number
+        Option(f.getTranscriptId).getOrElse("~"),
+        f.getAttributes.getOrElse("exon_number", Int.MaxValue.toString).toInt
```
fnothaft: Prefer `f.getAttributes.get("exon_number").map(_.toInt).getOrElse(Int.MaxValue)`

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/Features.scala
```diff
@@ -0,0 +1,238 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.bdgenomics.adam.rdd.features
+
+import scala.collection.JavaConversions._
+
```
fnothaft: Nit: whitespace.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/Features.scala
```diff
@@ -0,0 +1,238 @@
...
+import scala.collection.mutable.ArrayBuffer
+import scala.collection.mutable.HashMap
+import scala.collection.mutable.MutableList
+
+import org.bdgenomics.formats.avro.{ Dbxref, Feature, OntologyTerm, Strand }
```
fnothaft: Nit: import sort order --> `org` before `scala`.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/Features.scala
```diff
@@ -0,0 +1,238 @@
...
+    }
+  }
+
+  /**
+   * Convert the specified strand to its string value.
+   *
+   * @param strand strand to convert
+   * @return the specified strand converted to its string value
+   */
+  def toString(strand: Strand): String = {
```
fnothaft: I would avoid `toString` as a function name for Singleton objects.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/Features.scala
```diff
@@ -0,0 +1,238 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.bdgenomics.adam.rdd.features
+
+import scala.collection.JavaConversions._
+
+import scala.collection.mutable.ArrayBuffer
```
fnothaft: Prefer `import scala.collection.mutable.{...}` to imports from same package across multiple lines.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/Features.scala
```diff
@@ -0,0 +1,238 @@
...
+    }
+  }
+
+  /**
+   * Convert the specified database cross reference to its string value.
+   *
+   * @param dbxref database cross reference to convert
+   * @return the specified database cross reference converted to its string value
+   */
+  def toString(dbxref: Dbxref): String = {
```
fnothaft: Ditto RE naming of `toString`.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/Features.scala
```diff
@@ -0,0 +1,238 @@
...
+    }
+  }
+
+  /**
+   * Convert the specified ontology term to its string value.
+   *
+   * @param ontologyTerm ontology term to convert
+   * @return the specified ontology term converted to its string value
+   */
+  def toString(ontologyTerm: OntologyTerm): String = {
```
fnothaft: Ditto

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/Features.scala
```diff
@@ -0,0 +1,238 @@
...
+    }
+    if (!dbxrefs.isEmpty) {
+      f.setDbxrefs(dbxrefs)
+    }
+    if (!ontologyTerms.isEmpty) {
+      f.setOntologyTerms(ontologyTerms)
+    }
+
+    // set remaining attributes if non-empty;
+    // any duplicate keys are lost at this point, last one in wins
```
fnothaft: Would it make sense to log duplicate keys?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/Features.scala
```diff
@@ -0,0 +1,238 @@
...
+   */
+  def gatherAttributes(feature: Feature): Seq[(String, String)] = {
+    val attrs = new ArrayBuffer[(String, String)]
+
+    // not sure why scala compiler requires this
+    def addBooleanTuple(b: java.lang.Boolean) = {
+      attrs += Tuple2("Is_circular", b.toString)
+    }
+
+    // ... and why Tuple2 is required here
```
fnothaft: OOC, does it compile OK if you don't use an anon argument? E.g., does `Option(feature.getFeatureId).foreach(id => attrs += ("ID", id))` compile?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/Features.scala
```diff
@@ -0,0 +1,238 @@
...
+    Option(feature.getFeatureId).foreach(attrs += Tuple2("ID", _))
+    Option(feature.getName).foreach(attrs += Tuple2("Name", _))
+    Option(feature.getTarget).foreach(attrs += Tuple2("Target", _))
+    Option(feature.getGap).foreach(attrs += Tuple2("Gap", _))
+    Option(feature.getDerivesFrom).foreach(attrs += Tuple2("Derives_from", _))
+    Option(feature.getIsCircular).foreach(addBooleanTuple)
+    Option(feature.getGeneId).foreach(attrs += Tuple2("gene_id", _))
+    Option(feature.getTranscriptId).foreach(attrs += Tuple2("transcript_id", _))
+    Option(feature.getExonId).foreach(attrs += Tuple2("exon_id", _))
+    for (alias <- feature.getAliases) attrs += Tuple2("Alias", alias)
```
fnothaft: I generally prefer `feature.getAliases.foreach` to `for (...)`, but that's just me.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/Features.scala
```diff
@@ -0,0 +1,238 @@
...
+
+  /**
+   * Build a name for the specified feature considering its attributes.  Used when
+   * writing out to lossy feature formats such as BED, NarrowPeak, and IntervalList.
+   *
+   * @param feature feature
+   * @return a name for the specified feature considering its attributes
+   */
+  def nameOf(feature: Feature): String = {
+    Option(feature.getName).foreach(return _)
```
fnothaft: I strongly dislike explicit returns in Scala code. I would rather:

```
def nameOf(feature: Feature): String = {
  Option(feature.getName).getOrElse(
    Option(feature.getFeatureId).getOrElse(feature.getFeatureType match {
      case "exon" => Option(feature.getExonId)
      case "transcript" => Option(feature.getTranscriptId)
      case "gene" => Option(feature.getGeneId)
      case s: String => Some(s)
      case _ => None
    }.getOrElse("feature")))
}
```

fnothaft:
Just made a pass. Overall, LGTM! Dropped various style nits and the like inline.

> I'd be ok pushing this and adding -single and header support later.

I +1.

Also, this adds all the code needed for handling the changes in https://github.com/bigdatagenomics/bdg-formats/pull/82, correct?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/Features.scala
heuermh: +1, will add logging

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
heuermh: the unit tests should've caught this typo!

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureRDDFunctions.scala
heuermh: do nulls sort first or last in scala?  I'm used to e.g. Ordering.nullsFirst in Guava.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/Features.scala
heuermh: I don't think that worked before, maybe for the same anon argument issue you mention above.  Will give it a try.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
heuermh: here I find it better to short circuit return than to put the whole method in an if block

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureRDDFunctions.scala
heuermh: nulls don't work in scala Ordering, and this line doesn't compile or handle nulls

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/Features.scala
heuermh: this and various permutations doesn't work for me; compile errors or `value getOrElse is not a member of java.io.Serializable`

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureRDDFunctions.scala
fnothaft: Ah, sorry I mentally had the return type of `f.getAttributes` as a Scala Map. `f.getAttributes` returns a `java.util.String`, right? If you convert this return value into a Scala Map (e.g., via [`scala.collection.JavaConversions`](http://www.scala-lang.org/api/2.10.4/index.html#scala.collection.JavaConversions$) or etc), `.get("exon_number")` returns an `Option[String]`, and then the `.getOrElse` handles the `null` case, no?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureRDDFunctions.scala
fnothaft: Not sure. I was figuring that this was to handle the null case. If so, can you add a line comment to indicate so?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/Features.scala
fnothaft: You'd need to convert the `java.util.List` into a Scala collection for it to work.

heuermh:
Addressed most review comments and rebased.

fnothaft:
Jenkins, test this please.

heuermh:
The unit test failure is due to indeterminate sorting. Is there a way to easily sort by deep compare (nested arrays and maps)?

heuermh:
Maybe that's not the problem.  I don't see this error locally.

heuermh:
Jenkins, test this please.

fnothaft:
[bdg-formats 0.9.0 has pushed](http://search.maven.org/#artifactdetails%7Corg.bdgenomics.bdg-formats%7Cbdg-formats%7C0.9.0%7Cjar).

fnothaft:
Jenkins, retest this please.

fnothaft:
Thanks @heuermh! I've squashed this down to a single commit and merged as https://github.com/bigdatagenomics/adam/commit/8abb47ad2447bbc0376f9b5052e2e029bdf6b604.

heuermh:
Thanks!

Issue #1011, opened 2016-04-24
Factor out *RDDFunctions classes


fnothaft:
We have many *RDDFunctions classes in org.bdgenomics.adam.rdd that should be factored out. I think that if we did this, we could eliminate most of the implicit conversions in `org.bdgenomics.adam.rdd.ADAMContext` too.

heuermh:
Please elaborate a bit more.  I had some further clean up and moving stuff around in mind.

ADAMContext is a nasty bugger in terms of cyclic package dependencies and extensibility.  Perhaps it needs to be that way so that users only have to ```import ADAMContext._``` though.

fnothaft:
I just opened #1015, which I think will shade more light on my thoughts. Specifically, I'd like to eliminate all the `*RDDFunctions` classes and their implicits by migrating the code into the classes that extend `GenomicRDD`s.

fnothaft:
Punting to 0.21.0.

heuermh:
Not sure I agree.  I don't see the utility in cutting a release halfway through a major refactoring.

fnothaft:
That's a fair point. Then maybe we push 0.20.0 out to end of June?

heuermh:
Yeah, let's try to do a round of triage and estimation while we're all together in person.

fnothaft:
This will have big conflicts with #1081 and #964. I would like to merge this in first, since it is the vastly more complex PR, and to then rebase the other two on top.

PR #1015, opened 2016-04-24
[ADAM-909] Refactoring variation RDDs.

### LINKED ISSUE #909: Load/store sequence dictionaries alongside Genotype RDDs
From discussion with @akmorrow13 and @erictu. For certain queries against an `RDD[Genotype]`, we wind up recreating the `SequenceDictionary` via the `ADAMSequenceDictionaryRDDAggregator`. This works, but is slow. We can take a similar approach to #906 to eliminate this problem.

fnothaft:
Resolves #909:

* Refactors `org.bdgenomics.adam.rdd.variation` to add `GenomicRDD`s for `Genotype`, `Variant`, and `VariantContext`. These classes write sequence and sample metadata to disk.
* Refactors `ADAMRDDFunctions` to an abstract class in preparation for further refactoring in #1011.
* Added `AvroGenomicRDD` trait which consolidates Parquet + Avro metadata writing code across all Avro data models.


fnothaft:
Jenkins, retest this please.

adam-core/src/test/scala/org/bdgenomics/adam/rdd/ADAMContextSuite.scala
```diff
@@ -270,7 +270,7 @@ class ADAMContextSuite extends ADAMFunSuite {
   sparkTest("can read a small .vcf file") {
     val path = resourcePath("small.vcf")
 
-    val vcs = sc.loadGenotypes(path).toVariantContext.collect.sortBy(_.position)
+    val vcs = sc.loadGenotypes(path).toVariantContextRDD.rdd.collect.sortBy(_.position)
```
heuermh: Since we have this
```scala
object ADAMContext {
  implicit def genomicRDDToRDD[T](gRdd: GenomicRDD[T]): RDD[T] = gRdd.rdd
```
https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala#L114

why is the ```.rdd``` here and elsewhere necessary?  Implicits work fine except for when they don't?

fnothaft: Oh man. I zoned on the implicit. That's my bad. I'll fix this.

heuermh: Sorry, what I mean was that sometimes the ```genomicRDDToRDD``` implicit works for me and sometimes it doesn't.  I assume because you are using ```.rdd``` here that they are necessary.  The scala compiler and I are still getting to know each other.

fnothaft: Nah I was using them because I forgot that we had the implicit.

fnothaft:
Just pushed an update with upgrades to bdg-formats 0.8.0 and cleanup of the `.rdd` bits. This fails unit tests right now, which I'll patch up this AM. CC @erictu @heuermh 

The 0.8.0 upgrade is split out into a separate commit. This will need to be squashed down before merge.

fnothaft:
Just pushed another commit to address the failing unit tests. Please review @heuermh @erictu.

@heuermh you may be interested to find that GZIPed VCF loading doesn't work. Alas, the VCFHeaderReader code from Hadoop-BAM seems to fail when trying to load GZIPed VCF. I think there's a fall-through case that wasn't considered properly. I'm going to open an issue here to test more, and then if I can isolate/test/resolve it, I'll go upstream to Hadoop-BAM.

heuermh:
> you may be interested to find that GZIPed VCF loading doesn't work.

It was working for me, and there is a unit test to verify it.
https://github.com/bigdatagenomics/adam/blame/master/adam-core/src/test/scala/org/bdgenomics/adam/rdd/ADAMContextSuite.scala#L284

fnothaft:
>> you may be interested to find that GZIPed VCF loading doesn't work.
> It was working for me, and there is a unit test to verify it.
> https://github.com/bigdatagenomics/adam/blame/master/adam-core/src/test/scala/org/bdgenomics/adam/rdd/ADAMContextSuite.scala#L284

Oh, yeah; didn't mean to imply that it was currently misimplemented. The issue comes up because we now rely on reading the VCF header separately to pull out sequence/sample metadata:

https://github.com/bigdatagenomics/adam/pull/1015/files#diff-d36ea7d0742decd0b040a73a96af06e9R136

The `VCFHeaderReader` in Hadoop-BAM appears to mishandle GZIP/BGZIPed VCFs. As far as I can tell, it falls through and tries to treat them as BCF files.

heuermh:
Yep, in agreement.  There's some duplicated code in Hadoop-BAM that should be shared, and things probably diverged between the VCF input format and VCF header reader in this case.

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Vcf2ADAM.scala
```diff
@@ -65,22 +65,24 @@ class Vcf2ADAM(val args: Vcf2ADAMArgs) extends BDGSparkCommand[Vcf2ADAMArgs] wit
...
       log.info("Using contig translation")
 
-    var adamVariants: RDD[VariantContext] = sc.loadVcf(args.vcfPath, sd = dictionary)
-    if (args.coalesce > 0) {
-      if (args.coalesce > adamVariants.partitions.size || args.forceShuffle) {
-        adamVariants = adamVariants.coalesce(args.coalesce, shuffle = true)
+    val variantContextRdd = sc.loadVcf(args.vcfPath, sdOpt = dictionary)
+    var variantContextsToSave = if (args.coalesce > 0) {
+      if (args.coalesce > variantContextRdd.partitions.size || args.forceShuffle) {
+        variantContextRdd.transform(_.coalesce(args.coalesce, shuffle = true))
```
heuermh: could you explain what the `.transform(_.` part is doing here?  is it not possible to call `variantContextRdd.coalesce` directly?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMRDDFunctions.scala
```diff
@@ -81,7 +131,25 @@ class ADAMRDDFunctions[T <% IndexedRecord: Manifest](rdd: RDD[T]) extends Serial
       ContextUtil.getConfiguration(job)
     )
   }
+}
+
+@deprecated("Extend ADAMRDDFunctions and mix in GenomicRDD wherever possible in new development.",
+  since = "0.20.0")
+class ConcreteADAMRDDFunctions[T <% IndexedRecord: Manifest] private[rdd] (val rdd: RDD[T]) extends ADAMRDDFunctions[T] {
```
heuermh: what is this used for?  if just for the implicit conversion could it be private to `ADAMContext`?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -576,32 +594,67 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
       classOf[VCFInputFormat], classOf[LongWritable], classOf[VariantContextWritable],
       ContextUtil.getConfiguration(job)
     )
+
+    // attach instrumentation
     if (Metrics.isRecording) records.instrument() else records
 
-    records.flatMap(p => vcc.convert(p._2.get))
+    // load vcf metadata
+    val (vcfSd, samples) = loadVcfMetadata(filePath)
```
heuermh: the header has already been read via `VCFInputFormat` above, and `VCFRecordReader` holds a reference to it.  Is there some way to pull that out?  `VCFHeaderReader` doesn't contain the decompression bits that `VCFRecordReader` does.

adam-core/src/test/scala/org/bdgenomics/adam/rdd/ADAMContextSuite.scala
```diff
@@ -281,13 +281,13 @@ class ADAMContextSuite extends ADAMFunSuite {
...
 
-  sparkTest("can read a gzipped .vcf file") {
+  ignore("can read a gzipped .vcf file") {
     val path = resourcePath("test.vcf.gz")
     val vcs = sc.loadVcf(path, None)
     assert(vcs.count === 6)
   }
 
-  sparkTest("can read a BGZF gzipped .vcf file") {
+  ignore("can read a BGZF gzipped .vcf file") {
```
heuermh: bummer that a feature just added for users is gone again.  Is there any way to continue to support this without waiting for another release of Hadoop-BAM?

heuermh:
Created https://github.com/HadoopGenomics/Hadoop-BAM/pull/96 to document missing support for gzipped and BGZF VCF formats.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/variation/VariationRDDFunctions.scala
```diff
@@ -50,103 +48,13 @@ class VariantContextRDDFunctions(rdd: RDD[VariantContext]) extends ADAMSequenceD
...
 }
 
 class GenotypeRDDFunctions(rdd: RDD[Genotype]) extends Serializable with Logging {
-  def toVariantContext(): RDD[VariantContext] = {
-    rdd.keyBy({ g => RichVariant.variantToRichVariant(g.getVariant) })
-      .groupByKey
-      .map { case (v: RichVariant, g) => new VariantContext(ReferencePosition(v), v, g, None) }
-  }
 
   def filterByOverlappingRegion(query: ReferenceRegion): RDD[Genotype] = {
```
erictu: Can we get this operation supported over the GenotypeRDD class?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -576,32 +594,67 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
   def loadParquetGenotypes(
     filePath: String,
     predicate: Option[FilterPredicate] = None,
-    projection: Option[Schema] = None): RDD[Genotype] = {
-    loadParquet[Genotype](filePath, predicate, projection)
+    projection: Option[Schema] = None): GenotypeRDD = {
+    val rdd = loadParquet[Genotype](filePath, predicate, projection)
+
+    // load sequence info
+    val sd = loadAvroSequences(filePath)
```
erictu: This means that providing the referenceDictionary upon vcf2adam conversion is required. I tried converting a vcf without providing the optional -dict flag, and I'm unable to load it in. Should we require the dict to be provided when converting from vcf to adam?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -576,32 +594,67 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+
+    // load avro record group dictionary and convert to samples
+    val rgd = loadAvroSampleMetadata(filePath)
```
erictu: This means that providing the rgdict upon vcf2adam conversion is required. I tried converting a vcf without providing the optional -dict flag, and I'm unable to load it in. Should we require the dict to be provided when converting from vcf to adam?

heuermh: No, we should generate it from the VCF header

jpdna:
When I run `mvn clean package` on the genotype-rdd branch I get the following error
```
[INFO] 'compiler-interface' not yet compiled for Scala 2.9.3. Compiling...
error: error while loading CharSequence, class file '/usr/java/jdk1.8.0_71/jre/lib/rt.jar(java/lang/CharSequence.class)' is broken
(bad constant pool tag 15 at byte 1501)
error: error while loading AnnotatedElement, class file '/usr/java/jdk1.8.0_71/jre/lib/rt.jar(java/lang/reflect/AnnotatedElement.class)' is broken
(bad constant pool tag 15 at byte 2713)
error: error while loading Arrays, class file '/usr/java/jdk1.8.0_71/jre/lib/rt.jar(java/util/Arrays.class)' is broken
(bad constant pool tag 15 at byte 12821)
error: error while loading Comparator, class file '/usr/java/jdk1.8.0_71/jre/lib/rt.jar(java/util/Comparator.class)' is broken
(bad constant pool tag 15 at byte 5003)
/tmp/sbt_d7ce1bc8/API.scala:384: error: java.util.Comparator does not take type parameters
	private[this] val sortClasses = new Comparator[Symbol] {

```
Interestingly, in exact same environment I can compile the current ADAM repo master just fine, so the breaking change is in this brnach.
Is my use of jdk1.8 here or scala 2.1-.4 not supported?

jpdna:
Ignore previous comment - I didn't switch branches to the current one.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
jpdna: > No, we should generate it from the VCF header

I'm going to try to add this functionality and make a PR against this PR

fnothaft: I believe this functionality already exists?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
fnothaft: Actually, a _really convoluted_ way to do this is to read the VCF via the input format and to then pull the header from a record.

I don't like this for various reasons, but I think this is a good WAR for the VCF.GZ/BCF issue. And when I say "good", I mean "passible".

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
jpdna: @erictu - using this PR as is I was able to run
`adam-submit vcf2adam small.vcf out.adam`
and it appears to have worked fine without the `-dict` and produced `_seqdict.avro` by extracting from the VCF header as we would expect.
Can you describe more about how to reproduce the error you saw when -dict is ommited?  
Did your VCF lack a `##contig=` in the header ?

erictu: The issue is with not producing an `_rgdict.avro`. 
I used `loadGenotypes` in the adam shell and it threw an error because the `_rgdict.avro` wasn't there.

jpdna: I see the same  think @erictu  Indeed the error is like
```
scala> val g1 = ac.loadGenotypes("/test_vcf/out5.adam")

java.io.FileNotFoundException: File /test_vcf/out5.adam/_rgdict.avro does not exist
```
I believe the cause is that  vcf2adam  produces `_samples.avro` but `loadAvroSampleMetadata` called by `loadParquetGenotypes`  looks for `_rgdict.avro`.  I don't think this issue is related to the `-dict` to vcf2adam flag - which deals anyway with `seqdict`.    Logic needs to be added to choose the correct sample/record metadata file name when loading genotypes from Parquet.  I'll test out a solution.

jpdna: @fnothaft - please have a look at https://github.com/fnothaft/adam/pull/9 to see if you want to include in this PR to solve this `_rgdict.avro` `_samples.avro` issue

fnothaft: Awesome! Thanks for the code. I get the problem now and will review and merge fnothaft#9 in to address.

fnothaft:
Just pushed https://github.com/bigdatagenomics/adam/pull/1015/commits/0f9030230148f617c71f6c93f4a44f019c39d172 to address the VCF.GZ issues. I will merge @jpdna's PR as well. @heuermh I'm still working on the BCF issue. There's something more going on here. I will ping back with more of an error trace.

heuermh:
> I'm still working on the BCF issue.

Which BCF issue?  BCF files probably won't work in ADAM due to issue(s) upstream in HTSJDK, where they don't follow the latest BCF specification.

fnothaft:
>> I'm still working on the BCF issue.
>
> Which BCF issue? BCF files probably won't work in ADAM due to issue(s) upstream in HTSJDK, where they don't follow the latest BCF specification.

Ah, OK! That is the exact issue I was running into. I won't hassle myself trying to fix that then.

(I tried regenerating BCF using Picard, which amusingly enough, doesn't seem to work)

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
jpdna: > Are samples in a VCF really the same as record groups in a SAM/BAM file?

Good point @heuermh - I also had this concern about conflating `SAM/BAM RecordGroup` metadata with concept of a `VCF sample` - I was just focused so far on fixing the bug in code as written.    I need to study the schema as a whole a bit more before stating an opinion - but if there is even a marginal distinction between `SAM/BAM (Record/Read Group)` and `VCF sample` I'd suggest erring on the side of breaking the concepts apart for clarity and adding a new "SampleMetadata" avro schema and object to old VCF sample dict metadata.

fnothaft: > I need to study the schema as a whole a bit more before stating an opinion - but if there is even a marginal distinction between SAM/BAM (Record/Read Group) and VCF sample I'd suggest erring on the side of breaking the concepts apart for clarity and adding a new "GenotypedSampleMetadata" avro schema and object to hold VCF sample dict metadata.

I think that sounds reasonable, but maybe we can open that as future work? My general reasoning for starting with a record group is that a VCF sample descends from one or more record groups. I would like to punt this to after this PR for a few reasons:

* This PR resolves a blocking issue for a downstream tool.
* Since we present just a `Seq[String]` with the sample metadata and the actual way that the sample is saved to disk is never visible to the end user, I don't think it is a critical issue.
* We should do a broader refactoring of the various metadata bits --> factor out computational provenance from wetlab provenance, allow VCF sample to descend from read group samples, etc.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
fnothaft: Resolved by https://github.com/bigdatagenomics/adam/pull/1015/commits/0f9030230148f617c71f6c93f4a44f019c39d172.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMRDDFunctions.scala
fnothaft: This is necessary for the `flatten` command line tool, IIRC.

fnothaft: But yes, I think making it private is a good idea.

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Vcf2ADAM.scala
fnothaft: `.transform(_.)` transforms the RDD that underlies the `VariantContextRDD` and emits a new `VariantContextRDD`.

fnothaft: If you did `variantContextRdd.coalesce` that would emit a `RDD[VariantContext]` and you'd lose the `.saveAsVcf`/etc functions, which we use later.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
jpdna: > to punt this to after this PR

+1  - Just created https://github.com/bigdatagenomics/adam/issues/1039 to hold that thought for future

heuermh: +1 to later formats changes, -0 on adding any new records with Metadata in the name.

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Vcf2ADAM.scala
heuermh: I'm liking this pattern less as time goes on.  Perhaps we should drop RDD from the VariantContextRDD class names for something more generic and remove the implicit conversion.  Or at least remove the implicit conversion.  What the user gets back from an ADAMContext load method should `has_a` RDD and not pretend to `is_a` RDD.

jpdna: > Perhaps we should drop RDD from the VariantContextRDD

I agree there may be some nomenclature concern that our many *RDD container objects for rdd+dicts are not RDDs in same true sense as IntervalRDD and IndexedRDD which actually extend RDD.  If I am reading correctly, there is not an implicit conversion that would even allow these to be treated as an RDD directly.   The meaning seems more  like `Holder ofRDDofVariantContextWithDicts` though I don't advocate that name.  What do you think @fnothaft   ?  At this point though we have already gone down this path a ways using the suffix "RDD" in a broad sense, so a renaming should perhaps come in the future after this PR.

heuermh: There's this implicit conversion from GenotypeRDD (which VariantContextRDD and others extend from) to RDD
https://github.com/fnothaft/adam/blob/822c64e3530d7ed8bb86daeb7b900bd7bf2f7854/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala#L116

heuermh: Created a new issue https://github.com/bigdatagenomics/adam/issues/1040, I'm ok with this for now.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/variation/VariationRDDFunctions.scala
fnothaft: Added.

fnothaft:
Just pushed a commit that (I think) addresses the remaining review comments. Can I get a final pass on this? I will then squash this down to two commits.

heuermh:
+1, thanks!

jpdna:
+1 - I have no outstanding concerns

erictu:
+1, seems good to me!

fnothaft:
Squashed down into three commits and cleaned up the history.

fnothaft:
Hold on the merge; I see a unit test failure in one of the intermediate commits when building locally.

fnothaft:
Fixed and pushed.

fnothaft:
Jenkins, retest this please.

heuermh:
Do you think our logging related changes might have anything to do with that failure?  We removed some of the logging configuration initialization bits in #1028

erictu:
Can we add in the corresponding changes to `org.bdgenomics.adam.projections.GenotypeField`?

heuermh:
@fnothaft do you want to push another commit for `GenotypeField` and then re-squash?  I could pr against this one but it probably isn't worth it for such a small change.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -132,13 +133,38 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+    try {
+      val vcfHeader = VCFHeaderReader.readHeaderFrom(WrapSeekable.openPath(sc.hadoopConfiguration,
+        new Path(filePath)))
+      headerToMetadata(vcfHeader)
+    } catch {
+      case e: Throwable => {
+
+        // due to a bug upstream in Hadoop-BAM, the VCFHeaderReader class errors when reading
+        // headers from .vcf.gz files
+        //
```
heuermh: add link to https://github.com/HadoopGenomics/Hadoop-BAM/pull/96?

fnothaft:
> @fnothaft do you want to push another commit for GenotypeField and then re-squash? I could pr against this one but it probably isn't worth it for such a small change.

Sorry about the delay on this @heuermh @erictu! Just pushed the change.

heuermh:
+1

erictu:
+1

heuermh:
Merged manually as 449a882e2b70c90d5b26f336a288d99c1db6daf2, 2c073824f574bcd940b66c05f26ad2031e582127 and 525fda83754ddc7082f17c7d4cbe68d47f0efb38.  Thanks!

PR #1017, opened 2016-04-25
bump commons-io version

ryan-williams:
cf. https://github.com/hammerlab/guacamole/pull/444#issuecomment-214047984

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -321,7 +321,7 @@
       <dependency>
         <groupId>commons-io</groupId>
         <artifactId>commons-io</artifactId>
-        <version>1.3.2</version>
+        <version>2.4</version>
       </dependency>
       <dependency>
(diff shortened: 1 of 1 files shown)


fnothaft:
LGTM! Will merge after tests pass.

ryan-williams:
I suppose upgrading further to Hadoop 2.7.0 / Guava 18.0 (as I did in Guacamole) is not desirable here, though, since we're trying to support older versions of Hadoop here?

fnothaft:
> I suppose upgrading further to Hadoop 2.7.0 / Guava 18.0 (as I did in Guacamole) is not desirable here, though, since we're trying to support older versions of Hadoop here?

I think we're holding on 2.6.2 for now.

fnothaft:
Merged. Thanks @ryan-williams!

PR #1027, opened 2016-04-30
 add maxScore

xubo245:
When we map reads to a ref,we need each read maxScore;Moreover ,a read maybe have many seed and each seed need extension,we need max score of eac extension in seed ,then we can take the high score in each max score in extension and decide the read mapping location in ref.
I add the maxScores function and lazy val maxScore


fnothaft:
Closing as we've made the changed interfaces package private to ADAM.

PR #1028, opened 2016-05-03
Logging to be done by ADAM utils code rather than Spark

jpdna:
This PR has a dependency on bdg-utils PR at:
https://github.com/bigdatagenomics/utils/pull/72
and will fail tests without it.

Note: also bumps Spark version 1.6.1

This PR solves the version compatibility logging problems caused by recent versions of Spark making `org.apache.spark.Logging` private to Spark

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,7 +21,7 @@
     <avro.version>1.7.7</avro.version>
     <!-- informative only, used in about template substitution -->
     <scala.version>2.10</scala.version>
-    <spark.version>1.5.2</spark.version>
+    <spark.version>1.6.1</spark.version>
     <parquet.version>1.8.1</parquet.version>
     <!-- Edit the following line to configure the Hadoop (HDFS) version. -->
     <hadoop.version>2.6.0</hadoop.version>
@@ -364,6 +364,17 @@
         <type>test-jar</type>
         <scope>test</scope>
         <exclusions>
+          <exclusion>
+            <groupId>org.apache.spark</groupId>
+            <artifactId>*</artifactId>
+          </exclusion>
+        </exclusions>
+      </dependency>
+      <dependency>
+        <groupId>org.bdgenomics.utils</groupId>
+        <artifactId>utils-misc_2.10</artifactId>
+        <version>${utils.version}</version>
+         <exclusions>
           <exclusion>
             <groupId>org.apache.spark</groupId>
             <artifactId>*</artifactId>
--- a/pom.xml
+++ b/pom.xml
@@ -26,7 +26,7 @@
     <!-- Edit the following line to configure the Hadoop (HDFS) version. -->
     <hadoop.version>2.6.0</hadoop.version>
     <scoverage.version>1.1.1</scoverage.version>
-    <utils.version>0.2.4</utils.version>
+    <utils.version>0.2.6</utils.version>
     <htsjdk.version>1.139</htsjdk.version>
   </properties>
 
--- a/adam-cli/pom.xml
+++ b/adam-cli/pom.xml
@@ -151,6 +151,10 @@
     </dependency>
     <dependency>
       <groupId>org.bdgenomics.utils</groupId>
+      <artifactId>utils-misc_2.10</artifactId>
+    </dependency>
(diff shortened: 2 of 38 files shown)


pom.xml
```diff
@@ -21,12 +21,12 @@
     <avro.version>1.7.7</avro.version>
     <!-- informative only, used in about template substitution -->
     <scala.version>2.10</scala.version>
-    <spark.version>1.5.2</spark.version>
+    <spark.version>1.6.1</spark.version>
```
heuermh: Might want to bump this in a separate pull request to test separately.

fnothaft: Similar to my comment upstream, I would just split this out into a separate commit.

pom.xml
```diff
@@ -370,6 +370,19 @@
           </exclusion>
         </exclusions>
       </dependency>
+
```
fnothaft: Nit: extra spacing around this addition.

jpdna:
Should be ready to go now  - along with its utils  dependency  https://github.com/bigdatagenomics/utils/pull/72

fnothaft:
@jpdna bdg-utils 0.2.5 is released and available through Maven Central now: http://search.maven.org/#artifactdetails%7Corg.bdgenomics.utils%7Cutils-parent_2.10%7C0.2.5%7Cpom

Can you modify the POMs to point at this? Once you're done and tests pass, I can merge.

jpdna:
I don't think the 0.2.5 release pushed to Maven Central contained my Logging PR as intended - and doesn't match https://github.com/bigdatagenomics/utils

I cleared Utils out of  my local mvn repo to be sure that I pulled down the public one, but find that there is no Logging class in `utils-misc_2.10-0.2.5.jar`

fnothaft:
@jpdna just pushed a fixed bdg-utils release. It should propagate to Maven Central in the next few hours.

jpdna:
@fnothaft  - it looks to me like the utils-misc_2.10-0.2.5 that we get from Maven Central still does not contain the Logging class as intended - so still is the old version which didn't include that commit which added Logging.
I've double checkout downloading directly from maven repo, and it doesn't seem to be there.
Any steps I should take to resolve this?

fnothaft:
@jpdna You'll want to look at 0.2.6. Once a release is pushed to Maven Central, I can't overwrite it, so I had to cut a new point release.

jpdna:
This should be ready to merge now if someone can take a look. Thanks!

heuermh:
The diff looks reasonable to me.  Should any of the commits be merged?

Since we've left out the default Spark logging configuration, I wonder how different the logging experience will be for our users.  UCSD had some complaints in this regard.  I spent a little time looking into it before, and came away without a solution or even a good understanding what was going on.  Would we want to look into that as part of this pull request or a follow on?

jpdna:
I have need for this PR in relation to the other Spark 2.0 work,  so if the current behavior still seems at least as good as status quo, which it does to me, then I'd vote for getting this PR in sooner and revisiting the issue of further improving login later.

Wrt to merging the commits, I was splitting out the version bumps as previously requested - so suggest to keep these 3 individual commits seperate as they are now.

fnothaft:
LGTM. I think 062caaa9d32a11261362ef6ad10074bed04dd459 and 33f4756603a3d923fcfdb956aa499cd3e04552a0 should be squashed, since 33f4756603a3d923fcfdb956aa499cd3e04552a0 can't be compiled standalone.

jpdna:
> think 062caaa and 33f4756 should be squashed, since 33f4756 can't be compiled standalone.

Done

heuermh:
+1

heuermh:
@fnothaft could you merge this?  the commit script expects only 1 commit

fnothaft:
Working on the merge now.

fnothaft:
Thanks @jpdna! Merged manually as https://github.com/bigdatagenomics/adam/commit/728459c39a2e3993d837c0f7bd1dc2a439fc1e9e and https://github.com/bigdatagenomics/adam/commit/33841c071f5820a229e9143ee0686c176bc2ac40.

PR #1032, opened 2016-05-13
Error in the README.md for kmer.scala example, need to get rdd first.

InvisibleTech:
When I ran the example from the README for kmer.scala I received errors trying to flatMap:
> error: value flatMap is not a member of org.bdgenomics.adam.rdd.read.AlignmentRecordRDD

By using .rdd:
> reads.rdd.flatMap(_.getSequence.sliding(21).map(k => (k, 1L)))....
It works.

### DIFF
--- a/README.md
+++ b/README.md
@@ -266,6 +266,7 @@ val reads = ac.loadAlignments(
 // Generate, count and sort 21-mers
 val kmers =
   reads
+    .rdd
     .flatMap(_.getSequence.sliding(21).map(k => (k, 1L)))
     .reduceByKey(_ + _)
     .map(_.swap)


fnothaft:
Jenkins, test this please.

fnothaft:
Thanks for the patch, @InvisibleTech! We've just made that API change recently and forgot to patch the docs, so thank you for updating the docs after you ran into the issue!

heuermh:
@fnothaft shouldn't the implicit take care of this?

fnothaft:
@heuermh good point, I was just thinking that. I think the correct approach would be to add the implicit conversion. We just import `import org.bdgenomics.adam.rdd.ADAMContext`, not `import org.bdgenomics.adam.rdd.ADAMContext._`

InvisibleTech:
Got it... I ran the example with the wildcard implicit and it works.  Though... not sure if this a Spark 1.6 or OSX Spark thing but if you don't  end the code lines for the val kmers = .... line you end up with spark shell errors running the kmer.scala sample.  I'll post a new PR for this.  thanks for the review... who'd have thought a README review could be so useful.

InvisibleTech:
Please see https://github.com/bigdatagenomics/adam/pull/1033 for my better fix per your suggestion.

heuermh:
Closing in favor of #1033, thanks!

PR #1036, opened 2016-05-20
Changed loadIndexedBam to use hadoop-bam InputFormat

fnothaft:
A bit of cleanup on top of @andrewmchen's work in #953.


adam-core/src/main/scala/org/bdgenomics/adam/models/Interval.scala
```diff
@@ -41,4 +42,19 @@ trait Interval {
    */
   def width: Long = end - start
 
+  /**
+   * Need to implement getStart function from Locatable. (1-based start position. closed)
```
heuermh: Mixing 0- and 1- based coordinate systems in the same class gives me the hives. Can we hide this as an implementation detail in an adapter somewhere in the i/o code?

fnothaft: I don't think we can hide it, but we could mark it as `@deprecated` to hint that it is only for use with legacy codebases?

heuermh: I guess I don't see where it is used, so I'm not sure what suggestion to make.  Will have to dig a bit.

fnothaft: No, this allows us to supply `ReferenceRegion`s to the Hadoop-BAM indexed input format code without going through an intermediate class.

fnothaft: Actually, think I may have misread.

> And is this the Interval that will need to go into bdg-utils to support IntervalRDD? Then there's no way I want to mix htsjdk stuff in.

@akmorrow13 @erictu, can you all comment?

fnothaft: Actually, I get what you're saying now, @heuermh. I will refactor this out.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -325,8 +325,7 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
    *        a single Bam file. The bam index file associated needs to have the same name.
    * @param viewRegion The ReferenceRegion we are filtering on
    */
-  def loadIndexedBam(
-    filePath: String, viewRegion: ReferenceRegion): RDD[AlignmentRecord] = {
+  def loadIndexedBam(filePath: String, viewRegion: ReferenceRegion): RDD[AlignmentRecord] = {
```
erictu: Should this use AlignmentRecordRDD?

fnothaft:
Rebased and addressed comments. Ping @heuermh @erictu for review.

adam-core/src/main/java/org/bdgenomics/adam/io/IndexedBamInputFormat.scala
```diff
@@ -1,102 +0,0 @@
...
-        val end: Long = chunk.getChunkEnd()
-        val locs = Array[String]()
-        val newSplit = new FileVirtualSplit(filePath, start, end, locs)
-        splits += newSplit
-      }
-      splits.toList
-    }
-  }
-
-}
```
heuermh: oh how I like to see code being removed  :)

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -63,6 +64,13 @@ import scala.collection.JavaConversions._
 import scala.collection.Map
 import scala.reflect.ClassTag
 
+// only used with indexedbamload
+private case class LocatableReferenceRegion(rr: ReferenceRegion) extends Locatable {
```
heuermh: perfect

heuermh:
LGTM

adam-core/src/main/java/org/bdgenomics/adam/io/IndexedBamInputFormat.scala
fnothaft: Indeed; it is uniquely satisfying!

erictu:
+1

fnothaft:
Rebased and ready to merge!

heuermh:
Oops, didn't let Jenkins have a final say before merging.  Builds & tests ok for me locally.

Thank you, @fnothaft @andrewmchen!

fnothaft:
Looks like the @AmplabJenkins failures were just the git `failed to determine` issue. We should be a-OK but I'll follow up if the ToT build fails.

PR #1043, opened 2016-06-03
overlap query reflects new formats

### LINKED ISSUE #1042: filterByOverlappingRegion Incorrect for Genotypes
Currently this:
```
  def filterByOverlappingRegion(query: ReferenceRegion): RDD[Genotype] = {
    def overlapsQuery(rec: Genotype): Boolean =
      rec.getVariant.getContigName == query.referenceName &&
        rec.getVariant.getStart < query.end &&
        rec.getVariant.getEnd > query.start
    rdd.filter(overlapsQuery)
  }
```
Should be this, where we get `start`, `end`, and `contigName` directly from the `genotype`. (I will submit a PR):
```
  def filterByOverlappingRegion(query: ReferenceRegion): RDD[Genotype] = {
    def overlapsQuery(rec: Genotype): Boolean =
      rec.getContigName == query.referenceName &&
        rec.getStart < query.end &&
        rec.getEnd > query.start
    rdd.filter(overlapsQuery)
  }
```
It seems like when we convert VCF to ADAM, the ADAM format won't have the `start`, `end`, and `contigName` fields populated in the `variant` record, but they will be populated in the `genotype` record.

The filter condition should get the `start`, `end`, and `contigName` fields from the `genotype` record, not the `variant` record. Is there any reason why the fields are null in the `variant`? Is this just so we don't replicate information?

erictu:
Resolves #1042

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/variation/GenotypeRDD.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/variation/GenotypeRDD.scala
@@ -35,9 +35,9 @@ case class GenotypeRDD(rdd: RDD[Genotype],
 
   def filterByOverlappingRegion(query: ReferenceRegion): RDD[Genotype] = {
     def overlapsQuery(rec: Genotype): Boolean =
-      rec.getVariant.getContigName == query.referenceName &&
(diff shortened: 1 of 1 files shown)


heuermh:
+1

heuermh:
Thanks, @erictu!

Issue #1044, opened 2016-06-05
Support VCF annotation ANN field in vcf2adam and adam2vcf


heuermh:
Schema were added to `bdg-formats` in https://github.com/bigdatagenomics/bdg-formats/pull/67.

fnothaft:
Should be resolved by #1250.

heuermh:
#1250 doesn't fully address this yet.

Mappings between the VCF INFO reserved key ANN and calls to `TranscriptEffectConverter.convertToTranscriptEffects` and `TranscriptEffectConverter.convertToVcfInfoAnnValue` need to be added to `VariantAnnotationConverter` as part of further refactoring.

jpdna:
#1271 should address this correct?

heuermh:
No, I haven't written the code yet. :)

Look for a new pull request that says it fixes this issue.

heuermh:
Fixed by #1288

PR #1062, opened 2016-07-01
Cleanup in org.bdgenomics.adam.converters package.

fnothaft:
* Remove unused GenotypesToVariantsConverter.
* Made all classes package private to [converters] if they had no external references, or package private to [adam] if they were referenced elsewhere in ADAM.
* Added documentation where necessary.
* Added SupportedHeaderLines object. This contains code previously in VariantAnnotationConverter that was called in the ADAMVCFOutputFormat. This allows us to make VariantAnnotationConverter package private to converters and allows us to make the AttrKeys class and object private.
* Additionally, pulled `mergeAnnotations` out to org.bdgenomics.adam.models.VariantContext, which is also needed to make VariantAnnotationConverter package private to converters. This also cleaned up some code in org.bdgenomics.adam.rdd.variation.VariationRDDFunctions.


fnothaft:
Jenkins, retest this please.

heuermh:
Thanks, I will probably want this merged before I pull request on the VariantAnnotation stuff (which unfortunately has some name collisions with what is there already).  Will review this afternoon.

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantAnnotationConverter.scala
```diff
@@ -133,19 +264,42 @@ object VariantAnnotationConverter extends Serializable {
...
   }
 
-  private def fillKeys[T <% SpecificRecord](keys: Seq[AttrKey], vc: VariantContext, record: T): T = {
-    fillRecord(createFieldMap(keys, record.getSchema), vc, record)
-  }
-
-  def convert(vc: VariantContext, annotation: DatabaseVariantAnnotation): DatabaseVariantAnnotation = {
+  /**
+   * Remaps fields from an htsjdk variant context into a site annotation.
+   *
```
heuermh: I've been searching for a better name for `DatabaseVariantAnnotation`, could `VariantSiteAnnotation` be it?

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantContextConverter.scala
```diff
@@ -219,10 +299,27 @@ class VariantContextConverter(dict: Option[SequenceDictionary] = None) extends S
...
 
+  /**
+   * Builds an avro Variant for a site with a defined alt allele.
+   *
+   * @param vc htsjdk variant context to use for building the site.
+   * @param alt The alternate allele to use for the site. If not provided, no
+   *   alternate allele will be defined.
+   * @return Returns an Avro description of the genotyped site.
+   */
   private def createADAMVariant(vc: BroadVariantContext, alt: Option[String]): Variant = {
```
heuermh: rename `BroadVariantContext` &rarr; `HtsjdkVariantContext`?

heuermh:
For #1044 I currently have a method
```scala
def createVariantAnnotations(variant: Variant, vc: VariantContext): VariantAnnotation = {
  ...
```
that pulls the `ANN` keyed INFO field attribute out of the variant context and populates an avro `VariantAnnotation` record (which contains an array of `TranscriptEffect` records).

This could be wrapped in a third `convert` method in `VariantAnnotationConverter`, however going back to https://github.com/bigdatagenomics/bdg-formats/issues/41 and related discussion around https://github.com/bigdatagenomics/bdg-formats/pull/67, is it really a good idea to have three different code paths around data associated with Variant?  Do we need three different sets of schema records?

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantAnnotationConverter.scala
fnothaft: I would +1 that! `DatabaseVariantAnnotation` is totes clunky.

fnothaft:
Addressed comment RE: `HtsjdkVariantContext`. That's a better name, given the repackaging in `htsjdk`. `VariantContext` used to live in `org.broadinstitute.variant`, IIRC.

fnothaft:
Thanks for bringing the following up, @heuermh:

> This could be wrapped in a third convert method in VariantAnnotationConverter, however going back to bigdatagenomics/bdg-formats#41 and related discussion around bigdatagenomics/bdg-formats#67, is it really a good idea to have three different code paths around data associated with Variant? Do we need three different sets of schema records?

I agree. I like having annotation and genotype separate, but I don't think we need two different annotation records (for 3 total paths). Perhaps we can sit down this next week and go through the paths with a fine toothed comb? I hadn't gone through this part of the codebase line-by-line in a while, and it was harrier than I remembered...

heuermh:
LGTM

heuermh:
Merged commit 038680c59d24b207adad1b5c80d685865aa5684e

Thank you, @fnothaft!

PR #1069, opened 2016-07-06
[ADAM-1044] Support VCF annotation ANN field

heuermh:
Work towards #1044.
Will need further discussion around conversion from VariantContext as mentioned in #1062.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantAnnotations.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantAnnotations.scala
@@ -1,0 +1,157 @@
+/**
+ * Licensed to Big Data Genomics (BDG) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The BDG licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.bdgenomics.adam.converters
+
+import htsjdk.samtools.ValidationStringency
+import htsjdk.variant.vcf.VCFConstants
+import htsjdk.variant.variantcontext.VariantContext
+import org.bdgenomics.formats.avro.VariantAnnotationMessage._
+import org.bdgenomics.formats.avro.{
+  TranscriptEffect,
+  Variant,
+  VariantAnnotation,
+  VariantAnnotationMessage
+}
+import org.bdgenomics.utils.misc.Logging
+import scala.collection.JavaConverters._
+
+object VariantAnnotations extends Serializable with Logging {
+  private val MESSAGES: Map[String, VariantAnnotationMessage] = Map(
+    // name -> enum
+    ERROR_CHROMOSOME_NOT_FOUND.name() -> ERROR_CHROMOSOME_NOT_FOUND,
+    ERROR_OUT_OF_CHROMOSOME_RANGE.name() -> ERROR_OUT_OF_CHROMOSOME_RANGE,
+    WARNING_REF_DOES_NOT_MATCH_GENOME.name() -> WARNING_REF_DOES_NOT_MATCH_GENOME,
+    WARNING_SEQUENCE_NOT_AVAILABLE.name() -> WARNING_SEQUENCE_NOT_AVAILABLE,
+    WARNING_TRANSCRIPT_INCOMPLETE.name() -> WARNING_TRANSCRIPT_INCOMPLETE,
+    WARNING_TRANSCRIPT_MULTIPLE_STOP_CODONS.name() -> WARNING_TRANSCRIPT_MULTIPLE_STOP_CODONS,
+    WARNING_TRANSCRIPT_NO_START_CODON.name() -> WARNING_TRANSCRIPT_NO_START_CODON,
+    INFO_REALIGN_3_PRIME.name() -> INFO_REALIGN_3_PRIME,
+    INFO_COMPOUND_ANNOTATION.name() -> INFO_COMPOUND_ANNOTATION,
+    INFO_NON_REFERENCE_ANNOTATION.name() -> INFO_NON_REFERENCE_ANNOTATION,
+    // message code -> enum
+    "E1" -> ERROR_CHROMOSOME_NOT_FOUND,
+    "E2" -> ERROR_OUT_OF_CHROMOSOME_RANGE,
+    "W1" -> WARNING_REF_DOES_NOT_MATCH_GENOME,
+    "W2" -> WARNING_SEQUENCE_NOT_AVAILABLE,
+    "W3" -> WARNING_TRANSCRIPT_INCOMPLETE,
+    "W4" -> WARNING_TRANSCRIPT_MULTIPLE_STOP_CODONS,
+    "W5" -> WARNING_TRANSCRIPT_NO_START_CODON,
+    "I1" -> INFO_REALIGN_3_PRIME,
+    "I2" -> INFO_COMPOUND_ANNOTATION,
+    "I3" -> INFO_NON_REFERENCE_ANNOTATION
+  )
+
+  private def parseEffects(s: String): List[String] = {
+    s.split("&").toList
+  }
+
+  private def parseMessages(s: String): List[VariantAnnotationMessage] = {
+    // todo: haven't seen a delimiter here, assuming it is also '&'
+    s.split("&").map(MESSAGES.get(_)).toList.flatten
+  }
+
+  private def parseFraction(s: String): (Option[Integer], Option[Integer]) = {
+    if ("".equals(s)) {
+      return (None, None)
+    }
+    val tokens = s.split("/")
+    tokens.length match {
+      case 0 => (None, None)
+      case 1 => (Some(Integer.parseInt(tokens(0))), None)
+      case _ => (Some(Integer.parseInt(tokens(0))), Some(Integer.parseInt(tokens(1))))
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


heuermh:
Rebased.  This can go in as-is, the further work described above will go in #1078.

fnothaft:
Thanks @heuermh! I will make a review pass today.

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantAnnotations.scala
```diff
@@ -0,0 +1,140 @@
...
+package org.bdgenomics.adam.converters
+
+import htsjdk.variant.vcf.VCFConstants
+import htsjdk.variant.variantcontext.VariantContext
+import org.bdgenomics.formats.avro.{ TranscriptEffect, Variant, VariantAnnotation, VariantAnnotationMessage }
+import org.bdgenomics.formats.avro.VariantAnnotationMessage._
+import org.bdgenomics.utils.misc.Logging
+import scala.collection.JavaConverters._
+
+object VariantAnnotations extends Serializable with Logging {
```
fnothaft: VariantAnnotations --> VariantAnnotationsConverter, to be consistent with the rest of the package?

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantAnnotations.scala
```diff
@@ -0,0 +1,140 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.bdgenomics.adam.converters
+
+import htsjdk.variant.vcf.VCFConstants
+import htsjdk.variant.variantcontext.VariantContext
+import org.bdgenomics.formats.avro.{ TranscriptEffect, Variant, VariantAnnotation, VariantAnnotationMessage }
```
fnothaft: Split long line.

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantAnnotations.scala
```diff
@@ -0,0 +1,140 @@
...
+import org.bdgenomics.formats.avro.VariantAnnotationMessage._
```
fnothaft: Prefer `._` imports above imports from same package.

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantAnnotations.scala
```diff
@@ -0,0 +1,140 @@
...
+
+  private def notEmpty(s: String): Boolean = Option(s) match {
+    case None    => false
+    case Some(s) => s.length > 0
+  }
+
+  private[converters] def parseTranscriptEffect(s: String): Seq[TranscriptEffect] = {
+    val fields = s.split("\\|", -1)
+    if (fields.length < 16) {
+      if (log.isWarnEnabled) log.warn("Invalid VCF ANN attribute, expected at least 16 fields, found {}: {}", fields.length, fields)
```
fnothaft: What's the `log.isWarnEnabled` for? Might this be a better place to use a `ValidationStringency` check?

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantAnnotations.scala
```diff
@@ -0,0 +1,140 @@
...
+    }
+    val tokens = s.split("/")
+    tokens.length match {
+      case 0 => (None, None)
+      case 1 => (Some(Integer.parseInt(tokens(0))), None)
+      case _ => (Some(Integer.parseInt(tokens(0))), Some(Integer.parseInt(tokens(1))))
+    }
+  }
+
+  private def notEmpty(s: String): Boolean = Option(s) match {
```
fnothaft: Nit: prefer fold over match. E.g.:

```
Option(s).fold(false)(s => s.nonEmpty)
```

`nonEmpty` should come in implicitly from http://www.scala-lang.org/api/2.10.3/index.html#scala.collection.immutable.StringOps.

fnothaft: Also, perhaps you could factor this a bit further into:

```
def setIfNotEmpty(s: String, setFn: String => Unit) {
  Option(s).filter(_.nonEmpty)
    .foreach(setFn)
}
```

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantAnnotations.scala
```diff
@@ -0,0 +1,140 @@
...
+    // multivalued fields
+    val effects = parseEffects(fields(1))
+    val (rank, total) = parseFraction(fields(8))
+    val (cdnaPosition, cdnaLength) = parseFraction(fields(11))
+    val (cdsPosition, cdsLength) = parseFraction(fields(12))
+    val (proteinPosition, proteinLength) = parseFraction(fields(13))
+    val messages = parseMessages(fields(15))
+
+    val te = TranscriptEffect.newBuilder()
+    if (notEmpty(alternateAllele)) te.setAlternateAllele(alternateAllele)
```
fnothaft: If you do https://github.com/bigdatagenomics/adam/pull/1069/files#r71388866, then you can eliminate all of these if statements.

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantAnnotations.scala
```diff
@@ -0,0 +1,140 @@
...
+    total.foreach(te.setTotal(_))
+    if (notEmpty(transcriptHgvs)) te.setTranscriptHgvs(transcriptHgvs)
+    if (notEmpty(proteinHgvs)) te.setProteinHgvs(proteinHgvs)
+    cdnaPosition.foreach(te.setCdnaPosition(_))
+    cdnaLength.foreach(te.setCdnaLength(_))
+    cdsPosition.foreach(te.setCdsPosition(_))
+    cdsLength.foreach(te.setCdsLength(_))
+    proteinPosition.foreach(te.setProteinPosition(_))
+    proteinLength.foreach(te.setProteinLength(_))
+    if (notEmpty(distance)) te.setDistance(Integer.parseInt(distance))
```
fnothaft: I might try catch around this to throw a more informative error message if parsing fails.

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantAnnotations.scala
```diff
@@ -0,0 +1,140 @@
...
+    if (!messages.isEmpty) te.setMessages(messages.asJava)
+
+    Seq(te.build())
```
fnothaft: Why are we returning a Seq here?

fnothaft:
Generally looks good. Could you add scaladoc inline?

Also, where will this code attach in? Will it get connected up in #1078?

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantAnnotations.scala
heuermh: In #1078 the code in this class will be merged with the already existing `VariantAnnotationConverter`.  Would you still want that change here?

fnothaft: Oh, good point. Nevermind then!

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantAnnotations.scala
heuermh: Avro-generated classes track whether or not a field has been set (enabling compression for un-set fields I would presume) so it is important not to set a field where the value is null or empty.  This reads nice; for `null` the option won't be true, then `_.nonEmpty` won't NPE, and `setFn` is only called if `s` is non-empty.

fnothaft: Exactly! I could wax philosophical about how I love the `Option` type in Scala...

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantAnnotations.scala
heuermh: We don't have log level checks in `o.b.utils.Logging` methods, so this uses `log` directly.

`ValidationStringency` would be nice to have, but would need to be pushed down all the way through VCF parsing, and I'm not sure Hadoop-BAM supports it.

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantAnnotations.scala
heuermh: Same pattern used in feature parsing, `TranscriptEffect` association is an array in `VariantAnnotation` schema and a `java.util.List<TranscriptEffect>` in the Avro-generated object.

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantAnnotations.scala
fnothaft: > We don't have log level checks in o.b.utils.Logging methods

Really? Drat!

> `ValidationStringency` would be nice to have, but would need to be pushed down all the way through VCF parsing, and I'm not sure Hadoop-BAM supports it.

We call these functions in a `map` over the RDD of `VariantContext`s we get from Hadoop-BAM, so we can pass it down. I think it'd be good to have here, as it is likely desirable to fail on a badly formatted annotation line.

heuermh: ok, will try to push it down then

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantAnnotations.scala
heuermh: I've added the validation stringency stuff for invalid ANN lines but am not sure what to do for NumberFormatExceptions.  They aren't explicitly handled elsewhere.

heuermh:
> Also, where will this code attach in? Will it get connected up in #1078?

Yes, I'll also add scaladoc at that point, since the method signatures will probably change to look more like the other converter classes.

jpdna:
+1 - this looks good and seems self contained at least looking backwards, I am going to merge now so that related PRs can build on top of.

jpdna:
Merged: c23aace7b173
Thanks @heuermh !

PR #1075, opened 2016-07-06
[ADAM-194] Use separate filtersFailed and filtersPassed arrays for variant quality filters

### LINKED ISSUE #194: Variant filter representation
VCF format specifies that the FILT column can take on the following values:

"PASS" - variant passed all filters
Comma-separated list of failing filters - filters that the variant call failed
"." - no filter applied

(see http://gatkforums.broadinstitute.org/discussion/3401/filter-column-in-vcf-file-from-haplotypecaller for a discussion on the output of the haplotype caller)

Right now, we have two fields in VariantCallingAnnotations:

``` c++
boolean variantIsPassing
array<string> variantFilters
```

This table indicates the value of each variable for the different scenarios:

| Scenario | variantIsPassing | variantFilters |
| --- | :-: | :-: |
| Variant Passed | true | empty |
| Variant failed filter | false | failing filter list |
| No filters applied | true | empty |

@mlinderm brought up that there is some ambiguity - we don't know the difference between variant passing and no filters applied without checking another field (the list of filters applied), and also people doing analysis were only likely to check the boolean, since that's what is done with the VCF field.  The proposal is to go back to the original way, which was:

``` c++
union { null, boolean }  variantIsFiltered = null;
array<string> failingFilters = null;
```

New table:

| Scenario | variantIsFiltered | failingFilters |
| --- | :-: | :-: |
| Variant Passed | false | empty |
| Variant failed filter | true | failing filter list |
| No filters applied | null | empty |

Note that in the case where the variant passed, the passing filters should still be available from the VCF header. 

Thoughts?

heuermh:
Fixes #194
Depends on https://github.com/bigdatagenomics/bdg-formats/pull/88

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantContextConverter.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantContextConverter.scala
@@ -495,9 +495,9 @@ private[adam] class VariantContextConverter(dict: Option[SequenceDictionary] = N
 
     // VCF QUAL, FILTER and INFO fields
     if (vc.filtersWereApplied && vc.isFiltered) {
-      call.setVariantIsPassing(false).setVariantFilters(new java.util.ArrayList(vc.getFilters))
+      call.setFiltersFailed(new java.util.ArrayList(vc.getFilters))
     } else if (vc.filtersWereApplied) {
-      call.setVariantIsPassing(true)
+      call.setFiltersPassed(new java.util.ArrayList(vc.getFilters))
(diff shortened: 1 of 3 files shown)


heuermh:
In order for this to build cleanly, it needs a specific snapshot version (not the latest one, in other words) of bdg-formats.  Should I close this pull request and move the commit over to #1078?

heuermh:
I believe I was able to properly cherry-pick the commit here to #1078.  Feel free to reopen if I git-borked it up.

PR #1051, opened 2016-06-15
[ADAM-1011] Refactor to add GenomicRDDs for all Avro types

fnothaft:
A bit of work towards #1011. I'll be dropping the rest of #1011 onto here. Not ready for merge yet, but ready for first review.

PR #1079, opened 2016-07-13
Extended TwoBitFile and NucleotideContigFragmentRDDFunctions to behave more similar

akmorrow13:
* Extended NucleotideContigFragmentRDDFunctions to ReferenceFile
* Added getSequenceDictionary function to TwoBitFile

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/util/TwoBitFile.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/util/TwoBitFile.scala
@@ -22,7 +22,7 @@ import java.nio.{ ByteOrder, ByteBuffer }
 import com.esotericsoftware.kryo.io.{ Output, Input }
 import com.esotericsoftware.kryo.{ Kryo, Serializer }
 import org.bdgenomics.utils.io.{ ByteArrayByteAccess, ByteAccess }
-import org.bdgenomics.adam.models.{ NonoverlappingRegions, ReferencePosition, ReferenceRegion }
+import org.bdgenomics.adam.models._
 
 object TwoBitFile {
   val MAGIC_NUMBER: Int = 0x1A412743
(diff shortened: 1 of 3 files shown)


fnothaft:
@akmorrow13 let's chat about this sometime. `NucleotideContigFragmentRDDFunctions` goes away in #1051 . The possibly misnamed [ReferenceContigMap](https://github.com/bigdatagenomics/adam/blob/b29d20480001be20e0be55e828a147b5f222ce42/adam-core/src/main/scala/org/bdgenomics/adam/util/ReferenceContigMap.scala) class may be sufficient for your needs.

adam-core/src/main/scala/org/bdgenomics/adam/util/TwoBitFile.scala
```diff
@@ -92,6 +92,16 @@ class TwoBitFile(byteAccess: ByteAccess) extends ReferenceFile {
     name -> contigOffset
   }
 
+  def getSequenceDictionary(performLexSort: Boolean = false): SequenceDictionary = {
+    val sd = new SequenceDictionary(seqRecords.toVector.map(r => SequenceRecord(r._1, r._2.dnaSize)))
+
+    if (performLexSort) {
```
fnothaft: Why's this necessary?

akmorrow13: I need the sequence dictionary from 2 bit file. Is there a better way to get it?

fnothaft: Oh, yeah, that makes sense. My comment was with regards to the lexicographic sort.

fnothaft: IIRC, we've added lex sort functions to the `SequenceDictionary`, so I'd rather not dupe the code.

PR #1091, opened 2016-07-20
[ADAM-582] Eliminate .get on option in FragmentCoverter.

### LINKED ISSUE #582: Improve error message when we can't find a ReferenceRegion for a contig
Seen on the [mailing list](https://groups.google.com/forum/#!topic/adam-developers/lgkvNbxrrK0). If calling `ReferenceRegion(contig: NucleotideContigFragment)` returns `None`, we throw a nondescript error.

fnothaft:
Resolves #582. In the FragmentConverter, we had code that called .get on an optional ReferenceRegion. If a NucleotideContigFragment didn't have a contig associated with it, then we would call .get on the empty option, which would throw an error. This commit changes that code to call map on the Option[ReferenceRegion], and then changes the usage of that method so that it is called in a flatMap. Additionally, I've added a test to cover this case.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/converters/FragmentConverter.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/converters/FragmentConverter.scala
@@ -35,11 +35,11 @@ private object FragmentCollector extends Serializable {
    * @return Returns key value pair where the key is the contig metadata and the
    *   value is a Fragment Collector object.
    */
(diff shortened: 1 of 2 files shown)


heuermh:
+1

heuermh:
Merged commit 7c2df6024fddba86df32a83c84c054725803aec6

Thank you, @fnothaft!

PR #1095, opened 2016-07-31
Added support for Indexed VCF files

akmorrow13:
* added functionality to load vcf using a tabix file


fnothaft:
Nice! I will do a deep dive on this next week. Exciting work!

PR #1096, opened 2016-07-31
added loadIndexedVcf and loadIndexedBam for multiple ReferenceRegions

akmorrow13:


adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -623,7 +631,41 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+   * Loads a VCF file indexed by a tabix (tbi) file into an RDD.
+   *
+   * @param filePath The file to load.
+   * @param viewRegion The ReferenceRegion we are filtering on.
+   * @param sdOpt An optional sequence dictionary, in case the sequence info
+   *              is not included in the VCF.
+   * @return Returns a VariantContextRDD.
+   */
+  def loadIndexedVcf(filePath: String,
+                     viewRegion: ReferenceRegion,
```
heuermh: Would it be useful to support multiple view regions?

akmorrow13: I'm not sure, but then we should support multiple regions for loadIndexedBam as well

heuermh: Good point.  If this were a Java API, I would propose two methods for both, one with vararg and one with Iterable
```java
void loadIndexedVcf(String filePath, ReferenceRegion... viewRegions) { ... }
void loadIndexedVcf(String filePath, Iterable<ReferenceRegion> viewRegions) { ... }
```
I'm not sure what to do in Scala, and then there's the complication of the defaulted sdOpt parameter, and that varargs parameters need to be last . . .

akmorrow13: I don't know how to get around this either. We could require that the user defines an Iterable[ReferenceRegion] to being with, but that is messy. Anyone else have a recommendation for this issue? There is some talk of supporting overloading in Scala 2.12, but that is a ways off.

heuermh: Well, the vararg thing is just a convenience.  For Scala, couldn't we just have two methods
```scala
def loadIndexedVcf(filePath: String, viewRegion: ReferenceRegion, sdOpt: Option[SequenceDictionary] ...
def loadIndexedVcf(filePath: String, viewRegions: Iterable[ReferenceRegion], sdOpt: Option[SequenceDictionary] ...
```

akmorrow13: No, you cannot overload methods in scala with the same number of parameters.

heuermh: Ah, well that's dumb.  😄

akmorrow13: Correction, the issue is sdOpt: Option[SequenceDictionary] = None. This throws an error for overloaded methods defining default arguments.

heuermh:
LGTM

jpdna:
LGTM

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
fnothaft: I would remove `sdOpt` (here and everywhere).

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -596,16 +597,23 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+  private def readVcfRecords(filePath: String, viewRegion: Option[ReferenceRegion]): RDD[(LongWritable, VariantContextWritable)] = {
     // load vcf data
     val job = HadoopUtil.newJob(sc)
     job.getConfiguration().setStrings("io.compression.codecs",
       classOf[BGZFCodec].getCanonicalName(),
       classOf[BGZFEnhancedGzipCodec].getCanonicalName())
+
+    val conf = ContextUtil.getConfiguration(job)
+    if (viewRegion.isDefined) {
+      val interval: Interval = new Interval(viewRegion.get.referenceName, viewRegion.get.start.toInt, viewRegion.get.end.toInt)
```
fnothaft: IIRC, this is off by 1. I always lose track of this, but I am 95% sure that HTSJDK is 1-based closed intervals, while we're 0-based open intervals.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
fnothaft: Also, I would make a ticket for the multiple reference regions. I think it is a logical thing for us to support.

adam-core/src/test/scala/org/bdgenomics/adam/rdd/ADAMContextSuite.scala
```diff
@@ -423,5 +423,12 @@ class ADAMContextSuite extends ADAMFunSuite {
     val reads = sc.loadIndexedBam(path, refRegion)
     assert(reads.rdd.count == 1)
   }
+
+  sparkTest("loadIndexedVcf") {
+    val path = resourcePath("bqsr1.vcf")
+    val refRegion = ReferenceRegion("22", 0, 16050822)
```
fnothaft: Can we use a more specific predicate than this? Ideally, we'd use one that illustrates that there is not a coordinate conversion issue.

adam-core/src/test/scala/org/bdgenomics/adam/rdd/ADAMContextSuite.scala
```diff
@@ -423,5 +423,12 @@ class ADAMContextSuite extends ADAMFunSuite {
...
+    val reads2 = sc.loadIndexedVcf(path, refRegion)
```
fnothaft: `s/reads/variants/g | tr -d 2`

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
heuermh: +1

akmorrow13: It doesn't seem that this is a problem for either vcf or bam indexes. If I start a region at 10 and the variant also starts at position 10, it will still retrieve that variant.

fnothaft:
Thanks for confirming that the index bit is not an issue, @akmorrow13!

fnothaft:
Merged! Thanks @akmorrow13!

fnothaft:
Also, just to confirm, there wasn't an issue for this, right? If there was, can you close the issue? I looked through the tracker but a cursory search didn't point me to one.

Issue #1059, opened 2016-06-28
Support writing IntervalList header for features


heuermh:
Follow on to #710, see also #1009

Issue #1058, opened 2016-06-28
Add -single support when writing features in native formats


heuermh:
Follow on to #710

Issue #676, opened 2015-05-26
BAMHeader not set when running on a cluster


arahuja:
When running `transform` with `.bam` file as the output, I see the following error:
```
Exception in thread "main" java.lang.AssertionError: assertion failed: Cannot return header if not attached.
        at scala.Predef$.assert(Predef.scala:179)
        at org.bdgenomics.adam.rdd.read.ADAMBAMOutputFormat$.getHeader(ADAMBAMOutputFormat.scala:60)
        at org.bdgenomics.adam.rdd.read.ADAMBAMOutputFormat.<init>(ADAMBAMOutputFormat.scala:68)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:57)
        at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
        at java.lang.Class.newInstance(Class.java:379)
        at org.apache.spark.rdd.InstrumentedOutputFormat.<init>(InstrumentedOutputFormat.scala:33)
        at org.bdgenomics.adam.rdd.read.InstrumentedADAMBAMOutputFormat.<init>(ADAMBAMOutputFormat.scala:71)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
        at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:57)
        at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
        at java.lang.reflect.Constructor.newInstance(Constructor.java:526)
        at java.lang.Class.newInstance(Class.java:379)
        at org.apache.spark.rdd.PairRDDFunctions.saveAsNewAPIHadoopDataset(PairRDDFunctions.scala:958)
```

I tried to fix this with a similar strategy used by @fnothaft for the VCF header (see https://github.com/bigdatagenomics/adam/compare/master...hammerlab:bam-save-header?expand=1), but did not seem to take.  Any ideas on this?

ryan-williams:
IIRC @fnothaft and I could not convince ourselves that the VCF header impl works like a year ago, though he claimed it was working for him (on a cluster) and I couldn't get it to work.

[...]

PR #686, opened 2015-05-31
add .interval_list file parser

ryan-williams:
fixes #685

PR #1104, opened 2016-08-08
Resolves various single file save/header issues

### LINKED ISSUE #1009: Port single file BAM and header code to VCF
I was looking over the VCF export code while working on #909 and noticed that we had a lot of cruft around setting the VCF headers. Also, we don't have code for writing out a single VCF file.

fnothaft:
Resolves #1009, #1059, #1058, #676.

### DIFF
--- a/adam-cli/src/main/scala/org/bdgenomics/adam/cli/TransformFeatures.scala
+++ b/adam-cli/src/main/scala/org/bdgenomics/adam/cli/TransformFeatures.scala
@@ -33,11 +33,11 @@ object TransformFeatures extends BDGCommandCompanion {
 
 class TransformFeaturesArgs extends Args4jBase with ParquetSaveArgs {
   @Argument(required = true, metaVar = "INPUT",
-    usage = "The features file to convert (e.g., .bed, .gff)", index = 0)
+    usage = "The features file to convert (e.g., .bed, .gff). If extension is not detected, Parquet is assumed.", index = 0)
   var featuresFile: String = _
 
   @Argument(required = true, metaVar = "OUTPUT",
-    usage = "Location to write ADAM features data", index = 1)
+    usage = "Location to write ADAM features data. If extension is not detected, Parquet is assumed.", index = 1)
   var outputPath: String = null
 
   @Args4jOption(required = false, name = "-num_partitions",
--- a/adam-cli/src/main/scala/org/bdgenomics/adam/cli/TransformFeatures.scala
+++ b/adam-cli/src/main/scala/org/bdgenomics/adam/cli/TransformFeatures.scala
@@ -22,32 +22,39 @@ import org.bdgenomics.adam.rdd.ADAMContext._
 import org.bdgenomics.utils.cli._
 import org.kohsuke.args4j.{ Argument, Option ⇒ Args4jOption }
 
-object Features2ADAM extends BDGCommandCompanion {
-  val commandName = "features2adam"
-  val commandDescription = "Convert a file with sequence features into corresponding ADAM format"
+object TransformFeatures extends BDGCommandCompanion {
+  val commandName = "transformFeatures"
+  val commandDescription = "Convert a file with sequence features into corresponding ADAM format and vice versa"
 
   def apply(cmdLine: Array[String]) = {
-    new Features2ADAM(Args4j[Features2ADAMArgs](cmdLine))
+    new TransformFeatures(Args4j[TransformFeaturesArgs](cmdLine))
   }
 }
 
-class Features2ADAMArgs extends Args4jBase with ParquetSaveArgs {
-  @Argument(required = true, metaVar = "FEATURES",
+class TransformFeaturesArgs extends Args4jBase with ParquetSaveArgs {
+  @Argument(required = true, metaVar = "INPUT",
     usage = "The features file to convert (e.g., .bed, .gff)", index = 0)
   var featuresFile: String = _
-  @Argument(required = true, metaVar = "ADAM",
+
+  @Argument(required = true, metaVar = "OUTPUT",
     usage = "Location to write ADAM features data", index = 1)
   var outputPath: String = null
+
   @Args4jOption(required = false, name = "-num_partitions",
     usage = "Number of partitions to load an interval file using.")
   var numPartitions: Int = _
+
+  @Args4jOption(required = false, name = "-single",
+    usage = "Save as a single file, for the text formats.")
+  var single: Boolean = false
 }
 
-class Features2ADAM(val args: Features2ADAMArgs)
-    extends BDGSparkCommand[Features2ADAMArgs] {
-  val companion = Features2ADAM
+class TransformFeatures(val args: TransformFeaturesArgs)
+    extends BDGSparkCommand[TransformFeaturesArgs] {
+  val companion = TransformFeatures
 
   def run(sc: SparkContext) {
-    sc.loadFeatures(args.featuresFile, None, Option(args.numPartitions)).saveAsParquet(args)
+    sc.loadFeatures(args.featuresFile, None, Option(args.numPartitions))
+      .save(args.outputPath, args.single)
   }
 }
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
@@ -17,6 +17,7 @@
  */
 package org.bdgenomics.adam.rdd.features
 
+import htsjdk.samtools.ValidationStringency
 import java.io.File
 import org.bdgenomics.adam.models.SequenceRecord
 import org.bdgenomics.formats.avro.{ Dbxref, Feature, Strand }
@@ -73,12 +74,17 @@ class GTFParser extends FeatureParser {
 
     val fields = line.split("\t")
     // skip empty or invalid lines
-    if (fields.length < 9) {
-      if (log.isDebugEnabled) log.debug("Empty or invalid GTF/GFF2 line: {}", line)
+    if (fields.length < 8 || fields.length > 9) {
+      log.warn("Empty or invalid GTF/GFF2 line: {}", line)
       return Seq()
     }
-    val (seqname, source, featureType, start, end, score, strand, frame, attributes) =
-      (fields(0), fields(1), fields(2), fields(3), fields(4), fields(5), fields(6), fields(7), fields(8))
+    val (seqname, source, featureType, start, end, score, strand, frame) =
+      (fields(0), fields(1), fields(2), fields(3), fields(4), fields(5), fields(6), fields(7))
+    val attributes = if (fields.length == 9) {
+      fields(8)
+    } else {
+      ""
+    }
 
     val f = Feature.newBuilder()
       .setSource(source)
@@ -133,12 +139,17 @@ class GFF3Parser extends FeatureParser {
 
     val fields = line.split("\t")
     // skip empty or invalid lines
-    if (fields.length < 9) {
-      if (log.isDebugEnabled) log.debug("Empty or invalid GFF3 line: {}", line)
+    if (fields.length < 8 || fields.length > 9) {
+      log.warn("Empty or invalid GFF3 line: {}", line)
       return Seq()
     }
-    val (seqid, source, featureType, start, end, score, strand, phase, attributes) =
-      (fields(0), fields(1), fields(2), fields(3), fields(4), fields(5), fields(6), fields(7), fields(8))
+    val (seqid, source, featureType, start, end, score, strand, phase) =
+      (fields(0), fields(1), fields(2), fields(3), fields(4), fields(5), fields(6), fields(7))
+    val attributes = if (fields.length == 9) {
+      fields(8)
+    } else {
+      ""
+    }
 
     val f = Feature.newBuilder()
       .setSource(source)
@@ -169,8 +180,9 @@ class GFF3Parser extends FeatureParser {
  * In lieu of a specification, see:
  * https://samtools.github.io/htsjdk/javadoc/htsjdk/htsjdk/samtools/util/IntervalList.html
  */
-class IntervalListParser extends Serializable {
-  def parse(line: String): (Option[SequenceRecord], Option[Feature]) = {
+class IntervalListParser extends Serializable with Logging {
+  def parse(line: String,
+            stringency: ValidationStringency): (Option[SequenceRecord], Option[Feature]) = {
     val fields = line.split("[ \t]+")
     if (fields.length < 2) {
       (None, None)
@@ -178,9 +190,18 @@ class IntervalListParser extends Serializable {
       if (fields(0).startsWith("@")) {
         if (fields(0).startsWith("@SQ")) {
           val (name, length, url, md5) = {
-            val attrs = fields.drop(1).map(field => field.split(":", 2) match {
-              case Array(key, value) => key -> value
-              case x                 => throw new Exception(s"Expected fields of the form 'key:value' in field $field but got: $x. Line:\n$line")
+            val attrs = fields.drop(1).flatMap(field => field.split(":", 2) match {
+              case Array(key, value) => Some((key -> value))
+              case x => {
+                if (stringency == ValidationStringency.STRICT) {
+                  throw new Exception(s"Expected fields of the form 'key:value' in field $field but got: $x. Line:\n$line")
+                } else {
+                  if (stringency == ValidationStringency.LENIENT) {
+                    log.warn(s"Expected fields of the form 'key:value' in field $field but got: $x. Line:\n$line")
+                  }
+                  None
+                }
+              }
             }).toMap
 
             // Require that all @SQ lines have name, length, url, md5.
@@ -200,17 +221,26 @@ class IntervalListParser extends Serializable {
           (if (fields.length < 5 || fields(4) == "." || fields(4) == "-") {
             (Nil, Map())
           } else {
-            val a = fields(4).split(Array(';', ',')).map(field => field.split('|') match {
+            val a = fields(4).split(Array(';', ',')).flatMap(field => field.split('|') match {
               case Array(key, value) =>
                 key match {
                   case "gn" | "ens" | "vega" | "ccds" =>
-                    (
+                    Some((
                       Some(Dbxref.newBuilder().setDb(key).setAccession(value).build()),
                       None
-                    )
-                  case _ => (None, Some(key -> value))
+                    ))
+                  case _ => Some((None, Some(key -> value)))
                 }
-              case x => throw new Exception(s"Expected fields of the form 'key|value;' but got: $field. Line:\n$line")
+              case x => {
+                if (stringency == ValidationStringency.STRICT) {
+                  throw new Exception(s"Expected fields of the form 'key|value;' but got: $field. Line:\n$line")
+                } else {
+                  if (stringency == ValidationStringency.LENIENT) {
+                    log.warn(s"Expected fields of the form 'key|value;' but got: $field. Line:\n$line")
+                  }
+                  None
+                }
+              }
             })
 
             (a.flatMap(_._1).toList, a.flatMap(_._2).toMap)
@@ -308,7 +338,7 @@ class NarrowPeakParser extends FeatureParser {
 
     // skip empty or invalid lines
     if (fields.length < 3) {
-      if (log.isDebugEnabled) log.debug("Empty or invalid NarrowPeak line: {}", line)
+      log.warn("Empty or invalid NarrowPeak line: {}", line)
       return Seq()
     }
     val fb = Feature.newBuilder()
@@ -321,7 +351,7 @@ class NarrowPeakParser extends FeatureParser {
 
     if (hasColumn(3)) fb.setName(fields(3))
     if (hasColumn(4)) fb.setScore(fields(4).toDouble)
-    if (hasColumn(5)) Features.toStrand(fields(5)).foreach(fb.setStrand(_))
+    if (fields.length > 5) Features.toStrand(fields(5)).foreach(fb.setStrand(_))
 
     val attributes = new ArrayBuffer[(String, String)]()
     if (hasColumn(6)) attributes += ("signalValue" -> fields(6))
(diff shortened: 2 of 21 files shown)


fnothaft:
Resolves #1106 now too. Ready for review/merge.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureRDD.scala
```diff
@@ -95,10 +102,117 @@ object FeatureRDD {
...
+  /**
+   * @param feature Feature to write in IntervalList format.
+   * @return Feature as a one line interval list string.
+   */
+  private[rdd] def toInterval(feature: Feature): String = {
+    val sequenceName = feature.getContigName
+    val start = feature.getStart + 1 // IntervalList ranges are 1-based
+    val end = feature.getEnd // IntervalList ranges are closed
+    val strand = Features.asString(feature.getStrand, emptyUnknown = false)
+    val attributes = Features.gatherAttributes(feature)
```
heuermh: This goes back to shoving all attributes into the IntervalList format name column, right?  I saw that as an abuse to the format and generated a name with reasonable defaults, see https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/Features.scala#L207.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/variation/VariantContextRDD.scala
```diff
@@ -152,17 +139,75 @@ case class VariantContextRDD(rdd: RDD[VariantContext],
...
+      writableVCs.saveAsNewAPIHadoopFile(
+        filePath,
+        classOf[LongWritable],
+        classOf[VariantContextWritable],
+        classOf[ADAMVCFOutputFormat[LongWritable]],
+        conf
+      )
+
+      // remove header file
+      //fs.delete(headPath, true)
```
heuermh: Why is this commented out?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/Features.scala
```diff
@@ -47,15 +47,23 @@ private[features] object Features {
...
    */
-  def asString(strand: Strand): String = {
+  def asString(strand: Strand, emptyUnknown: Boolean = false): String = {
     strand match {
       case Strand.FORWARD     => "+"
       case Strand.REVERSE     => "-"
       case Strand.INDEPENDENT => "."
       case Strand.UNKNOWN     => "?"
-      case _                  => ""
+      case _ => {
```
heuermh: So `""` is read in as `null` and then written back out as `""` if `emptyUnknown`, right?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/variation/VariantContextRDD.scala
fnothaft: Ah, thanks for catching this. I had commented it out while debugging something.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureRDD.scala
fnothaft: I'm not 100% here—I'm a bit unfamiliar with the IntervalList format, even after this PR—but the implementation here passes round trip (with the exception of some ordering bits) with the interval list file in our test resources directory. For me to understand the difference a bit better, could you perhaps elucidate how this code would change the output? E.g., how would the resultant interval list line look different?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/Features.scala
fnothaft: Correct.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
```diff
@@ -308,7 +338,7 @@ class NarrowPeakParser extends FeatureParser {
 
     // skip empty or invalid lines
     if (fields.length < 3) {
-      if (log.isDebugEnabled) log.debug("Empty or invalid NarrowPeak line: {}", line)
+      log.warn("Empty or invalid NarrowPeak line: {}", line)
```
heuermh: Blank lines (and browser & track header lines) are valid in BED and NarrowPeak files, and encouraged between different tracks in the same file, so `log.warn` is probably too severe here and above.  See e.g. https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/test/resources/dvl1.200.bed#L34

fnothaft: Ah yeah, I wanted to make another pass back over this code and attach `ValidationStringency` flags and etc to this. How about I:

* Remove logging on blank/header lines
* Add a validation stringency flag:
  * Throw exception on malformed lines with strict stringency
  * log.warn on lenient
  * No logging on silent

adam-cli/src/main/scala/org/bdgenomics/adam/cli/TransformFeatures.scala
```diff
@@ -22,32 +22,39 @@ import org.bdgenomics.adam.rdd.ADAMContext._
 import org.bdgenomics.utils.cli._
 import org.kohsuke.args4j.{ Argument, Option ⇒ Args4jOption }
 
-object Features2ADAM extends BDGCommandCompanion {
-  val commandName = "features2adam"
-  val commandDescription = "Convert a file with sequence features into corresponding ADAM format"
+object TransformFeatures extends BDGCommandCompanion {
+  val commandName = "transformFeatures"
+  val commandDescription = "Convert a file with sequence features into corresponding ADAM format and vice versa"
```
heuermh: If I have this correctly, on either `INPUT` or `OUTPUT` if the file extension is not recognized as a plain-text feature file, then it is assumed to be parquet.  That way this one command can convert both ways.  The `usage` strings below should be updated.

fnothaft: Correct. I'll update the usage strings.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
heuermh: That sounds good, sorry I didn't get it in before.

fnothaft: No sweat!

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureRDD.scala
heuermh: The javadoc for IntervalList (all the spec I'm aware of) describes the last column as "Interval name (an, ideally unique, name for the interval)".  I haven't seen many of these files in the wild, so I don't know if `|` delimited attributes in the Interval name column is a reasonable thing to do.

Thus I found the previous code somewhat unsavory and arbitrary (as some hard coded keys are shoved into dbxrefs), see https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala#L199.

I hadn't got to round tripping; I guess I was planning to shove the Interval name column into `Feature.name` field when reading IntervalList format and use `Features.nameOf` when writing.  And then find a different file for the unit test.

heuermh:
Looks pretty good.  Will try it out on various files tomorrow (and also hopefully re-remember how to use our internal cluster).

jpdna:
I attempted a round-trip BAM file conversion like:
```
adam-submit transform testb.bam testb.adam
adam-submit transform testb.adam testb.roundtrip.bam -single
```
and although the resulting `testb.roundtrip.bam ` is 104MB ( the original was 101MB) running `samtools flagstat testb.roundtrip.bam` reports 0 counts for all stats, while there were 984817 reads reported by flagstat on the orignal file.  

Also `samtools view -h  -o testb.roundtrip.conver2sam.sam testb.roundtrip.bam` will not produce any reads data in an output sam file, only the header.  So although no corruption error is logged by samtools, the file must be corrupt from the point of view of samtools.  

When not using -single
`adam-submit transform testb.adam testb.roundtrip.bam` part-r-0000* parquet files appear to be correctly generated.

If you have trouble replicating - let me know and I'll get you the file.

fnothaft:
Thanks for looking at this @jpdna! By any chance, have you tried running any of the Picard metrics tools on the file? Our round trip SAM/BAM unit tests are passing, so I believe that htsjdk is able to read the file fine. I wonder if we're writing the BAM header in a way so that htsjdk and htslib read the file differently? I'll take a look tomorrow.

jpdna:
Interestingly - Picard seems to not like even the ORIGINAL test input file
```
java -jar ~/Apps/picard/release_2.2.2/picard-tools-2.2.2/picard.jar SamToFastq I=testb.bam F=out.fastq SECOND_END_FASTQ=out_second_end.fastq
[Fri Aug 19 00:02:08 EDT 2016] picard.sam.SamToFastq INPUT=testb.bam FASTQ=out.fastq SECOND_END_FASTQ=out_second_end.fastq    OUTPUT_PER_RG=false RG_TAG=PU RE_REVERSE=true INTERLEAVE=false INCLUDE_NON_PF_READS=false CLIPPING_MIN_LENGTH=0 READ1_TRIM=0 READ2_TRIM=0 INCLUDE_NON_PRIMARY_ALIGNMENTS=false VERBOSITY=INFO QUIET=false VALIDATION_STRINGENCY=STRICT COMPRESSION_LEVEL=5 MAX_RECORDS_IN_RAM=500000 CREATE_INDEX=false CREATE_MD5_FILE=false GA4GH_CLIENT_SECRETS=client_secrets.json
[Fri Aug 19 00:02:08 EDT 2016] Executing as jp@jp1 on Linux 4.3.5-300.fc23.x86_64 amd64; Java HotSpot(TM) 64-Bit Server VM 1.8.0_71-b15; Picard version: 2.2.2(20d49152d0840a960fcb97df76dbaca260b39244_1461168806) IntelDeflater
[Fri Aug 19 00:02:09 EDT 2016] picard.sam.SamToFastq done. Elapsed time: 0.01 minutes.
Runtime.totalMemory()=2024275968
To get help, see http://broadinstitute.github.io/picard/index.html#GettingHelp
Exception in thread "main" htsjdk.samtools.SAMFormatException: SAM validation error: ERROR: Record 40060, Read name ERR001301.6170436, bin field of BAM record does not equal value computed based on alignment start and end, and length of sequence to which read is aligned
	at htsjdk.samtools.SAMUtils.processValidationErrors(SAMUtils.java:441)
	at htsjdk.samtools.BAMFileReader$BAMFileIterator.advance(BAMFileReader.java:652)
	at htsjdk.samtools.BAMFileReader$BAMFileIterator.next(BAMFileReader.java:637)
	at htsjdk.samtools.BAMFileReader$BAMFileIterator.next(BAMFileReader.java:607)
	at htsjdk.samtools.SamReader$AssertingIterator.next(SamReader.java:545)
	at htsjdk.samtools.SamReader$AssertingIterator.next(SamReader.java:519)
	at picard.sam.SamToFastq.doWork(SamToFastq.java:174)
	at picard.cmdline.CommandLineProgram.instanceMain(CommandLineProgram.java:209)
	at picard.cmdline.PicardCommandLine.instanceMain(PicardCommandLine.java:95)
	at picard.cmdline.PicardCommandLine.main(PicardCommandLine.java:105)
```
Here is the original bam file I am feeding into this round-trip test:
https://drive.google.com/file/d/0B6jh69UgixwpVVJwdE1qTXhjcEk/view?usp=sharing

I'd write this off to just an incorrect BAM file that samtools is too permissive on, but the test file was derived from the 1000 genomes files available here:
ftp://ftp.1000genomes.ebi.ac.uk/vol1/ftp/technical/pilot2_high_cov_GRCh37_bams/data/NA12878/alignment/NA12878.chrom22.ILLUMINA.bwa.CEU.high_coverage.20100311.bam

I also just tried that whole 1000 genomes with the Picard command above, and Picard produces the same error, but samtools flagstat and view seems to work fine on it.   

So - it  would be interesting to know exactly what is up with these 1000 genomes files that makes htsjdk unable to read them ( related to "bin" field I guess)  - and thus ADAM can't either, and see if this jdk error message can at least be passed on more clearly in ADAM rather than just dropping the reads -  if indeed it is legitimate to refuse to process these.

jpdna:
I googled a bit and find that setting `VALIDATION_STRINGENCY=LENIENT` when running picard allows picard to process this 1000 genomes file file.
```
 java -jar ~/Apps/picard/release_2.2.2/picard-tools-2.2.2/picard.jar SamToFastq I=NA12878.chrom22.ILLUMINA.bwa.CEU.high_coverage.20100311.bam F=out.fastq SECOND_END_FASTQ=out_second_end.fastq VALIDATION_STRINGENCY=LENIENT
```
as discussed here : http://gatkforums.broadinstitute.org/gatk/discussion/4290/sam-bin-field-error-for-the-gatk-run

So perhaps we need to see if we can allow the same "lenient" mode in ADAM?

fnothaft:
> So perhaps we need to see if we can allow the same "lenient" mode in ADAM?

I don't think that applies for ADAM, as we're not using the index.

jpdna:
The ADAM parquet file `testb.adam` examined with `adam-submit flagstat testb.adam` does appear fine and agrees with`samtools flagstat` of the original test.bam.  

Also looking further at the per partition "bam" files named part-r-0000* produced by `adam-submit transform testb.adam testb.roundtrip.NOSINGLE.bam` without `-single` are reporting reasonable samtools flagstat count of reads, though the 
`[W::bam_hdr_read] EOF marker is absent. The input is probably truncated.` 
 warning is logged by samtools.

jpdna:
Trying to output a `-single` sam file (rather than bam) works fine as:
```adam-submit transform testb.adam testb.rountrip.sam -single```
and samtools is then happy to operate on that sam file, including converting it to bam.

Also - I did same test and got same result of samtools not being to read the round tripped bam when using the `unsorted.bam` file we use in our unit tests.

```
adam-submit transform unsorted.sam unsorted.adam
../adam/bin/adam-submit transform unsorted.adam unsorted.roundtrip.bam -single

 
samtools flagstat unsorted.roundtrip.bam
0 + 0 in total (QC-passed reads + QC-failed reads)
0 + 0 secondary
0 + 0 supplementary
0 + 0 duplicates
0 + 0 mapped (N/A : N/A)
0 + 0 paired in sequencing
0 + 0 read1
0 + 0 read2
0 + 0 properly paired (N/A : N/A)
0 + 0 with itself and mate mapped
0 + 0 singletons (N/A : N/A)
0 + 0 with mate mapped to a different chr
0 + 0 with mate mapped to a different chr (mapQ>=5)
```

**Comment by fnothaft** [comment]
<3 htslib/htsjdk format disagreements

@jpdna thanks for testing this on samtools. I'll go ahead and debug this further. As an aside, we should add an integration test that reads the files in from samtools, or some other tool that uses htslib. I know that htslib and htsjdk also disagree on VCF, but this one is somewhat shocking. If we can root cause the issue, we should report it upstream.

**Comment by jpdna** [comment]
I'd be happy to have a go at making such integration test with samtools, I'll open an issue.  We can then test the fix with it when ready.

**Comment by fnothaft** [comment]
Awesome! Thanks @jpdna!

**Comment by fnothaft** [comment]
I'm looking at this now. It looks like we write the header completely correctly. I'm not sure what exactly is going on, so I'm unrolling some of the code so that we can write uncompressed BAM (at least for the header) so that I can look at a hexdump to see what's going wrong.

Otherwise, we've got some bad stuff going on in the `AlignmentRecordRDD` round trip tests, so I'm going to patch this up in this PR as well.

**Comment by jpdna** [comment]
Just pinging on this @fnothaft - I think you are continuing to  dig through the hexdumps?

**Comment by fnothaft** [comment]
Thanks for the ping @jpdna ; let me get back on this...

**Comment by fnothaft** [comment]
Just rebased and pushed a fix for the samtools issue. It is isolated in https://github.com/bigdatagenomics/adam/pull/1104/commits/9bd6785481a4c99f6e103f5bdb78f2132a4d5693, I'll squash that down into one of the other commits later. I've tested this locally on a small file with samtools. @jpdna can you check this on your side as well?

**Comment by fnothaft** [comment]
Ping for review/merge?

**Comment by heuermh** [comment]
I'm not sure we resolved the IntervalList format name column discussion above, and a few of the other review comments still need addressing.

On my end, I want to run through some tests on the cluster.

**Comment by fnothaft** [comment]
Ah, sorry! I did a quick pass over the PR earlier and thought I'd addressed the comments. Evidently, I read way too fast. I'll make a pass over those tomorrow.

File: adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureRDD.scala (hunk 1 above)
Comment (fnothaft) [inline adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureRDD.scala]: Just wanted to get back to this, the reason I've shoved records in here is that this is the format we got in the interval list file that described the exome panel. I don't know what the "best" approach is, but I don't think interval list is a rigorously specified format, so we're somewhat making things up as we go here.

Comment (heuermh) [inline adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureRDD.scala]: Yeah, without seeing more examples of these in the wild, I'm not sure if the name column should be mapped to `Feature.name` or to `Feature.attributes`.

**Comment by fnothaft** [comment]
@heuermh just rebased and addressed your review comments. Let me know your thoughts.

File: adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala [hunk 6]
```diff
@@ -328,17 +327,22 @@ class BEDParser extends FeatureParser {
...
     def hasColumn(n: Int): Boolean = {
       fields.length > n && fields(n) != "."
     }
 
     // skip empty or invalid lines
     if (fields.length < 3) {
-      log.warn("Empty or invalid NarrowPeak line: {}", line)
+      if (stringency == ValidationStringency.STRICT) {
+        throw new IllegalArgumentException("Empty or invalid NarrowPeak line: %s".format(line))
+      } else if (stringency == ValidationStringency.STRICT) {
```
heuermh: this should be `LENIENT`

adam-cli/src/main/scala/org/bdgenomics/adam/cli/TransformFeatures.scala
heuermh: LGTM

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
fnothaft: Just patched, thanks for catching.

fnothaft:
@heuermh if this looks good, can you merge? Merging this will simplify #1145.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
```diff
@@ -169,18 +179,28 @@ class GFF3Parser extends FeatureParser {
  * In lieu of a specification, see:
  * https://samtools.github.io/htsjdk/javadoc/htsjdk/htsjdk/samtools/util/IntervalList.html
  */
-class IntervalListParser extends Serializable {
-  def parse(line: String): (Option[SequenceRecord], Option[Feature]) = {
+private[rdd] class IntervalListParser extends Serializable with Logging {
```
heuermh: Are we reusing the SAM header code for IntervalList format?  If so I was thinking we could remove all the `startsWith("@")` stuff here.

This method should only need to handle the five tab-delimited columns (Sequence name, Start position (1-based), End position (1-based, end inclusive), Strand (either + or -), Interval name (an, ideally unique, name for the interval).

heuermh:
Reviewing meow

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
fnothaft: We're reusing the SAM header _writing_ code, not the SAM header _reading_ code. We might be able to reuse that code, but that's less straightforward. Perhaps we can punt that to a later ticket?

fnothaft: Essentially, I just don't have the bandwidth to take that on now.

heuermh: ok, how about just getting rid of the arbitrary dbxref key stuff at

https://github.com/fnothaft/adam/blob/f21ead579471ab7854aa519ee7efc884f2f6f38b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala#L226

then?  Don't set dbxrefs at all, just attributes.

fnothaft: I assume this was meant to be on a different line? That change tracks all the way back to https://github.com/bigdatagenomics/adam/pull/686. I'd want to get @ryan-williams' thoughts on that code before we yanked it out or changed it. I think we're getting into a debate that's somewhat orthogonal to this PR, which is _what_ exactly the `.interval_list` format is. I've just added code to said parser so that it passes round trip tests. If we'd like to change the parser, I think we should open an issue and discuss that there, and not block this PR on it. That's just my opinion though. Perhaps let's discuss on the call tomorrow?

heuermh: Let's create an issue to 1) look into using SAM header code when reading IntervalList format and 2) clean up this dbxref key stuff

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureRDD.scala
heuermh: ok with me for now, will resolve in future issue

heuermh:
Created a new issue #1152 to resolve IntervalList discussions later.

Is this pull request ready to merge then?

fnothaft:
Just rebased. Should be good for a merge once tests pass.

heuermh:
Merged via multiple commits.

Thank you, @fnothaft!

PR #1108, opened 2016-08-09
Modified CalculateDepth to calcuate coverage from alignment files

akmorrow13:
* Calculates coverage from reads files on command line
* Creates CoverageRDD from FeatureRDD

### DIFF
--- a/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Reads2Coverage.scala
+++ b/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Reads2Coverage.scala
@@ -1,0 +1,86 @@
+/**
+ * Licensed to Big Data Genomics (BDG) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The BDG licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.bdgenomics.adam.cli
+
+import org.apache.spark.SparkContext
+import org.bdgenomics.adam.projections.AlignmentRecordField._
+import org.bdgenomics.adam.projections.Projection
+import org.bdgenomics.adam.rdd.ADAMContext._
+import org.bdgenomics.adam.rdd.features.CoverageRDD
+import org.bdgenomics.adam.rdd.read.{ AlignedReadRDD, AlignmentRecordRDD }
+import org.bdgenomics.utils.cli._
+import org.kohsuke.args4j.{ Argument, Option => Args4jOption }
+
+/**
+ * Reads2Coverage (accessible as the command 'reads2coverage' through the CLI) takes two arguments,
+ * an INPUT and OUTPUT, and calculates the number of reads from INPUT at every location in
+ * the file. Optional arguments are only_negative_strands, only_positive_strands and collapse.
+ * only_negative_strands and only_positive_strands save coverage computed from only negative and positive strands,
+ * respectively. Collapse specifies whether saved coverage should merge neighboring coverage with the same counts
+ * to one record.
+ */
+object Reads2Coverage extends BDGCommandCompanion {
+  val commandName: String = "reads2coverage"
+  val commandDescription: String = "Calculate the coverage from a given ADAM file"
+
+  def apply(cmdLine: Array[String]): BDGCommand = {
+    new Reads2Coverage(Args4j[Reads2CoverageArgs](cmdLine))
+  }
+}
+
+class Reads2CoverageArgs extends Args4jBase with ParquetArgs {
+  @Argument(required = true, metaVar = "INPUT", usage = "The reads file to use to calculate depths", index = 0)
+  var inputPath: String = null
+  @Argument(required = true, metaVar = "OUTPUT", usage = "Location to write the coverage data in ADAM/Parquet format", index = 1)
+  var outputPath: String = null
+  @Args4jOption(required = false, name = "-collapse", usage = "Collapses neighboring coverage records " +
+    "of equal counts into the same record")
+  var collapse: Boolean = true
+  @Args4jOption(required = false, name = "-only_negative_strands", usage = "Compute coverage for negative strands")
+  var onlyNegativeStrands: Boolean = false
+  @Args4jOption(required = false, name = "-only_positive_strands", usage = "Compute coverage for positive strands")
+  var onlyPositiveStrands: Boolean = false
+}
+
+class Reads2Coverage(protected val args: Reads2CoverageArgs) extends BDGSparkCommand[Reads2CoverageArgs] {
+  val companion: BDGCommandCompanion = Reads2Coverage
+
+  def run(sc: SparkContext): Unit = {
+
+    val proj = Projection(contigName, start, end, cigar)
+
+    // If saving strand specific coverage, require that only one direction is specified
+    require(!(args.onlyNegativeStrands && args.onlyPositiveStrands),
+      "Cannot compute coverage for both negative and positive strands separately")
+
+    // load reads
+    val readsRdd: AlignmentRecordRDD = sc.loadAlignments(args.inputPath, projection = Some(proj))
+
+    val finalReads = if (args.onlyNegativeStrands && !args.onlyPositiveStrands) {
+      readsRdd.transform(rdd => rdd.filter(_.getReadNegativeStrand))
+    } else if (!args.onlyNegativeStrands && args.onlyPositiveStrands) {
+      readsRdd.transform(rdd => rdd.filter(!_.getReadNegativeStrand))
+    } else {
+      readsRdd
+    }
+
+    finalReads.toCoverage(args.collapse)
+      .save(args.outputPath)
+  }
+}
+
--- a/adam-cli/src/test/scala/org/bdgenomics/adam/cli/Reads2CoverageSuite.scala
+++ b/adam-cli/src/test/scala/org/bdgenomics/adam/cli/Reads2CoverageSuite.scala
@@ -1,0 +1,38 @@
+/**
+ * Licensed to Big Data Genomics (BDG) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The BDG licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.bdgenomics.adam.cli
+
+import org.bdgenomics.adam.rdd.ADAMContext._
+import org.bdgenomics.adam.util.ADAMFunSuite
+import org.bdgenomics.utils.cli.Args4j
+
+class Reads2CoverageSuite extends ADAMFunSuite {
+
+  sparkTest("correctly calculates coverage from small sam file") {
+    val inputPath = copyResource("artificial.sam")
+    val outputPath = tmpFile("coverage.adam")
+
+    val args: Array[String] = Array(inputPath, outputPath)
+    new Reads2Coverage(Args4j[Reads2CoverageArgs](args)).run(sc)
+    val coverage = sc.loadCoverage(outputPath)
+
+    val pointCoverage = coverage.flatten.rdd.filter(_.start == 30).first
+    assert(pointCoverage.count == 5)
+  }
+}
+
--- a/adam-core/src/main/scala/org/bdgenomics/adam/models/Coverage.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/models/Coverage.scala
@@ -1,0 +1,86 @@
+/**
+ * Licensed to Big Data Genomics (BDG) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The BDG licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.bdgenomics.adam.models
+
+import org.apache.spark.rdd.RDD
+import org.bdgenomics.formats.avro.Feature
+
+/**
+ * Converts from avro Feature to Coverage.
+ */
+object Coverage {
+
+  /**
+   * Creates Coverage from ReferenceRegion and coverage count in that ReferenceRegion.
+   *
+   * @param region ReferenceRegion in which Coverage spans
+   * @param count Coverage count for each base pair in region
+   * @return Coverage spanning the specified ReferenceRegion
+   */
+  private[adam] def apply(region: ReferenceRegion, count: Double): Coverage = {
+    Coverage(region.referenceName, region.start, region.end, count)
+  }
+
+  /**
+   * Creates Coverage from Feature, extracting region information and feature score for coverage.
+   *
+   * @param feature Feature to create coverage from
+   * @return Coverage spanning the specified feature
+   */
+  private[adam] def apply(feature: Feature): Coverage = {
+    Coverage(feature.getContigName, feature.getStart, feature.getEnd, feature.getScore)
+  }
+
+  /**
+   * Creates an RDD of Coverage from RDD of Features.
+   *
+   * @param rdd RDD of Features to extract Coverage from
+   * @return RDD of Coverage spanning all features in rdd
+   */
+  private[adam] def apply(rdd: RDD[Feature]): RDD[Coverage] = {
+    rdd.map(f => Coverage(f))
+  }
+}
+
+/**
+ * Coverage record for CoverageRDD.
+ * Contains Region indexed by contig name, start and end, as well as count of coverage at
+ * each base pair in that region.
+ *
+ * @param contigName Specifies chromosomal location of coverage
+ * @param start Specifies start position of coverage
+ * @param end  Specifies end position of coverage
+ * @param count Specifies count of coverage at location
+ */
+case class Coverage(contigName: String, start: Long, end: Long, count: Double) {
+
+  /**
+   * Converts Coverage to Feature, setting Coverage count in the score attribute.
+   *
+   * @return Feature built from Coverage
+   */
+  def toFeature: Feature = {
+    val fb = Feature.newBuilder()
+    fb.setContigName(contigName)
+    fb.setStart(start)
+    fb.setEnd(end)
+    fb.setScore(count)
+    fb.build()
+  }
+}
+
--- a/adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
@@ -150,8 +150,24 @@ object ReferenceRegion {
     }
   }
 
+  /**
+   * Builds a reference region for a feature.
+   *
+   * @param feature Feature to extract ReferenceRegion from
+   * @return Extracted ReferenceRegion
+   */
   def apply(feature: Feature): ReferenceRegion = {
     new ReferenceRegion(feature.getContigName, feature.getStart, feature.getEnd)
+  }
+
+  /**
+   * Builds a reference region for a coverage site.
+   *
+   * @param coverage Coverage to extract ReferenceRegion from
+   * @return Extracted ReferenceRegion
+   */
+  def apply(coverage: Coverage): ReferenceRegion = {
+    new ReferenceRegion(coverage.contigName, coverage.start, coverage.end)
   }
 }
 
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
@@ -17,12 +17,12 @@
  */
 package org.bdgenomics.adam.rdd
 
-import collection.JavaConverters._
 import java.io.{ File, FileNotFoundException, InputStream }
 import java.util.regex.Pattern
+
 import com.google.common.collect.ImmutableList
-import htsjdk.samtools.{ SAMFileHeader, ValidationStringency }
 import htsjdk.samtools.util.{ Interval, Locatable }
+import htsjdk.samtools.{ SAMFileHeader, ValidationStringency }
 import htsjdk.variant.vcf.VCFHeader
 import org.apache.avro.Schema
 import org.apache.avro.file.DataFileStream
@@ -35,33 +35,35 @@ import org.apache.parquet.avro.{ AvroParquetInputFormat, AvroReadSupport }
 import org.apache.parquet.filter2.predicate.FilterPredicate
 import org.apache.parquet.hadoop.ParquetInputFormat
 import org.apache.parquet.hadoop.util.ContextUtil
+import org.apache.spark.SparkContext
 import org.apache.spark.rdd.MetricsContext._
 import org.apache.spark.rdd.RDD
-import org.apache.spark.SparkContext
 import org.bdgenomics.adam.converters._
 import org.bdgenomics.adam.instrumentation.Timers._
 import org.bdgenomics.adam.io._
 import org.bdgenomics.adam.models._
-import org.bdgenomics.adam.projections.{ AlignmentRecordField, NucleotideContigFragmentField, Projection }
+import org.bdgenomics.adam.projections.{
+  AlignmentRecordField,
+  FeatureField,
+  NucleotideContigFragmentField,
+  Projection
+}
 import org.bdgenomics.adam.rdd.contig.NucleotideContigFragmentRDD
 import org.bdgenomics.adam.rdd.features._
 import org.bdgenomics.adam.rdd.fragment.FragmentRDD
-import org.bdgenomics.adam.rdd.read.{
-  AlignedReadRDD,
-  AlignmentRecordRDD,
-  UnalignedReadRDD
-}
+import org.bdgenomics.adam.rdd.read.{ AlignedReadRDD, AlignmentRecordRDD, UnalignedReadRDD }
 import org.bdgenomics.adam.rdd.variation._
 import org.bdgenomics.adam.rich.RichAlignmentRecord
-import org.bdgenomics.adam.util.{ TwoBitFile, ReferenceContigMap, ReferenceFile }
+import org.bdgenomics.adam.util.{ ReferenceContigMap, ReferenceFile, TwoBitFile }
 import org.bdgenomics.formats.avro._
 import org.bdgenomics.utils.instrumentation.Metrics
 import org.bdgenomics.utils.io.LocalFileByteAccess
-import org.bdgenomics.utils.misc.HadoopUtil
-import org.bdgenomics.utils.misc.Logging
+import org.bdgenomics.utils.misc.{ HadoopUtil, Logging }
 import org.seqdoop.hadoop_bam._
 import org.seqdoop.hadoop_bam.util._
+
 import scala.collection.JavaConversions._
+import scala.collection.JavaConverters._
 import scala.collection.Map
 import scala.reflect.ClassTag
 
@@ -757,6 +759,17 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
     FragmentRDD.fromRdd(records.map(fastqRecordConverter.convertFragment))
   }
 
+  /**
+   * Loads Parquet file of Features to a CoverageRDD.
+   * Coverage is stored in the score attribute of Feature.
+   * @param filePath File path to load coverage from
+   * @return CoverageRDD containing an RDD of Coverage
+   */
+  def loadCoverage(filePath: String): CoverageRDD = {
+    val proj = Projection(FeatureField.contigName, FeatureField.start, FeatureField.end, FeatureField.score)
+    loadFeatures(filePath, projection = Some(proj)).toCoverage
+  }
+
   def loadGff3(filePath: String, minPartitions: Option[Int] = None): FeatureRDD = {
     val records = sc.textFile(filePath, minPartitions.getOrElse(sc.defaultParallelism)).flatMap(new GFF3Parser().parse)
     if (Metrics.isRecording) records.instrument() else records
@@ -999,7 +1012,6 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
     } else if (filePath.endsWith(".fa") ||
       filePath.endsWith(".fasta")) {
       log.info(s"Loading $filePath as FASTA and converting to AlignmentRecords. Projection is ignored.")
-      import ADAMContext._
       UnalignedReadRDD(loadFasta(filePath, fragmentLength = 10000).toReads,
         RecordGroupDictionary.empty)
     } else if (filePath.endsWith("contig.adam")) {
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
@@ -1,0 +1,221 @@
+/**
+ * Licensed to Big Data Genomics (BDG) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The BDG licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.bdgenomics.adam.rdd.features
+
+import org.apache.spark.rdd.RDD
+import org.bdgenomics.adam.models.{
+  Coverage,
+  ReferenceRegion,
+  SequenceDictionary,
+  SequenceRecord
+}
+import org.bdgenomics.adam.rdd.GenomicRDD
+
+import scala.annotation.tailrec
+
+/**
+ * An RDD containing Coverage.
+ *
+ * @param rdd Coverage
+ * @param sequences A dictionary describing the reference genome.
+ */
+case class CoverageRDD(rdd: RDD[Coverage],
+                       sequences: SequenceDictionary) extends GenomicRDD[Coverage, CoverageRDD] {
+
+  /**
+   * Saves coverage as feature file.
+   *
+   * @see FeatureRDD.save
+   *
+   * Supported file formats include bed, narrowPeak and parquet. Coverage is saved
+   * as a feature where coverage is stored in score attribute.
+   *   val chrom    = feature.getContigName
+   *   val start    = feature.getStart
+   *   val end      = feature.getEnd
+   *   val name     = Features.nameOf(feature)
+   *   val coverage = feature.getScore
+   *
+   * @param filePath The location to write the output.
+   */
+  def save(filePath: java.lang.String) = {
+    this.toFeatureRDD.save(filePath)
+  }
+
+  /**
+   * Merges adjacent ReferenceRegions with the same coverage value.
+   * This reduces the loss of coverage information while reducing the number of records in the RDD.
+   * For example, adjacent records Coverage("chr1", 1, 10, 3.0) and Coverage("chr1", 10, 20, 3.0)
+   * would be merged into one record Coverage("chr1", 1, 20, 3.0).
+   *
+   * @return merged tuples of adjacent ReferenceRegions and coverage.
+   */
+  def collapse: CoverageRDD = {
+    val newRDD: RDD[Coverage] = rdd
+      .mapPartitions(iter => {
+        if (iter.hasNext) {
+          val first = iter.next
+          collapse(iter, first, List.empty)
+        } else iter
+      })
+
+    transform(rdd => newRDD)
+  }
+
+  /**
+   * Tail recursion for merging adjacent ReferenceRegions with the same value.
+   *
+   * @param iter partition iterator of ReferenceRegion and coverage values.
+   * @param lastCoverage the last coverage from a sorted Iterator that has been considered to merge.
+   * @param condensed Condensed iterator of iter with adjacent regions with the same value merged.
+   * @return merged tuples of adjacent ReferenceRegions and coverage.
+   */
+  @tailrec private def collapse(iter: Iterator[Coverage],
+                                lastCoverage: Coverage,
+                                condensed: List[Coverage]): Iterator[Coverage] = {
+    if (!iter.hasNext) {
+      // if lastCoverage has not yet been added, add to condensed
+      val nextCondensed =
+        if (condensed.map(r => ReferenceRegion(r)).filter(_.overlaps(ReferenceRegion(lastCoverage))).isEmpty) {
+          lastCoverage :: condensed
+        } else {
+          condensed
+        }
+      nextCondensed.toIterator
+    } else {
+      val cov = iter.next
+      val rr = ReferenceRegion(cov)
+      val lastRegion = ReferenceRegion(lastCoverage)
+      val (nextCoverage, nextCondensed) =
+        if (rr.isAdjacent(lastRegion) && lastCoverage.count == cov.count) {
+          (Coverage(rr.merge(lastRegion), lastCoverage.count), condensed)
+        } else {
+          (cov, lastCoverage :: condensed)
+        }
+      collapse(iter, nextCoverage, nextCondensed)
+    }
+  }
+
+  /**
+   * Converts CoverageRDD to FeatureRDD.
+   *
+   * @return Returns a FeatureRDD from CoverageRDD.
+   */
+  def toFeatureRDD: FeatureRDD = {
+    val featureRdd = rdd.map(_.toFeature)
+    FeatureRDD(featureRdd, sequences)
+  }
+
+  /**
+   * Gets coverage overlapping specified ReferenceRegion.
+   * For large ReferenceRegions, base pairs per bin (bpPerBin) can be specified to bin together ReferenceRegions of
+   * equal size. The coverage of each bin is coverage of the first base pair in that bin.
+   *
+   * @param bpPerBin base pairs per bin, number of bases to combine to one bin.
+   * @return RDD of Coverage Records.
+   */
+  def coverage(bpPerBin: Int = 1): CoverageRDD = {
+
+    val flattened = flatten()
+
+    if (bpPerBin == 1) {
+      flattened // no binning, return raw results
+    } else {
+      // subtract region.start to shift mod to start of ReferenceRegion
+      val newRDD = flattened.rdd.filter(r => r.start % bpPerBin == 0)
+      flattened.transform(rdd => newRDD)
+    }
+  }
+
+  /**
+   * Gets coverage overlapping specified ReferenceRegion. For large ReferenceRegions,
+   * base pairs per bin (bpPerBin) can be specified to bin together ReferenceRegions of
+   * equal size. The coverage of each bin is the mean coverage over all base pairs in that bin.
+   *
+   * @param bpPerBin base pairs per bin, number of bases to combine to one bin.
+   * @return RDD of Coverage Records.
+   */
+  def aggregatedCoverage(bpPerBin: Int = 1): CoverageRDD = {
+
+    val flattened = flatten()
+
+    def reduceFn(a: (Double, Int), b: (Double, Int)): (Double, Int) = {
+      (a._1 + b._1, a._2 + b._2)
+    }
+
+    if (bpPerBin == 1) {
+      flattened // no binning, return raw results
+    } else {
+      val newRDD = flattened.rdd
+        .keyBy(r => {
+          // key coverage by binning start site mod bpPerbin
+          // subtract region.start to shift mod to start of ReferenceRegion
+          val start = r.start - (r.start % bpPerBin)
+          ReferenceRegion(r.contigName, start, start + bpPerBin)
+        }).mapValues(r => (r.count, 1))
+        .reduceByKey(reduceFn)
+        .map(r => {
+          // compute average coverage in bin
+          Coverage(r._1.referenceName, r._1.start, r._1.end, r._2._1 / r._2._2)
+        })
+      flattened.transform(rdd => newRDD)
+    }
+  }
+
+  /**
+   * Gets sequence of ReferenceRegions from Coverage element.
+   * Since coverage maps directly to a single genomic region, this method will always
+   * return a Seq of exactly one ReferenceRegion.
+   *
+   * @param elem The Coverage to get an underlying region for.
+   * @return Sequence of ReferenceRegions extracted from Coverage.
+   */
+  protected def getReferenceRegions(elem: Coverage): Seq[ReferenceRegion] = {
+    Seq(ReferenceRegion(elem.contigName, elem.start, elem.end))
+  }
+
+  /**
+   * @param newRdd The RDD to replace the underlying RDD with.
+   * @return Returns a new CoverageRDD with the underlying RDD replaced.
+   */
+  protected def replaceRdd(newRdd: RDD[Coverage]): CoverageRDD = {
+    copy(rdd = newRdd)
+  }
+
+  /**
+   * Gets flattened RDD of coverage, with coverage mapped to a ReferenceRegion at each base pair.
+   *
+   * @return CoverageRDD of flattened Coverage records.
+   */
+  def flatten(): CoverageRDD = {
+    transform(rdd => flatMapCoverage(rdd))
+  }
+
+  /**
+   * Flat maps coverage into ReferenceRegion and counts for each base pair.
+   *
+   * @param rdd RDD of Coverage.
+   * @return RDD of flattened Coverage.
+   */
+  private def flatMapCoverage(rdd: RDD[Coverage]): RDD[Coverage] = {
+    rdd.flatMap(r => {
+      val positions = r.start until r.end
+      positions.map(n => Coverage(r.contigName, n, n + 1, r.count))
+    })
+  }
+}
+
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureRDD.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureRDD.scala
@@ -246,6 +246,16 @@ case class FeatureRDD(rdd: RDD[Feature],
   }
 
   /**
+   * Converts the FeatureRDD to a CoverageRDD.
+   *
+   * @return CoverageRDD containing RDD of Coverage.
+   */
+  def toCoverage: CoverageRDD = {
+    val coverageRdd = rdd.map(f => Coverage(f))
+    CoverageRDD(coverageRdd, sequences)
+  }
+
+  /**
    * @param newRdd The RDD to replace the underlying RDD with.
    * @return Returns a new FeatureRDD with the underlying RDD replaced.
    */
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDD.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDD.scala
@@ -47,6 +47,7 @@ import org.bdgenomics.adam.rdd.{
   JavaSaveArgs,
   Unaligned
 }
+import org.bdgenomics.adam.rdd.features.CoverageRDD
 import org.bdgenomics.adam.rdd.read.realignment.RealignIndels
 import org.bdgenomics.adam.rdd.read.recalibration.BaseQualityRecalibration
 import org.bdgenomics.adam.rich.RichAlignmentRecord
@@ -82,6 +83,35 @@ sealed trait AlignmentRecordRDD extends AvroReadGroupGenomicRDD[AlignmentRecord,
     FragmentRDD(groupReadsByFragment().map(_.toFragment),
       sequences,
       recordGroups)
+  }
+
+  /**
+   * Converts this set of reads into a corresponding CoverageRDD.
+   *
+   * @param collapse Determines whether to merge adjacent coverage elements with the same score a single coverage.
+   * @return CoverageRDD containing mapped RDD of Coverage.
+   */
+  def toCoverage(collapse: Boolean = true): CoverageRDD = {
+    val covCounts =
+      rdd.rdd
+        .flatMap(r => {
+          val t: List[Long] = List.range(r.getStart, r.getEnd)
+          t.map(n => (ReferenceRegion(r.getContigName, n, n + 1), 1))
+        }).reduceByKey(_ + _)
+        .cache()
+
+    val coverage = (
+      if (collapse) covCounts.sortByKey()
+      else covCounts
+    ).map(r => Coverage(r._1, r._2.toDouble))
+
+    val coverageRdd =
+      if (collapse) CoverageRDD(coverage, sequences)
+        .collapse
+      else CoverageRDD(coverage, sequences)
+
+    covCounts.unpersist()
+    coverageRdd
   }
 
   /**
--- a/adam-core/src/test/scala/org/bdgenomics/adam/rdd/features/CoverageRDDSuite.scala
+++ b/adam-core/src/test/scala/org/bdgenomics/adam/rdd/features/CoverageRDDSuite.scala
@@ -1,0 +1,102 @@
+/**
+ * Licensed to Big Data Genomics (BDG) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The BDG licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.bdgenomics.adam.rdd.features
+
+import org.bdgenomics.adam.models.{ Coverage, SequenceDictionary, SequenceRecord }
+import org.bdgenomics.adam.rdd.ADAMContext._
+import org.bdgenomics.adam.util.ADAMFunSuite
+import org.bdgenomics.formats.avro.Feature
+
+class CoverageRDDSuite extends ADAMFunSuite {
+
+  val sd = new SequenceDictionary(Vector(SequenceRecord("chr1", 2000L)))
+
+  def generateCoverage(length: Int): Seq[Coverage] = {
+    // generate adjacent regions with coverage
+    var j = 0
+    val coverage =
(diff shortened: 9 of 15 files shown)


adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,85 @@
...
+  }
+}
+
+/**
+ * Coverage Recordfor CoverageRDD
+ * @param referenceName Specifies chromosomal location of coverage
+ * @param position Specifies position of coverage
+ * @param count Specifies count of coverage at location
+ */
+case class Coverage(referenceName: String, position: Long, count: Double)
```
akmorrow13: There is already a Coverage class. See [here](https://github.com/bigdatagenomics/adam/blob/1d0ace90f5f943704908550136e69c44ba357450/adam-core/src/main/scala/org/bdgenomics/adam/rdd/Coverage.scala). We do not seem to use this/ these are very different. Is there anything difference I can name my Coverage?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDD.scala
```diff
@@ -85,6 +86,35 @@ sealed trait AlignmentRecordRDD extends AvroReadGroupGenomicRDD[AlignmentRecord,
...
+   * counted at all overlapping positions
+   */
+  def toCoverage: CoverageRDD = {
+    val coverage: RDD[(ReferenceRegion, Int)] =
+      rdd.rdd
+        .flatMap(r => {
+          val t: List[Long] = List.range(r.getStart, r.getEnd)
+          val m: List[(ReferenceRegion, Int)] = t.map(n => (ReferenceRegion(r.getContigName, n, n + 1), 1))
+          m
+        }).reduceByKey(_ + _).sortByKey()
```
akmorrow13: Currently, this is storing an individual record for each position in the genome. I would like to run length encode this using ReferenceRegion. What would be the most efficient way to group by ReferenceRegion and count to merge adjacent regions?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,85 @@
...
+    featureRDD.save(filePath)
+  }
+
+  /**
+   * Gets coverage overlapping specified ReferenceRegion. For large, ReferenceRegions,
+   * base pairs per bin (bpPerBin) can be specified to bin together ReferenceRegions of
+   * equal size.
+   *
+   * @param region ReferenceRegion to fetch overlapping coverage from
+   * @param bpPerBin base pairs per bin, number of bases to combine to one bin
```
akmorrow13: This function does a mod for bpPerBin instead of aggregating the bin. However, I couldn't think of how to computing the average size of coverage bins without shuffling. Is this ok?

akmorrow13: I have added the option for either aggregation or taking the first element in the bin. There are use cases for both. The use case for no aggregation is visualization, because aggregation provides more information than is required and incurs high overhead. Aggregation of coverage is useful for algorithms that bin coverage by some heuristic.

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Reads2Coverage.scala
```diff
@@ -0,0 +1,95 @@
...
+    new Reads2Coverage(Args4j[Reads2CoverageArgs](cmdLine))
+  }
+}
+
+class Reads2CoverageArgs extends Args4jBase with ParquetArgs {
+  @Argument(required = true, metaVar = "INPUT", usage = "The reads file to use to calculate depths", index = 0)
+  var inputPath: String = null
+  @Argument(required = true, metaVar = "OUTPUT", usage = "Location to write the coverage data in ADAM/Parquet format", index = 1)
+  var outputPath: String = null
+  @Args4jOption(required = false, name = "-negativeStrands", usage = "Compute coverage for negative strands")
```
fnothaft: Is there a big use case for stranded coverage info?

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Reads2Coverage.scala
```diff
@@ -0,0 +1,95 @@
...
+      // save final features
+      val featureRDD: CoverageRDD = readsRDD.toCoverage
+      featureRDD.save(args.outputPath)
+
+    } else {
+
+      // if negative strand, save coverage of only negative strands
+      if (args.negativeStrands) {
+        // count sites for all strands
+        val negativeReadsRDD: AlignmentRecordRDD = AlignedReadRDD(readsRDD.rdd.filter(_.getReadNegativeStrand), readsRDD.sequences, readsRDD.recordGroups)
```
fnothaft: prefer `readsRDD.transform()` instead

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Reads2Coverage.scala
fnothaft: Also add "only" to both of these options.

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Reads2Coverage.scala
```diff
@@ -0,0 +1,95 @@
...
+
+  def run(sc: SparkContext): Unit = {
+
+    val proj = Projection(contigName, start, end, cigar)
+
+    // load reads
+    val readsRDD: AlignmentRecordRDD = sc.loadAlignments(args.inputPath, projection = Some(proj))
+
+    // if strand direction is not specified, save unified coverage
+    if (!args.negativeStrands && !args.positiveStrands) {
```
fnothaft: Mayhaps refactor to:

```
val readsRdd = sc.loadAlignments(args.inputPath, projection = Some(proj))

val finalReads = if (args.negativeStrands && !args.positiveStrands) {
  readsRdd.transform(rdd => rdd.filter(_.getReadNegativeStrand)
} else if (!args.negativeStrands && args.positiveStrands) {
  readsRdd.transform(rdd => rdd.filter(!_.getReadNegativeStrand)
} else {
  readsRdd
}

readsRdd.toCoverage
  .save(args.outputPath)
```

Tis a bit more concise.

fnothaft: Also, in ADAM, we generally prefer lowercase `rdd` in variable names.

adam-cli/src/test/scala/org/bdgenomics/adam/cli/Reads2CoverageSuite.scala
```diff
@@ -0,0 +1,37 @@
...
+    val outputPath = tmpFile("coverage.adam")
+
+    val args: Array[String] = Array(inputPath, outputPath)
+    new Reads2Coverage(Args4j[Reads2CoverageArgs](args)).run(sc)
+    val coverage = sc.loadCoverage(outputPath)
+
+    val pointCoverage = coverage.rdd.filter(_.position == 30).first
+    assert(pointCoverage.count == 5)
+  }
+}
```
fnothaft: Add newline at EOF

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Reads2Coverage.scala
```diff
@@ -0,0 +1,95 @@
...
+        // count sites for all strands
+        val positiveReadsRDD: AlignmentRecordRDD = AlignedReadRDD(readsRDD.rdd.filter(!_.getReadNegativeStrand), readsRDD.sequences, readsRDD.recordGroups)
+
+        // save final features
+        val coverageRDD: CoverageRDD = positiveReadsRDD.toCoverage
+        coverageRDD.save(s"negative_${args.outputPath}")
+      }
+    }
+  }
+}
```
fnothaft: Add newline at EOF

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
fnothaft: I don't think you can do the binning without shuffling, so that's OK to me.

fnothaft: Can you document `aggregatePerBin`? I don't intuitively grok what that parameter does.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,105 @@
...
+  def getCoverage(region: ReferenceRegion, bpPerBin: Int = 1, aggregateBins: Boolean = false): RDD[Coverage] = {
+    val coverage = featureRDD.filterByOverlappingRegion(region)
+
+    val flattenedCoverage = coverage.rdd.flatMap(r => {
+      val positions: List[Long] = List.range(r.getStart, r.getEnd)
+      positions.map(n => Coverage(r.getContigName, n, r.getScore))
+    }).filter(r => r.position >= region.start && r.position < region.end) // filter out positions from flanking regions
+
+    if (bpPerBin == 1) {
+
```
fnothaft: Nit: Extra whitespace

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Reads2Coverage.scala
```diff
@@ -0,0 +1,95 @@
...
+
+class Reads2CoverageArgs extends Args4jBase with ParquetArgs {
+  @Argument(required = true, metaVar = "INPUT", usage = "The reads file to use to calculate depths", index = 0)
+  var inputPath: String = null
+  @Argument(required = true, metaVar = "OUTPUT", usage = "Location to write the coverage data in ADAM/Parquet format", index = 1)
+  var outputPath: String = null
+  @Args4jOption(required = false, name = "-negativeStrands", usage = "Compute coverage for negative strands")
+  var negativeStrands: Boolean = false
+  @Args4jOption(required = false, name = "-positiveStrands", usage = "Compute coverage for positive strands")
```
heuermh: We're using `-positive_strands` (lowercase with underscores) elsewhere

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,105 @@
...
+ *
+ * @param featureRDD FeatureRDD holding records for coverage and corresponding
+ * referenceRegion
+ */
+case class CoverageRDD(featureRDD: FeatureRDD) {
+
+  /**
+   * Java friendly save function. Automatically detects the output format.
+   *
+   * If the filename ends in ".bed", we write a BED file. If the file name ends
```
heuermh: This doc is maybe not so useful, a link to `FeatureRDD.save` might be more appropriate.
Or a table of examples, showing what the coverage features look like in each format.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDD.scala
fnothaft:
```
import scala.annotation.tailrec
@tailrec def merge(iter: Iterator[(ReferenceRegion, Int)],
                                 lastRegion: ReferenceRegion = <nonsense_val>,
                                 lastCoverage: Int = -1,
                                 condensed: List[(ReferenceRegion, Int)]): Iterator[(ReferenceRegion, Int)] = {
  if (!iter.hasNext) {
    condensed.toIterator
  } else {
    val (rr, cov) = iter.next
    val (nextRegion, nextCoverage, nextCondensed) = if (rr.adjacent(lastRegion) && lastCoverage == nextCoverage) {
     (rr.collapseAdjacent(lastRegion), lastCoverage, condensed)
    } else {
      (rr, cov, (lastRegion, lastCoverage) :: condensed)
    }
    merge(iter, nextRegion, nextCoverage, nextCondensed)
  }
}

coverage.mapPartitions(merge)
```

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Reads2Coverage.scala
akmorrow13: The use case for stranded coverage is for featurization of coverage for DNase in many algorithms. These require separate featurization for negative and positive strands.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDD.scala
fnothaft: Also, `.cache()` before `sortByKey()`.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,105 @@
...
+  /**
+   * Gets coverage overlapping specified ReferenceRegion. For large, ReferenceRegions,
+   * base pairs per bin (bpPerBin) can be specified to bin together ReferenceRegions of
+   * equal size.
+   *
+   * @param region ReferenceRegion to fetch overlapping coverage from
+   * @param bpPerBin base pairs per bin, number of bases to combine to one bin
+   * @return RDD of Coverage Records
+   */
+  def getCoverage(region: ReferenceRegion, bpPerBin: Int = 1, aggregateBins: Boolean = false): RDD[Coverage] = {
```
akmorrow13: split into 2 functions

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Reads2Coverage.scala
```diff
@@ -0,0 +1,80 @@
...
+import org.bdgenomics.adam.rdd.ADAMContext._
+import org.bdgenomics.adam.rdd.features.CoverageRDD
+import org.bdgenomics.adam.rdd.read.{ AlignedReadRDD, AlignmentRecordRDD }
+import org.bdgenomics.utils.cli._
+import org.kohsuke.args4j.{ Argument, Option => Args4jOption }
+
+/**
+ * Reads2Coverage (accessible as the command 'coverage' through the CLI) takes two arguments,
+ * an Read file and an output file, and calculates the number of reads from the Read file
+ * at every location in the file. Optional arguments are negativeStrands, and positiveStrands,
```
heuermh: argument names should match below

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,148 @@
...
+   */
+  def rdd: RDD[Coverage] = {
+    featureRdd.rdd.map(r => {
+      Coverage(r.getContigName, r.getStart, r.getEnd, r.getScore)
+    })
+  }
+}
+
+/**
+ * Coverage Recordfor CoverageRDD
```
heuermh: Recordfor &rarr; Record for, end with period

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,148 @@
...
+ * @param contigName Specifies chromosomal location of coverage
+ * @param start Specifies start position of coverage
+ * @param end  Specifies end position of coverage
+ * @param count Specifies count of coverage at location
+ */
+case class Coverage(contigName: String, start: Long, end: Long, count: Double)
```
heuermh: I'm thinking `Coverage` here and `CoverageRDD` cover (sorry) your use case appropriately.

This case class and an RDD of continuous values over regions could be made more generic though, if only in name, and in turn support [wiggle](https://genome.ucsc.edu/goldenPath/help/wiggle.html)/[bigWig](https://genome.ucsc.edu/goldenPath/help/bigWig.html)/[bedGraph](https://genome.ucsc.edu/goldenPath/help/bedgraph.html) formats.  I'm not feeling very creative at the moment, there must be something better than `ContinuousValuedFeatureRDD`.  Something for a future feature request?

fnothaft: > cover (sorry)

\*rimshot\*

> This case class and an RDD of continuous values over regions could be made more generic though, if only in name, and in turn support wiggle/bigWig/bedGraph formats. I'm not feeling very creative at the moment, there must be something better than ContinuousValuedFeatureRDD. Something for a future feature request?

+1! Let's open a ticket after merging?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,148 @@
...
+
+/**
+ * RDD holding run length encoded coverage in a FeatureRDD. CoverageRDD stores
+ * coverage counts at a ReferenceRegion in the score attribute of
+ * a feature.
+ *
+ * @param featureRdd FeatureRDD holding records for coverage and corresponding
+ * referenceRegion
+ */
+case class CoverageRDD(featureRdd: FeatureRDD) {
```
fnothaft: I feel a little weird having this wrap a `FeatureRDD`. My preference would be to have the `CoverageRDD` be a `GenomicRDD` that wraps a `RDD[Coverage]`. I feel like this would be more consistent with our recent refactoring. Thoughts?

fnothaft: Then, where you currently return a `RDD[Coverage]`, you'd return a new `CoverageRDD`.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
fnothaft: Lowercase record, as well.

fnothaft: Also, thoughts on moving the `Coverage` case class out to `org.bdgenomics.adam.models`?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDD.scala
```diff
@@ -85,6 +86,78 @@ sealed trait AlignmentRecordRDD extends AvroReadGroupGenomicRDD[AlignmentRecord,
...
+  def toCoverage: CoverageRDD = {
+    val coverage: RDD[(ReferenceRegion, Int)] =
+      rdd.rdd
+        .flatMap(r => {
+          val t: List[Long] = List.range(r.getStart, r.getEnd)
+          t.map(n => (ReferenceRegion(r.getContigName, n, n + 1), 1))
+        }).reduceByKey(_ + _)
+        .cache()
+        .sortByKey()
+        .mapPartitions(i => merge(i)) // merge adjacent regions with the same coverage
```
fnothaft: IMO, this (cache --> sort --> merge) should get factored out to `CoverageRDD`, and the `CoverageRDD` should have some sort of a `collapse` method? There are places where it doesn't necessarily make things simpler to have coverage collapsed down, e.g., https://github.com/bigdatagenomics/quinine/blob/master/quinine-core/src/main/scala/org/bdgenomics/quinine/metrics/targeted/PanelStats.scala#L64-L101.

Also,

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDD.scala
```diff
@@ -85,6 +86,78 @@ sealed trait AlignmentRecordRDD extends AvroReadGroupGenomicRDD[AlignmentRecord,
...
+  }
+
+  /**
+   * Merges adjacent ReferenceRegions with the same value. This is used for merging coverage.
+   * @see toCoverage
+   *
+   * @param iter partition iterator of ReferenceRegion and coverage values
+   * @return merged tuples of adjacent ReferenceRegions and coverage
+   */
+  private def merge(iter: Iterator[(ReferenceRegion, Int)]): Iterator[(ReferenceRegion, Int)] = {
```
fnothaft: Make `private[rdd]` and write unit tests. We'll really want to make sure this one is right...!

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
heuermh: Sure, if there were conversions in both directions (`Feature` <--> `Coverage`)

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
heuermh: +1, with `toFeatureRDD` method on `CoverageRDD` and `toCoverageRDD` method on FeatureRDD.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
heuermh: Yep

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDD.scala
fnothaft: Also, I'd advise against the wrapper. With "reasonable" defaults for the arguments to the tail call function, you can eliminate this function.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDD.scala
```diff
@@ -85,6 +86,78 @@ sealed trait AlignmentRecordRDD extends AvroReadGroupGenomicRDD[AlignmentRecord,
...
+   * @see merge
+   *
+   * @param iter partition iterator of ReferenceRegion and coverage values
+   * @param lastRegion: the last region from a sorted Iterator of regions that has been considered to merge
+   * @param lastCoverage: the last coverage from a sorted Iterator that has been considered to merge
+   * @param condensed: Condensed iterator of iter with adjacent regions with the same value merged
+   * @return merged tuples of adjacent ReferenceRegions and coverage
+   */
+  @tailrec private def merge(iter: Iterator[(ReferenceRegion, Int)],
+                             lastRegion: ReferenceRegion,
```
fnothaft: E.g., lastRegion = _

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDD.scala
```diff
@@ -85,6 +86,78 @@ sealed trait AlignmentRecordRDD extends AvroReadGroupGenomicRDD[AlignmentRecord,
...
+                             lastCoverage: Int,
```
fnothaft: lastCoverage = -1

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDD.scala
```diff
@@ -85,6 +86,78 @@ sealed trait AlignmentRecordRDD extends AvroReadGroupGenomicRDD[AlignmentRecord,
...
+                             condensed: List[(ReferenceRegion, Int)]): Iterator[(ReferenceRegion, Int)] = {
```
fnothaft: condensed = List.empty

would do the trick

fnothaft:
Hate to be a PITA, but would it be possible to split the deletion of dead code out into a separate commit? LMK if that'd be too much of a hassle.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -757,6 +757,11 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
     FragmentRDD.fromRdd(records.map(fastqRecordConverter.convertFragment))
   }
 
+  def loadCoverage(filePath: String, minPartitions: Option[Int] = None): CoverageRDD = {
```
fnothaft: Add scaladoc. Drop unused `minPartitions` argument?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,148 @@
...
+import org.apache.spark.rdd.RDD
+import org.bdgenomics.adam.models.ReferenceRegion
+
+/**
+ * RDD holding run length encoded coverage in a FeatureRDD. CoverageRDD stores
+ * coverage counts at a ReferenceRegion in the score attribute of
+ * a feature.
+ *
+ * @param featureRdd FeatureRDD holding records for coverage and corresponding
+ * referenceRegion
```
fnothaft: Nit: indent 2 spaces

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,148 @@
...
+   */
+  def getCoverage(region: ReferenceRegion, bpPerBin: Int = 1): RDD[Coverage] = {
+
+    val flattenedCoverage = features2Coverage(region)
+
+    if (bpPerBin == 1) {
+      flattenedCoverage // no binning, return raw results
+    } else {
+      // subtract region.start to shift mod to start of ReferenceRegion
+      flattenedCoverage.filter(r => (r.start - region.start) % bpPerBin == 0)
```
fnothaft: So, when I hear "binning", I think "averaging" not "sampling".

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,148 @@
...
+  /**
+   * Gets coverage overlapping specified ReferenceRegion. For large, ReferenceRegions,
+   * base pairs per bin (bpPerBin) can be specified to bin together ReferenceRegions of
+   * equal size. The coverage of each bin is coverage of the first base pair in that bin.
+   *
+   * @param region ReferenceRegion to fetch overlapping coverage from
+   * @param bpPerBin base pairs per bin, number of bases to combine to one bin
+   * @return RDD of Coverage Records
+   */
+  def getCoverage(region: ReferenceRegion, bpPerBin: Int = 1): RDD[Coverage] = {
```
fnothaft: What exactly is the use case for this, as distinct from `filterByOverlappingRegion`?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,148 @@
...
+
+    if (bpPerBin == 1) {
+      flattenedCoverage // no binning, return raw results
+    } else {
+      flattenedCoverage.map(r => {
+        // key coverage by binning start site mod bpPerbin
+        // subtract region.start to shift mod to start of ReferenceRegion
+        val key = r.start - ((r.start - region.start) % bpPerBin)
+        (key, r)
+      }).groupByKey
```
fnothaft: Avoid `groupByKey`. I think this works as a reduce by key with the following changes:

```
def reduceFn(a: (Int, Int), b: (Int, Int)): (Int, Int) = {
  (a._1 + b._1, a._2 + b._2)
}
flattenedCoverage.map(r => {
         val key = r.start - ((r.start - region.start) % bpPerBin)
         (key, (r, 1))
      }).reduceByKey(reduceFn)
  .map(r => Coverage(r._1.contigName, r._1.pos, r._1.pos + bpPerBin, r._2._1.toDouble / r._2._2.toDouble))
```

fnothaft: Also, I don't like keying here with just the start position. Generally, I prefer keying with a `ReferencePosition`/`ReferenceRegion`.

fnothaft: Also, I'd like to break this out into `filterByOverlappingRegion` and `aggregateCoverage`.

Also, avoid `get` in function names unless they are a setter/getter.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,148 @@
...
+   * coverage.
+   * @param region Region to filter FeatureRDD
+   * @return Flattened RDD of coverage
+   */
+  private def features2Coverage(region: ReferenceRegion): RDD[Coverage] = {
+    val features = featureRdd.filterByOverlappingRegion(region)
+
+    // flatmap features in region to coverage
+    features.rdd.flatMap(r => {
+      val positions: List[Long] = List.range(r.getStart, r.getEnd)
```
fnothaft: Prefer `r.getStart to r.getEnd`

fnothaft: Actually, `until`, not `to`.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,148 @@
...
+      positions.map(n => Coverage(r.getContigName, n, n + 1, r.getScore))
+    }).filter(r => r.start >= region.start && r.start < region.end) // filter out positions from flanking regions
+  }
+
+  /**
+   * Gets flattened RDD of coverage, with coverage at each base pair
+   * @return RDD[Coverage] from underlying FeatureRDD
+   */
+  def flattenedRdd: RDD[Coverage] = {
```
fnothaft: Factor this out? Shows up both here and at https://github.com/bigdatagenomics/adam/pull/1108/files#diff-e3a56fa513aba425fb2edba5451d6883R112.

fnothaft: Also, for methods that "do heavy work", prefer () even if there are no parameters.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
fnothaft: Yeah I feel like there should be an `object Coverage` with methods `private[adam] def apply(feature: Feature)` and `private[adam] def apply(rdd: RDD[Feature])` and then a `def toFeature: Feature` method in the case class.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDD.scala
fnothaft: Sorry, forgot to finish writing the thought after the orphan `Also,`. Wherever you have a `cache`, that should be paired with an `unpersist`.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
akmorrow13: filterByOverlappingRegion returns pure records. getCoverage and getAggregatedCoverage return flattened coverage that can be binned through sampling and aggregation, respectively.

fnothaft: I'd prefer to chain `filterByOverlappingRegion` with `binCoverage` or `aggregateCoverage` than have them rolled together.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
akmorrow13: How could we fit getCoverage to filterByOverlappingRegion? I beleive these are different functions because filterByOverlappingRegion returns raw records. For example, if the coverage records are not merged to begin with, then filterByOverlappingRegion will return chunks of coverage. Also, filterByOverlappingRegion cannot specify binning.

akmorrow13: We could always merge getCoverage() and aggregatedCoverage() back into 1 function. Then there would be filterByOverlappingRegion and aggregatedCoverage, with the option to reduce by key or sample.

fnothaft: Ah, my earlier comment was unclear. What I mean is, currently, the method signature is:

`def getAggregatedCoverage(region: ReferenceRegion, bpPerBin: Int = 1)`

I don't like having the `region: ReferenceRegion` in here. I'd like the signature to be:

`def aggregateCoverage(bpPerBin: Int = 1)`

and if the user wants to filter out coverage by a specific region, they call `filterByOverlappingRegion` before calling `aggregateCoverage`.

heuermh: +1

akmorrow13: Ahh I see. Much cleaner!

akmorrow13:
Jenkins, retest this please.

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Reads2Coverage.scala
```diff
@@ -0,0 +1,85 @@
...
+import org.bdgenomics.adam.projections.AlignmentRecordField._
+import org.bdgenomics.adam.projections.Projection
+import org.bdgenomics.adam.rdd.ADAMContext._
+import org.bdgenomics.adam.rdd.features.CoverageRDD
+import org.bdgenomics.adam.rdd.read.{ AlignedReadRDD, AlignmentRecordRDD }
+import org.bdgenomics.utils.cli._
+import org.kohsuke.args4j.{ Argument, Option => Args4jOption }
+
+/**
+ * Reads2Coverage (accessible as the command 'coverage' through the CLI) takes two arguments,
```
fnothaft: Nit: should be "accessible as the command 'reads2coverage'".

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Reads2Coverage.scala
```diff
@@ -0,0 +1,85 @@
...
+
+class Reads2Coverage(protected val args: Reads2CoverageArgs) extends BDGSparkCommand[Reads2CoverageArgs] {
+  val companion: BDGCommandCompanion = Reads2Coverage
+
+  def run(sc: SparkContext): Unit = {
+
+    val proj = Projection(contigName, start, end, cigar)
+
+    // If saving strand specific coverage, require that only one direction is specified
+    require((!args.onlyNegativeStrands || !args.onlyPositiveStrands),
```
fnothaft: Nit: This isn't incorrect—DeMorgan's law, and all—but this would read more clearly as `!(args.onlyNegativeStrands && args.onlyPositiveStrands)`

adam-core/src/main/scala/org/bdgenomics/adam/models/Coverage.scala
```diff
@@ -0,0 +1,62 @@
...
+
+import org.apache.spark.rdd.RDD
+import org.bdgenomics.formats.avro.Feature
+
+/**
+ * Converts from avro Feature to Coverage.
+ */
+object Coverage {
+
+  private[adam] def apply(region: ReferenceRegion, count: Double): Coverage = {
```
fnothaft: Please scaladoc all methods.

adam-core/src/main/scala/org/bdgenomics/adam/models/Coverage.scala
```diff
@@ -0,0 +1,62 @@
...
+ * @param contigName Specifies chromosomal location of coverage
+ * @param start Specifies start position of coverage
+ * @param end  Specifies end position of coverage
+ * @param count Specifies count of coverage at location
+ */
+case class Coverage(contigName: String, start: Long, end: Long, count: Double) {
+
+  /**
+   * Converts Coverage to Feature, setting Coverage count in the score attribute
+   * @return Feature build from Coverage
```
fnothaft: Nit: s/build/built/g

adam-core/src/main/scala/org/bdgenomics/adam/models/Coverage.scala
```diff
@@ -0,0 +1,62 @@
...
+ * Coverage record for CoverageRDD
+ * @param contigName Specifies chromosomal location of coverage
+ * @param start Specifies start position of coverage
+ * @param end  Specifies end position of coverage
+ * @param count Specifies count of coverage at location
+ */
+case class Coverage(contigName: String, start: Long, end: Long, count: Double) {
+
+  /**
+   * Converts Coverage to Feature, setting Coverage count in the score attribute
```
fnothaft: Nit: should have whitespace here.

adam-core/src/main/scala/org/bdgenomics/adam/models/Coverage.scala
```diff
@@ -0,0 +1,62 @@
...
+    Coverage(feature.getContigName, feature.getStart, feature.getEnd, feature.getScore)
+  }
+
+  private[adam] def apply(rdd: RDD[Feature]): RDD[Coverage] = {
+    rdd.map(f => Coverage(f))
+  }
+}
+
+/**
+ * Coverage record for CoverageRDD
```
fnothaft: Nit: summary line should be more descriptive and followed by whitespace

adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
```diff
@@ -153,6 +153,10 @@ object ReferenceRegion {
   def apply(feature: Feature): ReferenceRegion = {
     new ReferenceRegion(feature.getContigName, feature.getStart, feature.getEnd)
   }
+
+  def apply(coverage: Coverage): ReferenceRegion = {
```
fnothaft: Add scaladoc

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -42,7 +42,7 @@ import org.bdgenomics.adam.converters._
 import org.bdgenomics.adam.instrumentation.Timers._
 import org.bdgenomics.adam.io._
 import org.bdgenomics.adam.models._
-import org.bdgenomics.adam.projections.{ AlignmentRecordField, NucleotideContigFragmentField, Projection }
+import org.bdgenomics.adam.projections.{ FeatureField, AlignmentRecordField, NucleotideContigFragmentField, Projection }
```
fnothaft: Please keep all imports alphabetically sorted (FeatureField should be between AlignmentRecordField/NCFField) and break this line as it is too long.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -757,6 +757,11 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
     FragmentRDD.fromRdd(records.map(fastqRecordConverter.convertFragment))
   }
 
+  def loadCoverage(filePath: String): CoverageRDD = {
```
fnothaft: Add scaladoc.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,225 @@
...
+ */
+package org.bdgenomics.adam.rdd.features
+
+import org.apache.spark.rdd.RDD
+import org.bdgenomics.adam.models.{ SequenceRecord, SequenceDictionary, Coverage, ReferenceRegion }
+import org.bdgenomics.adam.rdd.GenomicRDD
+
+import scala.annotation.tailrec
+
+object CoverageRDD {
```
fnothaft: Add scaladoc.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,225 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.bdgenomics.adam.rdd.features
+
+import org.apache.spark.rdd.RDD
+import org.bdgenomics.adam.models.{ SequenceRecord, SequenceDictionary, Coverage, ReferenceRegion }
+import org.bdgenomics.adam.rdd.GenomicRDD
+
```
fnothaft: Nit: extra whitespace.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,225 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.bdgenomics.adam.rdd.features
+
+import org.apache.spark.rdd.RDD
+import org.bdgenomics.adam.models.{ SequenceRecord, SequenceDictionary, Coverage, ReferenceRegion }
```
fnothaft: Nit: sort import.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,225 @@
...
+  /**
+   * @param elem The coverage element to extract a sequence record from.
+   * @return Gets the SequenceRecord for coverage.
+   */
+  private def getSequenceRecord(elem: Coverage): SequenceRecord = {
+    SequenceRecord(elem.contigName, 1L)
+  }
+
+  /**
+   * Builds a CoverageRDD without SequenceDictionary information by running an
```
fnothaft: Nit: first line of scaladoc should be a summary (<80 char).

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,225 @@
...
+   * aggregate to rebuild the SequenceDictionary.
+   *
+   * @param rdd The underlying Coverage RDD to build from.
```
fnothaft: Prefer "RDD of coverage"

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,225 @@
...
+import scala.annotation.tailrec
+
+object CoverageRDD {
+
+  /**
+   * @param elem The coverage element to extract a sequence record from.
+   * @return Gets the SequenceRecord for coverage.
+   */
+  private def getSequenceRecord(elem: Coverage): SequenceRecord = {
+    SequenceRecord(elem.contigName, 1L)
```
fnothaft: I don't like this, because we know that the reference sequence doesn't have length 1. Why do we need to have this? If we're converting from a `FeatureRDD`, we should have a well formed `SequenceDictionary` already.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,225 @@
...
+   * Builds a CoverageRDD without SequenceDictionary information by running an
+   * aggregate to rebuild the SequenceDictionary.
+   *
+   * @param rdd The underlying Coverage RDD to build from.
+   * @return Returns a new CoverageRDD.
+   */
+  private[rdd] def apply(rdd: RDD[Coverage]): CoverageRDD = {
+
+    // aggregate to create the sequence dictionary
+    val sd = new SequenceDictionary(rdd.map(getSequenceRecord)
```
fnothaft: See above comment.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,225 @@
...
+      .distinct
+      .collect
+      .toVector)
+
+    CoverageRDD(rdd, sd)
+  }
+}
+
+case class CoverageRDD(rdd: RDD[Coverage],
```
fnothaft: Add scaladoc.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,225 @@
...
+                       sequences: SequenceDictionary) extends GenomicRDD[Coverage, CoverageRDD] {
+
+  /**
+   * Saves coverage as feature file.
+   * @see FeatureRDD.save
```
fnothaft: Whitespace before this line.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,225 @@
...
+   *   val coverage = feature.getScore
+   *
+   * @param filePath The location to write the output.
+   */
+  def save(filePath: java.lang.String) = {
+    this.toFeatureRDD.save(filePath)
+  }
+
+  /**
+   * Merges adjacent ReferenceRegions with the same value. This is used for merging coverage.
```
fnothaft: > This is used for merging coverage.

This isn't really detailed. Split it out of the summary line and expand upon the explanation.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,225 @@
...
+   */
+  def collapse: CoverageRDD = {
+    val newRDD: RDD[Coverage] = rdd
+      .mapPartitions(iter => {
+        if (iter.hasNext) {
+          val first = iter.next
+          collapse(iter, first, List.empty)
+        } else iter
+      })
+    this.replaceRdd(newRDD)
```
fnothaft: Don't need the `this.`

fnothaft: Also, prefer whitespace before the line.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  /**
+   * Tail recursion for merging adjacent ReferenceRegions with the same value.
+   * @see merge
+   *
+   * @param iter partition iterator of ReferenceRegion and coverage values
+   * @param lastCoverage: the last coverage from a sorted Iterator that has been considered to merge
```
fnothaft: No colons here and in the line below after param name.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,225 @@
...
+   */
+  @tailrec private def collapse(iter: Iterator[Coverage],
+                                lastCoverage: Coverage,
+                                condensed: List[Coverage]): Iterator[Coverage] = {
+    if (!iter.hasNext) {
+      // if lastCoverage has not yet been added, add to condensed
+      val nextCondensed =
+        if (condensed.map(r => ReferenceRegion(r)).filter(_.overlaps(ReferenceRegion(lastCoverage))).isEmpty) {
+          lastCoverage :: condensed
+        } else condensed
```
fnothaft: Be consistent with braces. If the leading `if` clause has braces, the `else` clause (and any `else if`s) should as well.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,225 @@
...
+      nextCondensed.toIterator
+    } else {
+      val cov: Coverage = iter.next
+      val rr = ReferenceRegion(cov)
+      val lastRegion = ReferenceRegion(lastCoverage)
+      val (nextCoverage: Coverage, nextCondensed: List[Coverage]) =
+        if (rr.isAdjacent(lastRegion) && lastCoverage.count == cov.count) {
+          (Coverage(rr.merge(lastRegion), lastCoverage.count), condensed)
+        } else {
+          (cov, (lastCoverage) :: condensed)
```
fnothaft: Nit: remove parens around `lastCoverage`.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,225 @@
...
+                                condensed: List[Coverage]): Iterator[Coverage] = {
+    if (!iter.hasNext) {
+      // if lastCoverage has not yet been added, add to condensed
+      val nextCondensed =
+        if (condensed.map(r => ReferenceRegion(r)).filter(_.overlaps(ReferenceRegion(lastCoverage))).isEmpty) {
+          lastCoverage :: condensed
+        } else condensed
+      nextCondensed.toIterator
+    } else {
+      val cov: Coverage = iter.next
```
fnothaft: Stylistically, I don't love the explicit typing here. Is it needed, or can we let `scalac` infer the types?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,225 @@
...
+  /**
+   * @return Returns a FeatureRDD from CoverageRDD
+   */
+  def toFeatureRDD: FeatureRDD = {
+    val featureRdd = rdd.map(_.toFeature)
+    FeatureRDD(featureRdd, sequences)
+  }
+
+  /**
+   * Gets coverage overlapping specified ReferenceRegion. For large, ReferenceRegions,
```
fnothaft: Split after first sentence.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,225 @@
...
+   * base pairs per bin (bpPerBin) can be specified to bin together ReferenceRegions of
+   * equal size. The coverage of each bin is coverage of the first base pair in that bin.
+   *
+   * @param bpPerBin base pairs per bin, number of bases to combine to one bin
+   * @return RDD of Coverage Records
+   */
+  def coverage(bpPerBin: Int = 1): CoverageRDD = {
+
+    val flattened = this.flatten
```
fnothaft: No `this.`, `flatten` -> `flatten()`

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,225 @@
...
+
+    if (bpPerBin == 1) {
+      flattened // no binning, return raw results
+    } else {
+      // subtract region.start to shift mod to start of ReferenceRegion
+      val newRDD = flattened.rdd.filter(r => r.start % bpPerBin == 0)
```
fnothaft: Prefer `.transform(_.filter...)` to call to `.replaceRdd`

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,225 @@
...
+      flattened.replaceRdd(newRDD)
+    }
+  }
+
+  /**
+   * Gets coverage overlapping specified ReferenceRegion. For large, ReferenceRegions,
```
fnothaft: Split at `For large`, no comma before `ReferenceRegion`.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,225 @@
...
+   * base pairs per bin (bpPerBin) can be specified to bin together ReferenceRegions of
+   * equal size. The coverage of each bin is the mean coverage over all base pairs in that bin.
+   *
+   * @param bpPerBin base pairs per bin, number of bases to combine to one bin
+   * @return RDD of Coverage Records
+   */
+  def aggregatedCoverage(bpPerBin: Int = 1): CoverageRDD = {
+
+    val flattened = this.flatten
```
fnothaft: No `this.`, `flatten` -> `flatten()`.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,225 @@
...
+          // subtract region.start to shift mod to start of ReferenceRegion
+          val start = r.start - (r.start % bpPerBin)
+          ReferenceRegion(r.contigName, start, start + bpPerBin)
+        }).mapValues(r => (r.count, 1))
+        .reduceByKey(reduceFn)
+        .map(r => {
+          // compute average coverage in bin
+          Coverage(r._1.referenceName, r._1.start, r._1.end, r._2._1 / r._2._2)
+        })
+      flattened.replaceRdd(newRDD)
```
fnothaft: In general, I prefer `.transform`. I would factor L173–L184 into a function that is called by `.transform`.

fnothaft: The way I look at the `replaceRdd` method is that it exists solely so that the `transform` method could be implemented in the `GenomicRDD` trait. Ideally, it's never called elsewhere. However, I don't think this intention is documented. I'll go back and doc this.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,225 @@
...
+  protected def replaceRdd(newRdd: RDD[Coverage]): CoverageRDD = {
+    copy(rdd = newRdd)
+  }
+
+  /**
+   * Gets flattened RDD of coverage, with coverage mapped to a reference Region at each base pair
+   * @return CoverageRDD of flattened Coverage records
+   */
+  def flatten: CoverageRDD = {
+    this.replaceRdd(flatMapCoverage(rdd))
```
fnothaft: No `this.`, same comment above RE: transform --> this becomes `transform(flatMapCoverage)`

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
```diff
@@ -0,0 +1,225 @@
...
+  }
+
+  /**
+   * Flat maps coverage into ReferenceRegion and counts for each base pair.
+   * @param rdd
+   * @return
```
fnothaft: Complete scaladoc.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureRDD.scala
```diff
@@ -246,6 +246,14 @@ case class FeatureRDD(rdd: RDD[Feature],
   }
 
   /**
+   * @return Returns CoverageRDD build from FeatureRDD
```
fnothaft: s/build/built/g

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDD.scala
```diff
@@ -41,6 +41,7 @@ import org.bdgenomics.adam.converters.AlignmentRecordConverter
 import org.bdgenomics.adam.instrumentation.Timers._
 import org.bdgenomics.adam.models._
 import org.bdgenomics.adam.rdd.ADAMContext._
+import org.bdgenomics.adam.rdd.features.CoverageRDD
```
fnothaft: `org.bdgenomics.adam.rdd.features` should come after `org.bdgenomics.adam.rdd` and before `org.bdgenomics.adam.rdd.read`

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDD.scala
```diff
@@ -85,6 +86,32 @@ sealed trait AlignmentRecordRDD extends AvroReadGroupGenomicRDD[AlignmentRecord,
...
+   * counted at all overlapping positions
+   */
+  def toCoverage(collapse: Boolean = true): CoverageRDD = {
+    val coverage: RDD[Coverage] =
+      rdd.rdd
+        .flatMap(r => {
+          val t: List[Long] = List.range(r.getStart, r.getEnd)
+          t.map(n => (ReferenceRegion(r.getContigName, n, n + 1), 1))
+        }).reduceByKey(_ + _)
+        .cache()
```
fnothaft: You don't want the `.cache()` call here. I mean, you do, but you don't want the code after it. This should be broken up:

```
val covCounts: RDD[Coverage] =
       rdd.rdd
         .flatMap(r => {
           val t: List[Long] = List.range(r.getStart, r.getEnd)
           t.map(n => (ReferenceRegion(r.getContigName, n, n + 1), 1))
         }).reduceByKey(_ + _)
         .cache()

val coverage = covCounts.sortByKey()
        .map(r => Coverage(r._1, r._2.toDouble))
```

fnothaft: Also, why are we sorting here? If we don't collapse, there's no reason to do the sort. I think it would make sense to move the `sortByKey` into `collapse`, since the data being sorted is an invariant of collapse.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDD.scala
```diff
@@ -85,6 +86,32 @@ sealed trait AlignmentRecordRDD extends AvroReadGroupGenomicRDD[AlignmentRecord,
...
+        .sortByKey()
+        .map(r => Coverage(r._1, r._2.toDouble))
+
+    val coverageRdd =
+      if (collapse)
+        CoverageRDD(coverage, sequences).collapse
+      else CoverageRDD(coverage, sequences)
+
+    coverage.unpersist()
```
fnothaft: You don't want to unpersist `coverage`. You'd want to unpersist `covCounts` from https://github.com/bigdatagenomics/adam/pull/1108/files#r75248157. Calling unpersist on `coverage` should cause everything done to the `CoverageRDD` returned by this method to throw an exception...

adam-core/src/test/scala/org/bdgenomics/adam/rdd/features/CoverageRDDSuite.scala
```diff
@@ -0,0 +1,106 @@
...
+
+import java.io.File
+import org.bdgenomics.adam.models.{ Coverage, ReferenceRegion }
+import org.bdgenomics.adam.rdd.ADAMContext._
+import org.bdgenomics.adam.util.ADAMFunSuite
+import org.bdgenomics.formats.avro.Feature
+
+class CoverageRDDSuite extends ADAMFunSuite {
+
+  def tempLocation(suffix: String = ".adam"): String = {
```
fnothaft: Don't copy this; factor it out of `ADAMContextSuite` into `ADAMFunSuite`.

adam-core/src/test/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDSuite.scala
```diff
@@ -22,12 +22,14 @@ import java.nio.file.Files
 import htsjdk.samtools.ValidationStringency
 import org.apache.spark.rdd.RDD
 import org.bdgenomics.adam.models.{
+  ReferenceRegion,
```
fnothaft: `ReferenceRegion` should be after `RecordGroupDictionary`.

fnothaft:
Dropped a variety of nits inline; all in all looks good!

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/CoverageRDD.scala
akmorrow13: we are not converting from a FeatureRDD, but from an RDD[Coverage]. FeatureRDD does the same thing when a sequence dictionary is not provided (see https://github.com/bigdatagenomics/adam/blob/810806c64e39ff8964f8ed6e3becd4230abc660b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureRDD.scala line 84).

akmorrow13: Ya I see what you mean I don't think this code is even correct.  I don't think there is a correct way to extract a sequence dictionary from either RDD[Feature] or RDD[Coverage].

adam-core/src/main/scala/org/bdgenomics/adam/models/Coverage.scala
akmorrow13: What does this mean?

fnothaft: [It's sed syntax for "replace string"](http://www.grymoire.com/Unix/Sed.html#uh-1). All I'm saying is that `built` is misspelled as `build`.

akmorrow13:
Jenkins, test this please.

akmorrow13:
Jenkins, retest this please.

fnothaft:
Thanks for merging my PR against this PR, @akmorrow13! Can you remove the merge commit https://github.com/bigdatagenomics/adam/pull/1108/commits/b13eee5f4bff3fbc6816336b0b2afa69eed21809?

heuermh:
Closing in favor of #1128

PR #1116, opened 2016-08-14
[ADAM-1087] Migrate away from FileSystem.get

### LINKED ISSUE #1087: Be more rigorous with FileSystem.get
We often use [`FileSystem.get`](http://hadoop.apache.org/docs/r2.4.1/api/org/apache/hadoop/fs/FileSystem.html#get(org.apache.hadoop.conf.Configuration)) to get the file system that underlies a cluster, e.g. https://github.com/bigdatagenomics/adam/blob/810806c64e39ff8964f8ed6e3becd4230abc660b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMRDDFunctions.scala#L76. However, this checks the given file URI against the `fs.defaultFS` parameter, and throws if the file systems don't match. This isn't usually a problem, but when running Spark-on-YARN with data on an NFS (file:// URI), you can't get this to work. Overriding `fs.defaultFS` to `file:///` causes job startup issues.

fnothaft:
FileSystem.get returns the configured default file system. This can cause issues when loading a file off a non-default FileSystem, e.g., an NFS or S3. This PR changes to use Path.getFileSystem. Resolves #1087.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
@@ -296,7 +296,7 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
     val path = new Path(filePath)
     val fs =
       Option(
-        FileSystem.get(path.toUri, sc.hadoopConfiguration)
+        path.getFileSystem(sc.hadoopConfiguration)
       ).getOrElse(
           throw new FileNotFoundException(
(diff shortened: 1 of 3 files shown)


fnothaft:
I am -1ing this until I verify it on the cluster.

heuermh:
LGTM.  Yes, it would be good to verify on the cluster.

fnothaft:
+1, as this is working a-OK on the cluster. I will rebase on the latest changes, and then I think we're good for a merge.

fnothaft:
Ping for merge.

heuermh:
Merged as commit 4140095736332d6ae94418595c3911b8ffa171dd.

Thank you, @fnothaft!

PR #1119, opened 2016-08-18
remove duplicated dependency

ryan-williams:


fnothaft:
Thanks @ryan-williams! Merged.

PR #1130, opened 2016-08-25
Remove StructuralVariant and StructuralVariantType, add names field to Variant

heuermh:
Still left to do (in a later pull request) is to populate the `somatic` field.


heuermh:
Sorry, wrong base branch.

PR #1133, opened 2016-08-26
[ADAM-1132] Fix improper quoting of bash args in adam-shell.

### LINKED ISSUE #1132: adam-shell not handling bash args correctly
https://github.com/bigdatagenomics/adam/blob/33ab79520936d304b8a8a8384192e4b7fe240c97/bin/adam-shell#L74

can't do e.g.

```
adam-shell --conf "spark.driver.extraJavaOptions=-XX:+PrintGCDetails -XX:+PrintGCTimeStamps -verbose:gc"
```

though the same cmd with `s/adam/spark/` will work, as it should

fnothaft:
Resolves #1132. Eliminates the need for `append_to_option.py`, so I removed said file.

TIL http://stackoverflow.com/questions/1668649/how-to-keep-quotes-in-args/1669548#1669548


heuermh:
Did you run into something that didn't work with the way it was before, or is this more of a code cleanup?

heuermh:
Works for me.

ryan-williams:
wfm2, thx for quick fix frank

heuermh:
Merged as commit a55d2cbde573214f9c84ffbf0cae9c53ae0174f4.

Thank you, @fnothaft @ryan-williams

Issue #1139, opened 2016-08-29
JFYI: tiny forked adam-core "0.20.0" release


ryan-williams:
I'd been building and making available ADAM JARs from various HEAD SHAs for my lab over the last few months, since I've needed some of our scripts to pick up post-0.19.0 fixes (like the bash-arg-parsing fixes for #972 and #1132).

That was fine, but even more recently I needed to give pageant an ADAM Maven coordinate to depend on with some of the new APIs, so after weighing some options I went ahead and made a small fork of adam-core here https://github.com/hammerlab/adam/tree/adam-core and [released it to Maven as `org.hammerlab.adam:adam-core_2.10:0.20.0`](https://oss.sonatype.org/content/repositories/releases/org/hammerlab/adam/adam-core_2.10/0.20.0/).

A couple of interesting things happened along the way:
- I made `adam-core` its own standalone project, removed the other modules, and put it in the top of the repo.
- I did a few other POM hygiene improvements that I'd like to upstream here when I have a moment.

I don't plan to touch that fork any more, if I can help it, just wanted to mention this here and point out the repo in case anyone is curious to inspect some of the Maven-futzing involved with pulling the core module out to stand on its own.

Feel free to close this issue whenever.

fnothaft:
Thanks for the heads up, @ryan-williams!

> I did a few other POM hygiene improvements that I'd like to upstream here when I have a moment.

That'd be fantastic!

> I made adam-core its own standalone project, removed the other modules, and put it in the top of the repo.

OOC, what are the benefits from your side to having the different modules split out? From my side, the submodules make some usage hairy (specifically, pushing to spark-packages) but makes it easier to build the whole project.

ryan-williams:
K, just sent my changes in #1142.

> OOC, what are the benefits from your side to having the different modules split out? From my side, the submodules make some usage hairy (specifically, pushing to spark-packages) but makes it easier to build the whole project.

I didn't start with a goal of "having a repo that is just adam-core", but given that all I wanted to fork/deploy was adam-core, reducing the project to that one module seemed like the sane thing to do.

[...]

PR #1142, opened 2016-08-31
misc pom/test/resource improvements

ryan-williams:
these happened while i was doing #1139; individual commits should be reasonably self-explanatory


ryan-williams:
Some problem with where test-coverage reports are getting written after I changed directory configs from `surefire-reports` to `scalatest-reports`; I assumed that was find to just do everywhere but maybe something not in this repo has `surefire-reports` hard-coded? Looking…

ryan-williams:
The end of the log seems to get through the end of `jenkins-test` script successfully, so I can't tell what it's trying to do when the failure occurs:

```
+ echo 'All the tests passed'
All the tests passed
echo
+ echo

Recording test results
ERROR: Step ?Publish JUnit test result report? failed: No test report files were found. Configuration error?
Publishing Scoverage XML and HTML report...
Setting commit status on GitHub for https://github.com/bigdatagenomics/adam/commit/3722878646d30b56dcaa16c9624bc0467db0486c
Finished: FAILURE
```

fnothaft:
@ryan-williams IIRC, Jenkins has the test report paths hardcoded. Let me take a look at the config this AM!

heuermh:
Looks reasonable to me.

With `scala.version.prefix` will need to confirm the `move_to...` scripts still work correctly.

fnothaft:
Jenkins, test this please.

src/test/resources/.gitignore
```diff
@@ -0,0 +1,3 @@
+# We blanket-ignore .bam files in the top-level of the repo, but want to undo that here.
+!*.bam
```
fnothaft: Nifty! Didn't know you could do that.

fnothaft:
Also, I have done the `surefire_reports` --> `scalatest_reports` path change in Jenkins, so that should be good to go.

src/test/resources/.gitignore
ryan-williams: Yea this caused me some problems when I tried to `git mv adam-core/* ./` to put all of adam-core in the top directory (after `git rm`ing the other modules); `sorted.bam*` did not come with that move since they were `.gitignore`d (despite having been `git add`ed to the repo at some point, maybe before they became `.gitignore`d?), so I was just missing them in my builds, which took me a while to figure out.

In any case, yea `.gitignore` "whitelists" can be useful in addition to the way it's usually used.

ryan-williams:
OK, just pushed an attempt at updating the `move_to` scripts to use `2.11.8` and `2.10.6`

src/test/resources/.gitignore
fnothaft: Yeah originally we had a prohibition on adding binary files, but we've relaxed that over time. Thanks for patching this!

pom.xml
```diff
@@ -268,7 +268,8 @@
           </execution>
         </executions>
         <configuration>
-          <scalaVersion>2.10.4</scalaVersion>
+          <scalaVersion>${scala.version}</scalaVersion>
+          <recompileMode>incremental</recompileMode>
```
fnothaft: We'd removed incremental in the past. I'd need to dig through the commit history to find the exact reason, but it has something to do with causing transient build issues. The Jenkins build just failed with an issue related to such. Alternatively, we could do `-DrecompileMode=all` in `scripts/jenkins-test` as suggested by https://github.com/davidB/scala-maven-plugin/issues/185. I'm open to either. It'd be nice to have faster Scala compile times in the general case.

ryan-williams: Interesting; so incremental compilation in Scala 2.11 either doesn't work correctly at all, or only works in SBT?

Maybe a better option would be to enable a different Maven profile in the presence of a jenkins environment variable?

That way we would get incremental compiles locally, which would be nice, unless it's so broken we don't think we want that.

ryan-williams:
Just following up here, the last build failed on 2.6.0 x 1.5.2 x 2.10 with:

```
[error] Required file not found: sbt-interface.jar
[error] See zinc -help for information about locating necessary files
```

and 2.6.0 x 1.5.2 x 2.11 with:

```
[ERROR] The build could not read 1 project -> [Help 1]
[ERROR]   
[ERROR]   The project org.bdgenomics.adam:adam-core_2.11:0.19.1-SNAPSHOT (/home/jenkins/workspace/ADAM-prb/HADOOP_VERSION/2.6.0/SCALAVER/2.11/SPARK_VERSION/1.5.2/label/centos/adam-core/pom.xml) has 1 error
[ERROR]     'dependencies.dependency.version' for org.apache.parquet:parquet-scala_2.10:jar is missing. @ line 155, column 17
[ERROR] 
```

I don't have obvious leads about either one.

heuermh:
The `parquet-scala_2.10` dependency is strange as there is no 2.11 version.
http://search.maven.org/#search%7Cga%7C1%7Cparquet-scala

The `move_to_scala_2.11.sh` script has a workaround, maybe something went wrong
https://github.com/bigdatagenomics/adam/blob/master/scripts/move_to_scala_2.11.sh#L6

ryan-williams:
I think the missing sbt-interface.jar has to do with `zinc` not being installed or configured properly on the Jenkins build machine. Not sure what to do about it, maybe I'll try to just turn off incremental-compilation, overall or just for Jenkins.

ryan-williams:
All the Spark 2.0.0 builds failed with:

```
mvn clean -Dhadoop.version=2.3.0 -Dspark.version=2.0.0
Java HotSpot(TM) 64-Bit Server VM warning: ignoring option MaxPermSize=1g; support was removed in 8.0
[INFO] Scanning for projects...
[ERROR] The build could not read 4 projects -> [Help 1]
[ERROR]   
[ERROR]   The project org.bdgenomics.adam:adam-core-spark2_2.10:0.19.1-SNAPSHOT (/home/jenkins/workspace/ADAM-prb/HADOOP_VERSION/2.3.0/SCALAVER/2.10/SPARK_VERSION/2.0.0/label/centos/adam-core/pom.xml) has 5 errors
[ERROR]     'dependencies.dependency.version' for org.bdgenomics.utils:utils-misc-spark2_2.10:test-jar is missing. @ line 93, column 17
[ERROR]     'dependencies.dependency.version' for org.bdgenomics.utils:utils-metrics-spark2_2.10:jar is missing. @ line 99, column 17
[ERROR]     'dependencies.dependency.version' for org.bdgenomics.utils:utils-io-spark2_2.10:jar is missing. @ line 103, column 17
[ERROR]     'dependencies.dependency.version' for org.bdgenomics.utils:utils-cli-spark2_2.10:jar is missing. @ line 107, column 17
[ERROR]     'dependencies.dependency.version' for org.bdgenomics.utils:utils-intervalrdd-spark2_2.10:jar is missing. @ line 111, column 17
[ERROR]   
[ERROR]   The project org.bdgenomics.adam:adam-apis-spark2_2.10:0.19.1-SNAPSHOT (/home/jenkins/workspace/ADAM-prb/HADOOP_VERSION/2.3.0/SCALAVER/2.10/SPARK_VERSION/2.0.0/label/centos/adam-apis/pom.xml) has 3 errors
[ERROR]     'dependencies.dependency.version' for org.bdgenomics.utils:utils-misc-spark2_2.10:test-jar is missing. @ line 105, column 17
[ERROR]     'dependencies.dependency.version' for org.bdgenomics.adam:adam-core-spark2_2.10:jar is missing. @ line 115, column 17
[ERROR]     'dependencies.dependency.version' for org.bdgenomics.adam:adam-core-spark2_2.10:test-jar is missing. @ line 119, column 17
[ERROR]   
[ERROR]   The project org.bdgenomics.adam:adam-cli-spark2_2.10:0.19.1-SNAPSHOT (/home/jenkins/workspace/ADAM-prb/HADOOP_VERSION/2.3.0/SCALAVER/2.10/SPARK_VERSION/2.0.0/label/centos/adam-cli/pom.xml) has 9 errors
[ERROR]     'dependencies.dependency.version' for org.bdgenomics.utils:utils-misc-spark2_2.10:test-jar is missing. @ line 116, column 17
[ERROR]     'dependencies.dependency.version' for org.bdgenomics.utils:utils-misc-spark2_2.10:jar is missing. @ line 122, column 17
[ERROR]     'dependencies.dependency.version' for org.bdgenomics.utils:utils-io-spark2_2.10:jar is missing. @ line 126, column 17
[ERROR]     'dependencies.dependency.version' for org.bdgenomics.utils:utils-cli-spark2_2.10:jar is missing. @ line 130, column 17
[ERROR]     'dependencies.dependency.version' for org.bdgenomics.utils:utils-metrics-spark2_2.10:jar is missing. @ line 134, column 17
[ERROR]     'dependencies.dependency.version' for org.bdgenomics.adam:adam-core-spark2_2.10:jar is missing. @ line 142, column 17
[ERROR]     'dependencies.dependency.version' for org.bdgenomics.adam:adam-core-spark2_2.10:test-jar is missing. @ line 146, column 17
[ERROR]     'dependencies.dependency.version' for org.bdgenomics.adam:adam-apis-spark2_2.10:jar is missing. @ line 152, column 17
[ERROR]     'dependencies.dependency.version' for org.bdgenomics.adam:adam-apis-spark2_2.10:test-jar is missing. @ line 156, column 17
[ERROR]   
[ERROR]   The project org.bdgenomics.adam:adam-assembly-spark2_2.10:0.19.1-SNAPSHOT (/home/jenkins/workspace/ADAM-prb/HADOOP_VERSION/2.3.0/SCALAVER/2.10/SPARK_VERSION/2.0.0/label/centos/adam-assembly/pom.xml) has 1 error
[ERROR]     'dependencies.dependency.version' for org.bdgenomics.adam:adam-cli-spark2_2.10:jar is missing. @ line 115, column 17
```

Any ideas?

heuermh:
The version for those dependencies should be 0.2.8, which is on Maven Central.  Are those variable substitutions coming from a profile?  We had trouble with that before.

pom.xml
```diff
@@ -282,9 +282,9 @@
           </jvmArgs>
           <javacArgs>
             <javacArg>-source</javacArg>
-            <javacArg>${java.version}</javacArg>
+            <javacArg>1.7</javacArg>
```
heuermh: we require jdk8 because of the transitive dependency on htsjdk through Hadoop-BAM

pom.xml
```diff
@@ -19,8 +19,8 @@
   <properties>
     <java.version>1.8</java.version>
     <avro.version>1.8.0</avro.version>
-    <!-- informative only, used in about template substitution -->
-    <scala.version>2.10</scala.version>
+    <scala.version>2.10.6</scala.version>
+    <scala.version.prefix>2.10</scala.version.prefix>
```
heuermh: changing this also requires a change to https://github.com/bigdatagenomics/adam/blob/master/adam-cli/src/main/java-templates/org/bdgenomics/adam/cli/About.java#L28

pom.xml
```diff
@@ -334,40 +334,40 @@
       <dependency>
         <groupId>org.scala-lang</groupId>
         <artifactId>scala-library</artifactId>
-        <version>2.10.4</version>
+        <version>${scala.version}</version>
       </dependency>
       <dependency>
         <groupId>org.bdgenomics.adam</groupId>
-        <artifactId>adam-core_2.10</artifactId>
+        <artifactId>adam-core_${scala.version.prefix}</artifactId>
```
heuermh: I'm fairly certain there's a reason why we can't do this, I know because I tried to do it once.  Of course, I don't remember why it didn't work . . .

heuermh: https://github.com/bigdatagenomics/utils/pull/69

ryan-williams:
> The version for those dependencies should be 0.2.8, which is on Maven Central. Are those variable substitutions coming from a profile? We had trouble with that before.

Those errors are modules' `<dependency>`s not declaring `<version>`s (because they should pick them up from the parent POM), but something about that resolution breaking down and Maven thinking they are just missing a `<version>` specification.

pom.xml
ryan-williams: Interesting; what is the contract with that file? Every Maven property must be declared there? Why?

ryan-williams: OK, I looked around this a bit but don't see what change I need to make; help?

The templating-maven-plugin docs say it's useful "if you need for example to have things in that code replaced with some properties values."

`${scala.version}` is already exposed there (though not used anywhere in the project afaict) and I have no use for exposing `${scala.version.prefix}` there, if that's what you mean?

fnothaft: I think it is OK (preferable, really) to have `About` print the full Scala version, as opposed to the prefix.

pom.xml
fnothaft: It's OK to use the properties in the `artifactId` for _dependencies_, it doesn't work to use properties in the `artifactId` for the POM itself. I know we've gone back and forth on this a lot, but I think it's actually preferable to do what @ryan-williams is doing here. I'll make a comment at the end with my reasoning.

pom.xml
fnothaft: +1, needs to be changed. Although, is there a reason to move away from the `${java.version}` property here, if we're generally moving back towards using the parameters throughout the POM? I would prefer going back to `${java.version}` on this line.

scripts/jenkins-test
```diff
@@ -7,13 +7,15 @@ ADAM_MVN_TMP_DIR=$(mktemp -d -t adamTestMvnXXXXXXX)
...
 find . -name pom.xml \
-    -exec sed -i .bak \
+    -exec sed -i.bak \
     -e "s:sun.io.serialization.extendedDebugInfo=true:sun.io.serialization.extendedDebugInfo=true -Djava.io.tmpdir=${ADAM_MVN_TMP_DIR}:g" \
     {} \;
 find . -name "*.bak" -exec rm {} \;
 
 # variable declarations
-export JAVA_HOME=/usr/java/jdk1.8.0_60
+if [ -z "$JAVA_HOME" ]; then
```
fnothaft: This is a bit ugly, but we always need to overwrite `$JAVA_HOME` here due to Jenkins doing "shenanigans" with the Java version.  The short story is that Jenkins will always set `$JAVA_HOME` to a bad value, and we are powerless to stop it. There's a much longer and uninteresting backstory. Can you revert this back, or add some other logic to disable overwriting `$JAVA_HOME`?

pom.xml
ryan-williams: Ah that makes sense @heuermh, thanks for the explanation. Sounds like in the end we can just leave as is though, lmk if otherwise.

pom.xml
ryan-williams: Sorry, I got cornered into doing this for a reason that I didn't have paged in and meant to go back and try to reconstruct what that was.

I'm pretty sure it was related to my making the build use `zinc`; for some reason `1.8` was not acceptable in that setting, and I had to hard-code `1.7` here. I'll try to dig back and get the full state on it for posterity, but in the meantime I was going to give up on `zinc` in this PR anyway, since I didn't know how to get around the "`sbt-interface.jar` not found" problem on Jenkins discussed elsewhere on this PR (I believe due to `zinc` not being installed on Jenkins in some way), so I'll revert this bit momentarily. Thanks!

scripts/jenkins-test
ryan-williams: Cool! I'll remove it here then altogether, 1m.

ryan-williams:
OK, I reverted the `java.version` bit and removed the `jenkins-test` `JAVA_HOME`-handling; I don't know why the build would avoid [the failure above](https://github.com/bigdatagenomics/adam/pull/1142#issuecomment-251500908) this time around, so am sort of expecting that to happen again, and suspect that it is related to the `scala.version.prefix` usage throughout the POMs, so let's see what happens there.

ryan-williams:
Cool, so it turns out I just hadn't updated the `sed` regexs in `move_to_spark2` to properly handle lines that now have `_${scala.version.prefix}` instead of `_2.10`.

I am not sure what to do about this "**fnothaft** requested changes" error I'm still seeing.

fnothaft:
LGTM! Unless anyone has objections, I will squash this down and merge in the AM.

fnothaft:
Thanks @ryan-williams! I've merged this manually as ad50f7d95e4f88fc2b242fc7c0c37beb8d4729fe, e8b0fcdd3feba937bb66cd72bbbdcad86039eeb1, cd488dfd1ace53bc3aec0e694def59b8ab2c1344, d242a265b60e1ab39d20b71b37eb14c5c704d054, a0730f36655d8fbd5359d84b5564c83c90ff7cf5, and 89942fee598426c12d5fd6531c2fd25d1f40b4ea. As always, thank you for taking the time to make a nice cleanup pass! It is greatly appreciated.

PR #1147, opened 2016-09-02
Updated GenomicPartitioners to accept additional key.

akmorrow13:
GenomicRegionPartitioner and GenomicPositionPartitioner can now accept an extra key. This can be used for keyBy() on cell type, sampleId, etc.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicPartitioners.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicPartitioners.scala
@@ -67,6 +67,13 @@ case class GenomicPositionPartitioner(numParts: Int, seqLengths: Map[String, Lon
     // This allows partitions that cross chromosome boundaries.
     // The computation is slightly more complicated if you want to avoid this.
     def getPart(referenceName: String, pos: Long): Int = {
+      require(
+        seqLengths.contains(referenceName),
+        "Received key (%s) that did not map to a known contig. Contigs are:\n%s".format(
+          referenceName,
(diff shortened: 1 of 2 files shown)


adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicPartitioners.scala
```diff
@@ -89,6 +89,16 @@ case class GenomicPositionPartitioner(numParts: Int, seqLengths: Map[String, Lon
         )
         getPart(refpos.referenceName, refpos.pos)
       }
+      case (refpos: ReferencePosition, k: Any) => {
+        require(
```
akmorrow13: There has got to be a cleaner way to do this. How can I piggyback these cases?
case (refpos: ReferencePosition) | (refpos: ReferencePosition, k: Any) throws an error because refpos cannot be defined twice...

fnothaft: I think having them as separate cases is OK. I might pull the `require(...)` into the `getPart`, but this doesn't "smell" too bad to me.

fnothaft:
LGTM, other than the nit RE: pulling things out into the helper function. I'm excited to have this in, @akmorrow13!

heuermh:
+1

fnothaft:
@akmorrow13 do you have the time to make the update that I referred to above, or would you prefer that I take that update on?

fnothaft:
Merged! Thanks @akmorrow13!

PR #1159, opened 2016-09-09
Map IntervalList format column four to feature name

### LINKED ISSUE #1152: Map IntervalList format column four to feature name or attributes?
We are not sure if the name column should be mapped to `Feature.name` or to `Feature.attributes`.

Note discussion here
https://github.com/bigdatagenomics/adam/pull/1104/files#r77733324
https://github.com/bigdatagenomics/adam/pull/1104/files#r74367312

Ping @ryan-williams to comment.

heuermh:
Fixes #1152, #1168

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
@@ -27,20 +27,36 @@ import scala.collection.mutable.ArrayBuffer
 import scala.io.Source
 
 private[rdd] sealed trait FeatureParser extends Serializable with Logging {
-  def parse(line: String): Seq[Feature]
-}
 
-private[rdd] class FeatureFile(parser: FeatureParser) extends Serializable {
-  def parse(file: File): Iterator[Feature] = {
-    val src = Source.fromFile(file)
-    try {
-      src.getLines().flatMap { line =>
-        parser.parse(line)
-      }
-    } finally {
-      src.close()
+  /**
+   * If the specified validation strategy is STRICT, throw an exception,
+   * if LENIENT, log a warning, otherwise return None.
+   *
+   * @param message error or warning message
+   * @param line line
+   * @param stringency validation stringency
+   * @return None, if stringency is not STRICT or LENIENT
+   */
+  def throwWarnOrNone(message: String,
+                      line: String,
+                      stringency: ValidationStringency): Option[Feature] = {
+
+    if (stringency == ValidationStringency.STRICT) {
+      throw new IllegalArgumentException(message.format(line))
+    } else if (stringency == ValidationStringency.LENIENT) {
+      log.warn(message.format(line))
     }
+    None
   }
+
+  /**
+   * Parse the specified line and return a feature, if possible.
+   *
+   * @param line line to parse
+   * @param stringency validation stringency
+   * @return the specified line parsed into a feature, if possible
+   */
+  def parse(line: String, stringency: ValidationStringency): Option[Feature]
 }
 
 private[rdd] object GTFParser {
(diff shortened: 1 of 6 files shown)


adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
```diff
@@ -67,15 +54,28 @@ private[rdd] object GTFParser {
...
+    }
 
     val fields = line.split("\t")
-    // skip empty or invalid lines
+
+    // check for invalid lines
     if (fields.length < 8 || fields.length > 9) {
-      log.warn("Empty or invalid GTF/GFF2 line: {}", line)
-      return Seq()
+      if (stringency == ValidationStringency.STRICT) {
```
fnothaft: As an aside, I feel like we've got enough of this patten (print error message if STRICT, log if LENIENT, and return None if not STRICT) around where we might as well wrap it up into a function and factor it out.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
```diff
@@ -327,35 +338,44 @@ private[rdd] class BEDParser extends FeatureParser {
...
-    val fb = Feature.newBuilder()
-    fb.setContigName(fields(0))
 
-    // Peak files are 0-based space-coordinates, so conversion to
-    // our coordinate space should mean that the values are unchanged.
-    fb.setStart(fields(1).toLong)
-    fb.setEnd(fields(2).toLong)
+    val f = Feature.newBuilder()
+      .setContigName(fields(0))
+      .setStart(fields(1).toLong) // NarrowPeak ranges are 0-based
```
fnothaft: We should try/catch around the `fields(x).toLong`/`fields(x).toDouble` calls, both here and below and above and everywhere. We could factor it out like so:

```
def checkAndSet[T](field: String,
  convFn: String => T,
  setFn: T => Unit,
  stringency: ValidationStringency) {
  try {
    val t = convFn(field)
    setFn(t)
  } catch {
    case e: Throwable => {
      if (stringency == ValidationStringency.LENIENT) {
        log.warn("Failed to convert %s.".format(field))
      } else if (stringency == ValidationStringency.STRICT) {
        throw new IllegalArgumentException("Setting field from %s failed with %s.".format(
          field,
          e.msg))
      }
    }
  }
}
```

Perhaps we factor this out into a trait:

```
sealed trait FeatureConverterCanBeValidated {
  val stringency: ValidationStringency
  
  def validateOrNone(...): Option[Feature]

  def checkAndSet[T](field: String,
    convFn: String => T,
    setFn: T => Unit)
}
```

All of the classes in here could be turned into case classes that extend that trait. What do you think? Should you proceed, the trait name I proposed is terrible, so please change it.

fnothaft:
Generally LGTM! Thanks for taking this on @heuermh. I've dropped a variety of line notes inline.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
heuermh: Agreed

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
heuermh: Will check to see what htsjdk does with NumberFormatExceptions and validation stringency.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
fnothaft: How so?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/features/FeatureParser.scala
heuermh: This seems to fiddly to me, so I've punted for now

heuermh:
Addressed some review comments and rebased.

fnothaft:
LGTM! I will leave this open until Friday morning to wait for any further review comments.

fnothaft:
Thanks @heuermh! Merged as 53b8f48abe0b0536c39e155fb32de91a98c78876 and dff43effd9f52f8cdcec8952a0a27c5c258afd45.

heuermh:
Thanks!

PR #1180, opened 2016-09-20
[ADAM-1179] Improve error message when globbing a parquet file fails.

### LINKED ISSUE #1179: How to read glob of multiple parquet Genotype
Is it possible to load the data of several parquet files storing bdg-formats Genotype into a single GenotypeRDD?
I want something equivalent to: sc.loadVcf("*.vcf") typeglob but for ADAM Parquet files.

See discussions below from email:

On Tue, Sep 20, 2016 at 11:19 AM, Justin Paschall <justinpaschalldna@gmail.com> wrote:
For set of single samples VCFs Michael pointed out using  sc.loadVcf("*.vcf") can read in a bunch of single samples VCFs into a single RDD[VariantContext] of a VariantContextRDD

Similarly - is there any sense to to converting single-sample VCFs to single sample Parquet files of [Genotype]  as a data source of reference?

However, I am not sure there is a way to typeglob load a set of many individual ADAM parquet genotypes files upon loading, in the same way that we can load a bunch of VCF with the typeglob?

It is tricky to get the globs in correctly without having them expanded by the shell; it appears that even my tricks don't work for .adam files

$ ./bin/adam-submit transform "*.adam" -single combined.sam
Command body threw exception:
java.io.FileNotFoundException: Couldn't find any files matching *.adam
Exception in thread "main" java.io.FileNotFoundException: Couldn't find any files matching *.adam
    at org.bdgenomics.adam.rdd.ADAMContext.getFsAndFilesWithFilter(ADAMContext.scala:391)
    at org.bdgenomics.adam.rdd.ADAMContext.loadAvroSequences(ADAMContext.scala:207)
    at org.bdgenomics.adam.rdd.ADAMContext.loadParquetAlignments(ADAMContext.scala:637)

$ ./bin/adam-submit transform "file:///`pwd`/*.adam" -single combined.sam
Command body threw exception:
java.io.FileNotFoundException: Couldn't find any files matching file:///Users/heuermh/working/adam/*.adam
Exception in thread "main" java.io.FileNotFoundException: Couldn't find any files matching file:///Users/heuermh/working/adam/*.adam
    at org.bdgenomics.adam.rdd.ADAMContext.getFsAndFilesWithFilter(ADAMContext.scala:391)
    at org.bdgenomics.adam.rdd.ADAMContext.loadAvroSequences(ADAMContext.scala:207)
    at org.bdgenomics.adam.rdd.ADAMContext.loadParquetAlignments(ADAMContext.scala:637)

Will have to dig in the new code changes to look (getFsAndFilesWithFilter).

 
The VCF type glob works, but it is slower than I hoped, 10 minutes to load/count on 100 samples of chr22 on my workstation (testing on cluster shortly), and I want to make sure that I am not missing out on a prior one-time conversion to parquet on a per-sample-file basis if that would be better.

fnothaft:
Resolves #1179.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
@@ -306,7 +306,10 @@ class ADAMContext private (@transient val sc: SparkContext) extends Serializable
     // the path must match at least one file
     if (paths.isEmpty) {
       throw new FileNotFoundException(
-        s"Couldn't find any files matching ${path.toUri}"
+        s"Couldn't find any files matching ${path.toUri}. If you are trying to" +
+          " glob a directory of Parquet files, you need to glob inside the" +
+          " directory as well (e.g., \"glob.me.*.adam/*\", instead of" +
+          " \"glob.me.*.adam\"."
       )
     }
 


heuermh:
Merged as commit 8bd6c0e.

Thank you, @fnothaft!

PR #1181, opened 2016-09-21
Add scoverage aggregator and fail on low coverage.

fnothaft:
Small Scoverage update --> sets aggregate to true, which produces a report under `${ADAM_HOME}/target` at the end of the build containing the aggregated coverage across all submodules.


heuermh:
Merged as commit d914ee4aa023440ff52cef9dd72daa5ea47ed558.

Thank you, @fnothaft!

PR #1183, opened 2016-09-26
[ADAM-1182] isSorted=true should write SO:coordinate in SAM/BAM/CRAM header.

### LINKED ISSUE #1182: How to get "SO coordinate" instead of "SO unsorted" ?
image

When I save data to sam file,I can only get "SO unsorted"  sort_tag in the result file.
Can anyone help me ?
Thx!

fnothaft:
Resolves #1182. Moved to use SAMTextHeaderCodec + SAMFileWriter.writeHeader to
write the header, vs. relying on SAMFileWriter's internal header writing code.


adam-core/src/test/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDDSuite.scala
```diff
@@ -337,6 +337,7 @@ class AlignmentRecordRDDSuite extends ADAMFunSuite {
     val reads = ardd.rdd
 
     val actualSortedPath = tmpFile("sorted.sam")
+    println(actualSortedPath)
```
heuermh: remove this

fnothaft: Sorry, left that in when debugging. Fixed!

heuermh:
Merged via commit 58549b273dea3280ce9ede2ab29f8cd0ebc1090b

PR #1185, opened 2016-09-28
Fastq record converter

zyxue:
There are still problems:

* The description for `convertRead` does don't match with what it does (e.g. the return type). The method name is kind of vague, not sure what behavior is actually intended. The refactoring tries not to change its behavior.
* I am also a bit confused for what default values should be used when creating a AlignmentRecord instance based on a Fastq entry. The original code set many fields to `null`. I thought leaving them as default would make more sense. e.g. `ReadNegativeStrand` is set to `null` while the [default](https://github.com/bigdatagenomics/bdg-formats/blob/master/src/main/resources/avro/bdg.avdl) is `false`. `ProperPair` is set to `true`, which I don't think make sense since you can't really tell just based on Fastq, the default is also `fasle`. It would be helpful if someone can clarify what's the most sensible values for the following fields for both paired-end and single-end fastq entries.
```
.setReadPaired(readPaired)
.setReadInFragment(readInFragment)
.setReadNegativeStrand(null)
.setMateNegativeStrand(null)
.setPrimaryAlignment(null)
.setSecondaryAlignment(null)
.setSupplementaryAlignment(null)
```

* Also, what to do when the read length and that of qualities don't match? There is a stringency parameter involved in `convertRead`. If it's STRICT, then qualities must exist (CANNOT be `*`) and also match the length of reads. When it's NOT STRICT, qualities will be padded with `B` if not exist or shorter than read length. If it's longer than read length, `NotImplementedError` will be thrown. Such behavior seems quite arbitrary and doesn't make much sense to me, and it doesn't apply to `convertPair` and `convertFragment`, in which the read length and qualities must match without consideration for stringency.


heuermh:
Jenkins, test this please.

adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
```diff
@@ -41,6 +41,102 @@ import scala.collection.JavaConversions._
...
+      readSequence.length == readQualities.length,
+      s"The first read: ${readName}, has different sequence and qual length."
+    )
+
+    (readNameNoSuffix, readSequence, readQualities)
+  }
+
+  private def parseReadPairInFastq(input: String): (String, String, String, String, String, String) = {
+    val lines = input.toString.split('\n')
+    require(lines.length == 8,
```
heuermh: I've mentioned this before; perhaps now is the time to fix it?  FASTQ format allows for hard line wrapping, so there may be new line characters at any place in the record.

adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
```diff
@@ -41,6 +41,102 @@ import scala.collection.JavaConversions._
...
+  private def parseReadInFastq(input: String,
+                               setFirstOfPair: Boolean = false,
+                               setSecondOfPair: Boolean = false,
+                               stringency: ValidationStringency = ValidationStringency.STRICT): (String, String, String) = {
+    val lines = input.split('\n')
+    require(lines.length == 4,
+      s"Input must have 4 lines (${lines.length.toString} found):\n${input}")
+
+    val readName = lines(0).drop(1)
+    if (readName.endsWith("/1") && setSecondOfPair)
```
heuermh: I've seen files in the wild that use ` 1` and ` 2` (with space) instead of `/1` and `/2`.  Could we add that here?
See e.g. [PairedEndFastqReader.java#L59](https://github.com/heuermh/dishevelled-bio/blob/master/read/src/main/java/org/dishevelled/bio/read/PairedEndFastqReader.java#L59)

adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
```diff
@@ -41,6 +41,102 @@ import scala.collection.JavaConversions._
...
+
+    val readSequence = lines(1)
+    val readQualitiesRaw = lines(3)
+
+    val readQualities =
+      if (stringency == ValidationStringency.STRICT) readQualitiesRaw
+      else {
+        if (readQualitiesRaw == "*") "B" * readSequence.length
+        else if (readQualitiesRaw.length < readSequence.length) readQualitiesRaw + ("B" * (readSequence.length - readQualitiesRaw.length))
+        else if (readQualitiesRaw.length > readSequence.length) throw new NotImplementedError("Not implemented")
```
heuermh: `NotImplementedError` doesn't seem right, should be `IllegalArgumentException`.  These length checks should also happen with strict stringency.

adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
```diff
@@ -41,6 +41,102 @@ import scala.collection.JavaConversions._
...
+                                  qual: String,
+                                  readInFragment: Int,
+                                  readPaired: Boolean = true,
+                                  recordGroupOpt: Option[String] = None): AlignmentRecord = {
+    val builder = AlignmentRecord.newBuilder
+      .setReadName(readName)
+      .setSequence(sequence)
+      .setQual(qual)
+      .setReadPaired(readPaired)
+      .setProperPair(null)
```
heuermh: I'm not sure why these are explicitly set to null.

adam-core/src/test/scala/org/bdgenomics/adam/converters/FastqConverterSuite.scala
```diff
@@ -0,0 +1,142 @@
...
+ * limitations under the License.
+ */
+package org.bdgenomics.adam.converters
+
+import htsjdk.samtools.ValidationStringency
+import org.apache.hadoop.io.Text
+import org.scalatest.FunSuite
+
+/**
+ * Created by zyxue on 2016-09-27.
```
heuermh: Unit test suite for FastqRecordConverter.

adam-core/src/test/scala/org/bdgenomics/adam/converters/FastqConverterSuite.scala
```diff
@@ -0,0 +1,142 @@
...
+ */
+class FastqConverterSuite extends FunSuite {
```
heuermh: Should be `FastqRecordConverterSuite`.

heuermh:
Thanks @zyxue!  I've left some review comments but they may not completely answer your questions.  Hopefully others will chime in.

adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
zyxue: Can you give an example for hard line wrapping? What does it look like?

adam-core/src/test/scala/org/bdgenomics/adam/converters/FastqConverterSuite.scala
zyxue: Are you sure? I followed convention on [FastaConverterSuite](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/test/scala/org/bdgenomics/adam/converters/FastaConverterSuite.scala)

adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
zyxue: I have seen both, as well. I am not aware if there is a specification that lists all possibilities. I am thinking of using regex to account for all of them gradually.

adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
zyxue: Also, what's the justification for padding `B` in case qualities information is incomplete?

adam-core/src/test/scala/org/bdgenomics/adam/converters/FastqConverterSuite.scala
zyxue: OK, you're right. I realized it's

`FastaConverter.scala` & `FastaConverterSuite`

so it should be 

`FastqRecordConverter` & `FastqRecordConverterSuite`

adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
heuermh: There are all kinds of correct and incorrect examples here https://github.com/biojava/biojava/tree/master/biojava-sequencing/src/test/resources/org/biojava/nbio/sequencing/io/fastq

adam-core/src/test/scala/org/bdgenomics/adam/converters/FastqConverterSuite.scala
heuermh: Yep, although I find the word "record" redundant almost everywhere, so we could possibly drop it here

zyxue: I will stick to the original convention for now.

adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
heuermh: There isn't a specification, only convention.  See http://dx.doi.org/10.1093/nar/gkp1137

zyxue:
I have located the reason at least for the [first test failure](https://amplab.cs.berkeley.edu/jenkins//job/ADAM-prb/1505/HADOOP_VERSION=2.6.0,SCALAVER=2.10,SPARK_VERSION=1.5.2,label=centos/testReport/org.bdgenomics.adam.rdd.read/AlignmentRecordRDDSuite/round_trip_from_ADAM_to_FASTQ_and_back_to_ADAM_produces_equivalent_Read_values/). As mentioned, it's due to inconsistent default values. In the original [`convertRead`](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala#L259), it's`setReadNegativeStrand(false)`, setting it to null, which is done in [`convertPair`](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala#L61) and [`convertFragment`](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala#L128), would cause `java.lang.NullPointerException` when [`getReadNegativeStrand `](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/converters/AlignmentRecordConverter.scala#L90). I suggest dropping all the explicitly set null and use default values instead, what do you think?

adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
zyxue: explicitly setting null is dropped in newer commits. If you rerun the tests, it should all pass.

adam-core/src/test/scala/org/bdgenomics/adam/converters/FastqConverterSuite.scala
zyxue: What does this comment mean?

heuermh: Sorry, am suggesting a doc comment change.

zyxue: ok, removed the unnecessary comment.

zyxue:
Is it easy to rerun all tests on jenkins?

adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
```diff
@@ -41,6 +41,96 @@ import scala.collection.JavaConversions._
 private[adam] class FastqRecordConverter extends Serializable with Logging {
 
   /**
+   * Parse 4 lines at a time
+   * @see parseReadPairInFastq
+   * *
+   */
+  private def parseReadInFastq(input: String,
```
zyxue: Is the use of `private` like this a good practice? It will be used in other methods in `FastqRecordConverter` via `this.parseReadInFastq`

heuermh:
> Is it easy to rerun all tests on jenkins?

Jenkins should re-run when you push new commits. If not, let me know when you've pushed everything you want, and I'll explicitly ask for a retest.

`./scripts/jenkins-test` lets you run the Jenkins tests locally, although you may have to mess with some path definitions.

heuermh:
Jenkins, retest this please.

zyxue:
The failed setup is limited to hadoop-2.6.0, scala-2.10 and spark-1.5.2. Looking at the [failed test log](https://amplab.cs.berkeley.edu/jenkins/job/ADAM-prb/1507/HADOOP_VERSION=2.6.0,SCALAVER=2.10,SPARK_VERSION=1.5.2,label=centos/consoleFull). I am not sure how it fails. On one hand, it says all tests passed.

```
Run completed in 1 minute, 10 seconds.
Total number of tests run: 457
Suites: completed 52, aborted 0
Tests: succeeded 457, failed 0, canceled 0, ignored 2, pending 0
All tests passed.
```

On the other hand, it stills fails.

```
[INFO] ------------------------------------------------------------------------
[INFO] BUILD SUCCESS
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 6.679s
[INFO] Finished at: Wed Sep 28 14:01:36 PDT 2016
[INFO] Final Memory: 20M/1153M
[INFO] ------------------------------------------------------------------------
+ popd
~/workspace/ADAM-prb/HADOOP_VERSION/2.6.0/SCALAVER/2.10/SPARK_VERSION/1.5.2/label/centos
++ git status --porcelain
+ test -n ' M adam-core/src/test/scala/org/bdgenomics/adam/converters/FastqRecordConverterSuite.scala'
+ echo 'Please run '\''./scripts/format-source'\'''
Please run './scripts/format-source'
+ exit 1
Build step 'Execute shell' marked build as failure
Recording test results
Publishing Scoverage XML and HTML report...
Setting commit status on GitHub for https://github.com/bigdatagenomics/adam/commit/7c99373154de4939570142573401cb93d93874dd
Finished: FAILURE
```
 @heuermh, do you know where is the problem? It's not obvious to me yet. I tried to run `'./scripts/format-source'`, but it doesn't seem to make any difference.

heuermh:
Hmm...  I'm not sure.  It appears that Jenkins is complaining about source formatting, but after checking out your branch, as you say `format-source` does not make any changes.

When I run `scripts/jenkins-test` locally, it fails with
```bash
[INFO] ------------------------------------------------------------------------
[INFO] Building ADAM_2.10: Core 0.19.1-SNAPSHOT
[INFO] ------------------------------------------------------------------------
...
[INFO] --- scalatest-maven-plugin:1.0:test (test) @ adam-core_2.10 ---
Discovery starting.
*** RUN ABORTED ***
  java.lang.RuntimeException: Unable to load a Suite class that was discovered in the runpath: org.bdgenomics.adam.rdd.variation.VariantContextRDDSuite
  at org.scalatest.tools.DiscoverySuite$.getSuiteInstance(DiscoverySuite.scala:84)
  at org.scalatest.tools.DiscoverySuite$$anonfun$1.apply(DiscoverySuite.scala:38)
  at org.scalatest.tools.DiscoverySuite$$anonfun$1.apply(DiscoverySuite.scala:37)
  at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
  at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
  at scala.collection.Iterator$class.foreach(Iterator.scala:727)
  at scala.collection.AbstractIterator.foreach(Iterator.scala:1157)
  at scala.collection.IterableLike$class.foreach(IterableLike.scala:72)
  at scala.collection.AbstractIterable.foreach(Iterable.scala:54)
  at scala.collection.TraversableLike$class.map(TraversableLike.scala:244)
  at scala.collection.AbstractTraversable.map(Traversable.scala:105)
  at org.scalatest.tools.DiscoverySuite.<init>(DiscoverySuite.scala:37)
  at org.scalatest.tools.Runner$.genDiscoSuites$1(Runner.scala:2390)
  at org.scalatest.tools.Runner$.doRunRunRunDaDoRunRun(Runner.scala:2475)
  at org.scalatest.tools.Runner$$anonfun$runOptionallyWithPassFailReporter$2.apply(Runner.scala:1044)
  at org.scalatest.tools.Runner$$anonfun$runOptionallyWithPassFailReporter$2.apply(Runner.scala:1043)
  at org.scalatest.tools.Runner$.withClassLoaderAndDispatchReporter(Runner.scala:2722)
  at org.scalatest.tools.Runner$.runOptionallyWithPassFailReporter(Runner.scala:1043)
  at org.scalatest.tools.Runner$.main(Runner.scala:860)
  at org.scalatest.tools.Runner.main(Runner.scala)
  Cause: java.lang.IllegalStateException: Failed to create directory within 10000 attempts (tried 1475104108290-0 to 1475104108290-9999)
  at com.google.common.io.Files.createTempDir(Files.java:600)
  at org.bdgenomics.adam.rdd.variation.VariantContextRDDSuite.<init>(VariantContextRDDSuite.scala:35)
  at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
  at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
  at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
  at java.lang.reflect.Constructor.newInstance(Constructor.java:422)
  at java.lang.Class.newInstance(Class.java:442)
  at org.scalatest.tools.DiscoverySuite$.getSuiteInstance(DiscoverySuite.scala:69)
  at org.scalatest.tools.DiscoverySuite$$anonfun$1.apply(DiscoverySuite.scala:38)
  at org.scalatest.tools.DiscoverySuite$$anonfun$1.apply(DiscoverySuite.scala:37)
  at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
  at scala.collection.TraversableLike$$anonfun$map$1.apply(TraversableLike.scala:244)
  at scala.collection.Iterator$class.foreach(Iterator.scala:727)
  at scala.collection.AbstractIterator.foreach(Iterator.scala:1157)
  at scala.collection.IterableLike$class.foreach(IterableLike.scala:72)
  at scala.collection.AbstractIterable.foreach(Iterable.scala:54)
  at scala.collection.TraversableLike$class.map(TraversableLike.scala:244)
  at scala.collection.AbstractTraversable.map(Traversable.scala:105)
  at org.scalatest.tools.DiscoverySuite.<init>(DiscoverySuite.scala:37)
  at org.scalatest.tools.Runner$.genDiscoSuites$1(Runner.scala:2390)
  at org.scalatest.tools.Runner$.doRunRunRunDaDoRunRun(Runner.scala:2475)
  at org.scalatest.tools.Runner$$anonfun$runOptionallyWithPassFailReporter$2.apply(Runner.scala:1044)
  at org.scalatest.tools.Runner$$anonfun$runOptionallyWithPassFailReporter$2.apply(Runner.scala:1043)
  at org.scalatest.tools.Runner$.withClassLoaderAndDispatchReporter(Runner.scala:2722)
  at org.scalatest.tools.Runner$.runOptionallyWithPassFailReporter(Runner.scala:1043)
  at org.scalatest.tools.Runner$.main(Runner.scala:860)
  at org.scalatest.tools.Runner.main(Runner.scala)
```

Which doesn't seem like the same problem (an OSX thing, perhaps?).

heuermh:
Jenkins, retest this please.

zyxue:
The most [recent test failure](https://amplab.cs.berkeley.edu/jenkins/job/ADAM-prb/1508/HADOOP_VERSION=2.6.0,SCALAVER=2.11,SPARK_VERSION=1.5.2,label=centos/console) is complaining about being unable to create directory

```
*** RUN ABORTED ***
  java.lang.RuntimeException: Unable to load a Suite class that was discovered in the runpath: org.bdgenomics.adam.rdd.variation.VariantContextRDDSuite
  at org.scalatest.tools.DiscoverySuite$.getSuiteInstance(DiscoverySuite.scala:84)
 ...
  at org.scalatest.tools.Runner$.main(Runner.scala:860)
  at org.scalatest.tools.Runner.main(Runner.scala)
  Cause: java.lang.IllegalStateException: Failed to create directory within 10000 attempts (tried 1475104460351-0 to 1475104460351-9999)

**Comment by heuermh** [comment]
Yeah same as I saw locally.

**Comment by zyxue** [comment]
What do you think could be the reason? It's interesting that it only happens to a small set of the configuration matrix.

**Comment by heuermh** [comment]
The way Jenkins is set up, it only does the whole matrix if the "primary" builds fail, which are Spark 1.5.2 Hadoop 2.6.0 Scala 2.10 and Scala 2.10.  So unfortunately the green circles for all the other build targets aren't real in this case.

**Comment by zyxue** [comment]
I think the error is due to unintended change of 3 `pom.xml` files, I have reverted the changes.

**Comment by heuermh** [comment]
Jenkins, retest this please.

**Comment by heuermh** [comment]
Looks like that was it, tests pass now.  I'll give this another round of review tomorrow, and hopefully we can get another set of eyes on it.

**Comment by zyxue** [comment]
Passing all current tests is one thing. I read your paper, http://dx.doi.org/10.1093/nar/gkp1137, and I think it would be a good idea to add more tests for irregular FASTQ formats. My confusion in the original code from that it's not obvious what the original author intended to do in certain cases. I guess I will just decide the scope accordingly, and do some more refactoring later.

**Comment by zyxue** [comment]
@heuermh, have you tested a private function before? I am trying to test [`parseReadInFastq `](https://github.com/zyxue/adam/blob/FastqRecordConverter/adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala#L48), directly. Not sure how to do so. Or do you have any suggestion, please?

**Comment by fnothaft** [comment]
Hi @zyxue! Thank you so much for taking this on; I'm really excited to see this PR and will take a closer look at the code later today.

> @heuermh, have you tested a private function before? I am trying to test parseReadInFastq, directly. Not sure how to do so. Or do you have any suggestion, please? I followed the last comment in this post, but still haven't made it work yet. I am wondering if I used private properly.

What I typically do is relax the protection from private to package private. E.g., in this case, I'd change:

```
private def parseReadInFastq(input: String,
```

to:

```
private[converters] def parseReadInFastq(input: String,
```

Then, you can call the function from the test suite, since the test suite is in the same package. Ideally, we'd be able to keep the function private, but (IIRC) the only way to keep the function fully private and to call it from test code is through the Java reflection API, which is a real hassle. I think this is a reasonable compromise.

**Comment by zyxue** [comment]
Cool, @fnothaft. It looks that I can access the method now. I am not familiar with Java reflection API, but in Scalatest, there is a [`PrivateMethodTester`](http://www.scalatest.org/user_guide/other_goodies#privateMethodTester) trait that can be used to test a private method. It's also via reflection, but it also looks quite a bit of hassle, too. I tried it, but haven't made it work. I prefer your solution for now. Thanks!

**Comment by fnothaft** [comment]
I just made a review pass, @zyxue. I've highlighted a few small syntax nits, but overall, this looks like a very solid PR; thank you for the great contribution!

**Comment by heuermh** [comment]
@fnothaft I don't see any of your review comments.  Were they lost or resolved?

File: adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala (hunk 3 above)
Comment (zyxue) [inline adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala]: https://en.wikipedia.org/wiki/FASTQ_format
>Also, in Illumina runs using PhiX controls, the character 'B' was observed to represent an "unknown quality score". The error rate of 'B' reads was roughly 3 phred scores lower the mean observed score of a given run.

**Comment by zyxue** [comment]
Where can I find more information on `ValidationStringency.STRICT`, what should the behavior be when it's STRICT or not, please?

File: adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala (hunk 2 above)
Comment (zyxue) [inline adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala]: addressed by regex like `[/ +_]1$`

File: adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala (hunk 3 above)
Comment (zyxue) [inline adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala]: fixed `NotImplementedError` => `IllegalArgumentException`

File: adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala [hunk 8]
```diff
@@ -193,77 +231,24 @@ private[adam] class FastqRecordConverter extends Serializable with Logging {
...
-      .setReadNegativeStrand(false)
-      .setMateNegativeStrand(null)
-      .setPrimaryAlignment(null)
-      .setSecondaryAlignment(null)
-      .setSupplementaryAlignment(null)
-
-    recordGroupOpt.foreach(builder.setRecordGroupName)
-
-    builder.build()
+    this.makeAlignmentRecord(
```
zyxue: I have removed all `this.`, wondering when  will `this.` ever be necessary?

adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
zyxue: I tried to implement parsing for wrapped lines. Then I found that it would require exact match of sequence length and quality length. Otherwise, it's ambiguous to tell when the quality lines stop. This makes padding with `B` when length(qual line) < length(seq line), is that right?

zyxue:
@heuermh, Can you test it again, please? If there is no further request, I think it's ready to be merged.

heuermh:
Jenkins, test this please.

adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
heuermh: `this` is only necessary if there is a collision between say a field and a local variable with the same name

heuermh:
> Where can I find more information on `ValidationStringency.STRICT`, what should the behavior be when it's STRICT or not, please?

We're borrowing the enum and concept from [htsjdk](https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/samtools/ValidationStringency.java).  Briefly, on errors strict should throw exceptions, lenient should log warnings, and silent should not complain.

zyxue:
Jenkins, test this please.

zyxue:
(took a shot, trying to see if it works by me...)

heuermh:
Jenkins, retest this please

fnothaft:
LGTM! Ping @heuermh for a review pass.
I'm thinking that we should merge this manually, since the commit history is pretty long. @heuermh let me know if/when it looks good to you, and I will merge it.

fnothaft:
@heuermh is this still pending changes from your side?

heuermh:
Will complete review on Monday

adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
```diff
@@ -41,6 +41,114 @@ import scala.collection.JavaConversions._
 private[adam] class FastqRecordConverter extends Serializable with Logging {
 
   /**
+   * Parse 4 lines at a time
```
heuermh: This doc comment doesn't match the method.

Perhaps something like `Return true if the read name suffix and flags match.`

adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
```diff
@@ -41,6 +41,114 @@ import scala.collection.JavaConversions._
...
+    val firstReadSuffix = """[/ +_]1$""".r
+    val secondReadSuffix = """[/ +_]2$""".r
+
+    val isSecondOfPair = !isFirstOfPair
+
+    val match1 = firstReadSuffix.findAllIn(readName)
+    val match2 = secondReadSuffix.findAllIn(readName)
+
+    if (match1.nonEmpty && isSecondOfPair)
+      throw new IllegalArgumentException(
```
heuermh: These exceptions are thrown without considering `ValidationStringency`.  If the stringency is lenient or silent, is it possible to continue processing?

adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
```diff
@@ -41,6 +41,114 @@ import scala.collection.JavaConversions._
...
+    require(!(setFirstOfPair && setSecondOfPair))
+
+    val lines = input.split('\n')
+    require(lines.length == 4,
+      s"Input must have 4 lines (${lines.length.toString} found):\n${input}")
+
+    val readName = lines(0).drop(1)
+    if (setFirstOfPair || setSecondOfPair) readNameSuffixAndIndexOfPairMustMatch(readName, setFirstOfPair)
+
+    val suffix = """([/ +_]1$)|([/ +_]2$)""".r
```
heuermh: I wonder if this regex and those above on line 50 and 51 might be combined as static private fields so that the two methods don't get out of sync

adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
```diff
@@ -193,77 +249,21 @@ private[adam] class FastqRecordConverter extends Serializable with Logging {
...
-      } else {
-        readName
-      }
-    }
-
-    // get fields for first read in pair
-    val readName = trimTrailingReadNumber(lines(0).drop(1))
-    val readSequence = lines(1)
+    if (setFirstOfPair && setSecondOfPair)
+      throw new IllegalArgumentException("setFirstOfPair and setSecondOfPair cannot be true at the same time")
```
heuermh: This exception is also thrown without considering ValidationStringency. If the stringency is lenient or silent, is it possible to continue processing?

adam-core/src/main/scala/org/bdgenomics/adam/converters/FastqRecordConverter.scala
```diff
@@ -41,6 +41,114 @@ import scala.collection.JavaConversions._
...
+                                              readPaired: Boolean = true,
+                                              recordGroupOpt: Option[String] = None): AlignmentRecord = {
+    val builder = AlignmentRecord.newBuilder
+      .setReadName(readName)
+      .setSequence(sequence)
+      .setQual(qual)
+      .setReadPaired(readPaired)
+      .setReadInFragment(readInFragment)
+
+    if (recordGroupOpt != None)
```
heuermh: With `opt.foreach` you don't also need to check against `None`

fnothaft:
Moved over to #1208.

PR #1187, opened 2016-09-28
[ADAM-1186] Mask null from fs.globStatus.

### LINKED ISSUE #1186: Hadoop globStatus returns null if no glob matches
As a result, [`paths.isEmpty` throws an NPE](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala#L307).

fnothaft:
Resolves #1186. Check for a null pointer in ADAMContext.getFiles. Make ADAMContext.getFsAndFiles package private so that we can unit test it. Added unit tests to check for proper exception with empty directories and bad globs.


heuermh:
Thank you, @fnothaft!

PR #1198, opened 2016-10-06
[ADAM-1192] Correctly handle other whitespace in FASTA description.

### LINKED ISSUE #1192: MDTagging fails on GRCh38
I was running the `transform -add_md_tags` code on GRCh38, and was getting contig name mismatches that caused failures. It looked like a header parsing problem. TBD. I need to look back in the logs for the stack trace.

fnothaft:
Resolves #1192. More broadly handles whitespace in the FASTA description, and also disregards FASTA metadata which is inlined.

Testing this on the cluster as we speak.


adam-cli/src/test/scala/org/bdgenomics/adam/cli/Fasta2ADAMSuite.scala
```diff
@@ -30,8 +30,8 @@ class Fasta2ADAMSuite extends ADAMFunSuite {
     val contigFragments = sc.loadParquetContigFragments(convertPath)
     assert(contigFragments.rdd.count() === 26)
     val first = contigFragments.rdd.first()
-    assert(first.getContig.getContigName === "gi|224384749|gb|CM000682.1|")
-    assert(first.getDescription === "Homo sapiens chromosome 20, GRCh37 primary reference assembly")
+    assert(first.getContig.getContigName === null)
```
heuermh: Not sure I understand, the string before the first space is typically the sequence name in FASTA format.  You might want to compare to the various open-bio projects.

fnothaft: Well, `gi` is the Entrez ID, and `gb` is the GenBank ID. I assure you that they are not the sequence name.

fnothaft:
Whoops! Forgot to commit the test resource. Just pushed.

fnothaft:
With this, recomputing the MD tags worked on the cluster with GRCh38.

heuermh:
Thank you, @fnothaft!

Issue #1202, opened 2016-10-08
gVCF - can't load multi-allelic sites


jpdna:
I'm trying to input the file:
http://bioinformaticstools.mayo.edu/research/wp-content/plugins/download.php?url=https://s3-us-west-2.amazonaws.com/mayo-bic-tools/variant_miner/gvcfs/NA12878.chr22.g.vcf.gz

I believe I solved the SB tag problem of #1199 

However, now on the VCF row below I get the following error:

```
chr22   18027817        .       CTTTT   C,CT,CTT,CTTT,<NON_REF> 1364.73 .       DP=78;MLEAC=0,0,1,1,0;MLEAF=0.00,0.00,0.500,0.500,0.00;MQ=70.00;MQ0=0   GT:AD:DP:GQ:PL:SB       3/4:0,4,6,28,17,0:55:99:1402,1244,2007,972,1382,1297,371,355,301,301,792,617,486,0,653,1038,1016,883,315,597,935:0,0,14,3
```

Note, clearly the row above is not a reference block, it is called multi-allelic site within the gVCF file 

```
 x2.rdd.count
2016-10-08 13:04:26 ERROR Executor:95 - Exception in task 0.0 in stage 0.0 (TID 0)
java.lang.IllegalArgumentException: Multi-allelic site with non-ref symbolic allele[VC Unknown @ chr22:18027817-18027821 Q1364.73 of type=MIXED alleles=[CTTTT*, <NON_REF>, C, CT, CTT, CTTT] attr={DP=78, MLEAC=[0, 0, 1, 1, 0], MLEAF=[0.00, 0.00, 0.500, 0.500, 0.00], MQ=70.00, MQ0=0} GT=GT:AD:DP:GQ:PL:SB 3/4:0,4,6,28,17,0:55:99:1402,1244,2007,972,1382,1297,371,355,301,301,792,617,486,0,653,1038,1016,883,315,597,935:0,0,14,3
    at org.bdgenomics.adam.converters.VariantContextConverter.convert(VariantContextConverter.scala:214)
    at org.bdgenomics.adam.rdd.ADAMContext$$anonfun$loadVcf$1.apply(ADAMContext.scala:823)
    at org.bdgenomics.adam.rdd.ADAMContext$$anonfun$loadVcf$1.apply(ADAMContext.scala:823)
    at scala.collection.Iterator$$anon$13.hasNext(Iterator.scala:371)
[...]

PR #1114, opened 2016-08-12
Add pipes API

fnothaft:
WIP towards #1112, resolves #1113. Will post TODOs later.

PR #1203, opened 2016-10-09
fix SB tag parsing

jpdna:
The SB tag defined as "Per-sample component statistics which comprise the Fisher's Exact Test to detect strand bias":
```
##FORMAT=<ID=SB,Number=4,Type=Integer,Description="Per-sample component statistics which comprise the Fisher's Exact Test to detect strand bias.">
```
is not parsed properly in existing code or covered by tests.
This came to light in trying to parse the gVCF file:
http://bioinformaticstools.mayo.edu/research/wp-content/plugins/download.php?url=https://s3-us-west-2.amazonaws.com/mayo-bic-tools/variant_miner/gvcfs/NA12878.chr22.g.vcf.gz

This PR correctly parses this SB tag and adds a variant with this tag to `small.vcf` and updates tests.


jpdna:
Note - I realize its less than ideal that so many count assertions had to be changed in the tests in this PR due to changing of small.vcf   For the multi-allele issue in #1202 I'll look at making a new test vcf file in another directory so as to not have side effects on the counts, and we may also want to find an alternative in this PR's test component.

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantAnnotationConverter.scala
```diff
@@ -17,6 +17,9 @@
  */
 package org.bdgenomics.adam.converters
 
+import scala.collection.JavaConversions._
```
fnothaft: Nit: these should go at the end. Actually, we should lex sort all the imports here, since it should be `htsjdk`, `org.apache`, then `scala`.

fnothaft: Also, no whitespace.

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantAnnotationConverter.scala
```diff
@@ -124,6 +127,17 @@ private[converters] object VariantAnnotationConverter extends Serializable {
   }
 
   /**
+   * Converts a java String of comma delimited integers to a
+   * java.util.List of Integer
+   *
+   * @param attr Attribute to convert.
+   * @return Attribute as a java.util.List[Integer]
+   */
+  private def attrAsIntList(attr: Object): Object = attr match {
```
fnothaft: Can you add a test that covers this function?

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantAnnotationConverter.scala
```diff
@@ -187,7 +201,7 @@ private[converters] object VariantAnnotationConverter extends Serializable {
     AttrKey("phaseQuality", attrAsInt _, new VCFFormatHeaderLine(VCFConstants.PHASE_QUALITY_KEY, 1, VCFHeaderLineType.Float, "Read-backed phasing quality")),
     AttrKey("phaseSetId", attrAsInt _, new VCFFormatHeaderLine(VCFConstants.PHASE_SET_KEY, 1, VCFHeaderLineType.Integer, "Phase set")),
     AttrKey("minReadDepth", attrAsInt _, new VCFFormatHeaderLine("MIN_DP", 1, VCFHeaderLineType.Integer, "Minimum DP observed within the GVCF block")),
-    AttrKey("strandBiasComponents", attrAsInt _, new VCFFormatHeaderLine("SB", 4, VCFHeaderLineType.Integer, "Per-sample component statistics which comprise the Fisher's Exact Test to detect strand bias."))
+    AttrKey("strandBiasComponents", attrAsIntList _, new VCFFormatHeaderLine("SB", 4, VCFHeaderLineType.Integer, "Per-sample component statistics which comprise the Fisher's Exact Test to detect strand bias."))
```
heuermh: Do you have a doc link for the `SB` tag being used for genotypes (i.e. as a FORMAT key)?  What are the 4 different values?  Are there always 4 values?

jpdna: GATK docs define it as always with `Number=4`
https://software.broadinstitute.org/gatk/gatkdocs/org_broadinstitute_gatk_tools_walkers_annotator_StrandBiasBySample.php

fnothaft: SB components are a 2x2 matrix: forward/rev strand on one axis, ref/alt on the other axis

fnothaft:
Ping @jpdna for adding tests, small import cleanup, then I am good to merge this.

jpdna:
This test failure is weird - and only on Spark 1.5.2, and I wouldn't think that the very last commit would have broken anything related to these failures.  Thoughts?

heuermh:
Note that the build matrix chart doesn't accurately show what happened here, to keep build times down, Jenkins only builds 2.6.0,2.10,1.5.2,centos and 2.6.0,2.11,1.5.2,centos.  If those builds succeed, then it builds out the rest of the matrix.  So in this case, the two red dots represent new failed builds from this pull request, and the green dots are successful builds from something earlier.

fnothaft:
Test failures repro for me locally. But, it's a simple fix. Will shoot your way in 1 min @jpdna .

fnothaft:
@jpdna I've opened PR #1209 with https://github.com/bigdatagenomics/adam/pull/1209/commits/3536385f5133f5d0d055493bdcf5b21de8cd5064, which fixes the issue. We added two tests in #1114 that used the VCF that you added a record to, so I just updated the tests to account for the new record. I also rebased your PR on ToT. Let me know if #1209 looks good to you, and I will squash it down and merge it when tests pass.

jpdna:
My mistake here is that I should have re-based before the PR myself first, and then I would have seen the failure locally that Jenkins saw -  correct?

Yes - #1209 looks good to me.  Thanks!

fnothaft:
> My mistake here is that I should have re-based before the PR myself first, and then I would have seen the failure locally that Jenkins saw - correct?

Yeah, exactly. No worries! I did the same thing yesterday with #1114, actually...

Issue #1212, opened 2016-10-14
Release 0.20.0 artifacts failed Sonatype Nexus validation


heuermh:
The `adam-assembly` module does not produce javadoc or source jars, and as such fails the Sonatype Nexus validation step:

Event: Failed: Javadoc Validation
Friday, October 14, 2016 18:06:02 CDT (GMT-0500)
typeId	javadoc-staging
failureMessage	Missing: no javadoc jar found in folder '/org/bdgenomics/adam/adam-assembly_2.10/0.20.0'
failureMessage	Missing: no javadoc jar found in folder '/org/bdgenomics/adam/adam-assembly-spark2_2.10/0.20.0'
failureMessage	Missing: no javadoc jar found in folder '/org/bdgenomics/adam/adam-assembly-spark2_2.11/0.20.0'
failureMessage	Missing: no javadoc jar found in folder '/org/bdgenomics/adam/adam-assembly_2.11/0.20.0'

Event: Failed: Sources Validation
Friday, October 14, 2016 18:06:05 CDT (GMT-0500)
typeId	sources-staging
failureMessage	Missing: no sources jar found in folder '/org/bdgenomics/adam/adam-assembly_2.10/0.20.0'
failureMessage	Missing: no sources jar found in folder '/org/bdgenomics/adam/adam-assembly-spark2_2.10/0.20.0'
failureMessage	Missing: no sources jar found in folder '/org/bdgenomics/adam/adam-assembly-spark2_2.11/0.20.0'
failureMessage	Missing: no sources jar found in folder '/org/bdgenomics/adam/adam-assembly_2.11/0.20.0'

heuermh:
Sent an email to ossrh-users@sonatype to see if it might be possible to deploy signed empty sources and javadoc jars to the currently open staging repository.  Otherwise I think we might be hosed and may need to move to an 0.20.1 release.

fnothaft:
OOC, how have we been getting around this with [adam-assembly](http://search.maven.org/#search%7Cga%7C1%7Ca%3A%22adam-distribution%22)? Is the requirement only for artifacts that push JARs?

heuermh:
The adam-distribution module has `<packaging>pom</packaging>`, while adam-assembly has `<packaging>jar</packaging>`.

From http://central.sonatype.org/pages/requirements.html#supply-javadoc-and-sources:

"Projects with packaging other than pom have to supply JAR files that contain Javadoc and sources."

[...]

PR #1215, opened 2016-10-17
[ADAM-1212] Add empty assembly object, allows Maven build to create sources and javadoc artifacts

heuermh:
Progress towards #1212


fnothaft:
LGTM! Do you want to squash these two commits down, or would you prefer each to stand alone? I can see the merits to both.

heuermh:
If you're going to let the second one slide, then sure, as separate commits is probably better.  :)

fnothaft:
Merged! Thank you @heuermh! 

> If you're going to let the second one slide, then sure, as separate commits is probably better. :)

"I want to document our strife." ;)

Issue #1224, opened 2016-10-24
broadcast join using interval tree 


jpdna:
Broadcast interval join currently is being done in a way that does a Cartesian product and then filter:

https://github.com/bigdatagenomics/adam/blob/bd3c62acec6902c002867c750f41c4d9f8ad4db0/adam-core/src/main/scala/org/bdgenomics/adam/rdd/BroadcastRegionJoin.scala#L55

that feel inefficient.

Most genomics tools for finding interval overlap use some kind of interval or R-tree as discussed:
https://www.biostars.org/p/2244/

Proposed implementation, when one side of the join is small enough ( < 100 MB) to be broadcast.

1) read the "smaller"  side into an intervaltree data structure, either the one for Mango or, https://samtools.github.io/htsjdk/javadoc/htsjdk/htsjdk/samtools/util/IntervalTree.html
Possible as a map of intervaltrees, one for each chromosome/contig

2) Broadcast that map of interval trees

3) do a map or mapPartitions, which loops through each member of the RDD, and does the "map side join" using overlap operation of the interval tree.

At the least I think we should implement this approach as a comparison with other join modes we develop.

Thoughts?  This doesn't already exist in ADAM code does it?

fnothaft:
I concur! (see #1110, which I've closed in favor of this issue). I spoke with @tdanford a while back and there were historical reasons for why the BroadcastJoin was called BroadcastJoin. The documents in the BroadcastJoin code are actually kinda wrong as well; we don't actually do a cartesian and filter. I would +1 your proposed implementation.

fnothaft:
Also see https://bmcbioinformatics.biomedcentral.com/articles/10.1186/s12859-016-0904-1.

Also, @akmorrow13's interval tree code in https://github.com/bigdatagenomics/utils/tree/master/utils-intervalrdd/src/main/scala/org/bdgenomics/utils/intervaltree should allow us to do an efficient `broadcastJoinAndGroupByRight` implementation, akin to our `shuffleJoinAndGroupByLeft` code.

jpdna:
@fnothaft you mentioned on call today some other code relevant to this that I should look at.  When you get the chance can you paste it on this thread?

fnothaft:
Yes! You'll want to take a look at the following commits:

* https://github.com/bigdatagenomics/avocado/commit/ceaee3bb709428a5e20c0202b49a1d26e238f283
[...]

PR #1234, opened 2016-11-01
[ADAM-1224] Replace BroadcastRegionJoin with tree based algo.

fnothaft:
WIP towards #1224. To be done:

* [x] Add more tests for outer joins
* [x] Fix serialization stuff (note disconnected `ForestSerializer`)

The custom serializer is necessary because Kryo does not serialize large arrays properly, which is... concerning.


jpdna:
Is this ready for someone else to test/review or do you have further work to do @fnothaft ?  Ping when you do want review.

fnothaft:
@jpdna I need to write some more unit tests, but otherwise the core of the code is ready for review!

adam-core/src/main/scala/org/bdgenomics/adam/rdd/TreeRegionJoin.scala
```diff
@@ -0,0 +1,328 @@
...
+  ReferenceRegion,
+  ReferenceRegionSerializer
+}
+import scala.annotation.tailrec
+import scala.reflect.ClassTag
+
+/**
+ * Companion object for building a forest from an RDD.
+ */
+private[adam] object Forest extends Serializable {
```
akmorrow13: If we are using the implementation for IntervalRDD, we would most likely be moving Forest and ForestSerializer to utils. Is there anything else that should be moved to utils?

fnothaft: I think that is all.

jpdna:
I notice that in the `overlaps` functions of `referenceRegion` called in `binarySearch`  that there is a requirement that orientation be the same in order to return true: https://github.com/bigdatagenomics/adam/blob/748b9521789097a17691fe3c811f42e9fba1dcaf/adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala#L344

At first glance, I'm wondering if this is a problem, for example if joining a deletion or other feature expressed on the positive strand against a gene or transcript where `Orientation=Strand.REVERSE`

jpdna:
Also, I have small itch of concern ( which I haven't tested yet) that there could a scenario where 
`expand`
https://github.com/bigdatagenomics/adam/pull/1234/files#diff-7dca5f64826de300ea13b67c5ce39687R94
stops searching further into the sorted referenceRegion broadcast array because the next referenceRegion it finds doesn't overlap, but further upstream or downstream in the array there is larger referenceRegion feature that does overlap.

Like
```
                                                       
query                                         ---

brod_array                        --       ----------   --
                            --------------------------------------------------
```

this may well be just fine, just wanted to note it.
They broadcast array is sorted by start position of ReferenceRegions correct?

jpdna:
ping wrt to my two comments above

fnothaft:
I've upgraded this on top of https://github.com/bigdatagenomics/utils/pull/94 in the last commit (and rebased) but it fails one unit test due to some serializer funniness. I'm going to make a cleanup pass on utils/#94 tomorrow to fix this.

fnothaft:
Jenkins, retest this please.

fnothaft:
Jenkins, test this please.

fnothaft:
Jenkins, retest this please.

fnothaft:
Jenkins, retest this please.

fnothaft:
This is squashed down and good to go!

scripts/move_to_scala_2.11.sh
```diff
@@ -2,7 +2,7 @@
 
 set +x
 
-find . -name "pom.xml" -exec sed -e "s/2.10.6/2.11.8/g" -e "s/2.10/2.11/g" -i.2.11.bak '{}' \;
+find . -name "pom.xml" -exec sed -e "s/2.10.6/2.11.8/g" -e "/bdg-utils.version/!s/2.10/2.11/g" -i.2.11.bak '{}' \;
```
heuermh: What is this for?  Do we need something similar in `move_to_scala_2.10.sh`?

fnothaft: The utils version bumps to 0.2.10 in this PR. If you don't filter out that line when running the sed replacements, then it changes the utils version to 0.2.11. This doesn't happen going 2.11->2.10, since the replacement is different.

fnothaft:
Rebased and resolved conflicts.

scripts/move_to_scala_2.11.sh
heuermh: I see, good catch  :)

heuermh:
Thank you, @fnothaft

Issue #493, opened 2014-11-19
Update usage docs running for EC2 and CDH 


arahuja:
- Update to use adam-submit
- Update CDH docs to have YARN and standalone information

jpdna:
Regarding EC2, the existing (out of date) documentation is here: 
https://github.com/bigdatagenomics/adam/blob/master/docs/source/40_running_on_EC2.md

To confirm, do we want the updated ADAM EC2 quickstart to deploy Spark via `spark-ec2` scripts as described at?: 
http://spark.apache.org/docs/1.6.1/ec2-scripts.html

jpdna:
So Spark's spark-ec2 script as of Spark 1.6.1 doesn't work for us, unless we modify it, due to it using java 7 and we need java 8.

I still feel like cgcloud/toil is a bit heavy weight for some users, and perhaps we should modify spark-ec2 script, or make an equivalent, that can be used for simple ADAM deployment on EC2.

But - if you think we should only document a cgcloud/toil path for ADAM on EC2 that is fine too.
What do  you think @fnothaft  and @heuermh

fnothaft:
> So Spark's spark-ec2 script as of Spark 1.6.1 doesn't work for us, unless we modify it, due to it using java 7 and we need java 8.

Officially, the `spark-ec2` script is beta/unsupported, so we shouldn't use it anyways.

> I still feel like cgcloud/toil is a bit heavy weight for some users, and perhaps we should modify spark-ec2 script, or make an equivalent, that can be used for simple ADAM deployment on EC2.

I would do `cgcloud-spark`, which is fairly battle tested, well supported by Hannes + co, etc. I would definitely not modify the `spark-ec2` script or roll our own. There's a non-trivial amount of work needed to get all of this working, and you run into a shocking number of bugs and weird edge cases.

If time allowed, I would also document ADAM on Elastic Spark, but I would not plan on that as a primary documented route, since the cost model for Elastic Spark is different. If we are interested in that, I can connect you to some folks at AWS.

jpdna:
It looks like cgcloud-spark is on Spark 1.5.2
and java7.

We should have java8 and Spark 1.6.1 to match 
the next release and current master

Should we talk with Hannes about updating `spark_box` ?

Also - what is the interaction here with the joda time stuff?

fnothaft:
I'm OK with staying on Spark 1.5.2 or moving to 1.6.1. It's a one line change. I'll open a PR to move to Java 8.

[...]

PR #1246, opened 2016-11-07
HBase backend for Genotypes

jpdna:
Requesting here a first pass review for the integration of HBase input and output functions into Master.

The code in this PR excludes the "custom encoding" options for clarity and thus sticks with saving the full Avro schema of a Genotype as the values stored in HBase.

Todo:
- [ ] Add Unit tests using an embedded mini-hbase cluster
- [ ] Add more parameter checking and informative error messages, such as if database table doesn't exist
- [ ] Integrate and rebase against upcoming changes to Genotype schemas when available

Questions:
- This PR has a new hbase/HBaseFunctions.scala module parallel to RDD in the package structure in adam-core, is this OK, or other suggestions?

- For the first round of HBase code integration I suggest we target CDH HBase only as that is where we are doing testing and deployment.  In future we can add a profile for POM for  Apache Hbase 2.0, though this  would seem to require use of HBase 2.0 snapshot as Spark support doesn't exist in the last official non-cloudera apache release.

### DIFF
--- a/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Vcf2HBase.scala
+++ b/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Vcf2HBase.scala
@@ -1,0 +1,82 @@
+/**
+ * Licensed to Big Data Genomics (BDG) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The BDG licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.bdgenomics.adam.cli
+
+import org.apache.spark.SparkContext
+
+import org.bdgenomics.adam.hbase.HBaseFunctions
+import org.bdgenomics.adam.instrumentation.Timers._
+import org.bdgenomics.adam.models.{ RecordGroupDictionary, SequenceDictionary, SnpTable }
+import org.bdgenomics.adam.projections.{ AlignmentRecordField, Filter }
+import org.bdgenomics.adam.rdd.ADAMContext._
+import org.bdgenomics.adam.rdd.ADAMSaveAnyArgs
+import org.bdgenomics.adam.rdd.read.{ AlignedReadRDD, AlignmentRecordRDD, MDTagging }
+import org.bdgenomics.adam.rich.RichVariant
+import org.bdgenomics.formats.avro.AlignmentRecord
+import org.bdgenomics.utils.cli._
+import org.bdgenomics.utils.misc.Logging
+import org.kohsuke.args4j.{ Argument, Option => Args4jOption }
+
+object Vcf2HBase extends BDGCommandCompanion {
+  val commandName = "Vcf2HBase"
+  val commandDescription = "Write a VCF file to HBase"
+
+  def apply(cmdLine: Array[String]) = {
+    new Vcf2HBase(Args4j[Vcf2HBaseArgs](cmdLine))
+  }
+}
+
+class Vcf2HBaseArgs extends Args4jBase {
+
+  @Argument(required = true, metaVar = "VCF", usage = "The VCF file to convert", index = 0)
+  var vcfPath: String = _
+
+  @Args4jOption(required = true, name = "-hbase_table", usage = "HBase table name in which to load VCF file")
+  var hbaseTable: String = null
+
+  @Args4jOption(required = true, name = "-seq_dict_id", usage = "User defined name to apply to the sequence dictionary create from this VCF, or name of existing sequence dictionary to be used")
+  var seqDictId: String = null
+
+  @Args4jOption(required = false, name = "-use_existing_seqdict", usage = "Use an existing sequence dictionary, don't write a new one")
+  var useExistingSeqDict: Boolean = false
+
+  @Args4jOption(required = false, name = "-repartition_num", usage = "Reparition into N partitions prior to writing to HBase")
+  var repartitionNum: Int = 0
+
+  @Args4jOption(required = true, name = "-staging_folder", usage = "location for temporary files during bulk load")
+  var stagingFolder: String = null
+
+}
+
+class Vcf2HBase(protected val args: Vcf2HBaseArgs) extends BDGSparkCommand[Vcf2HBaseArgs] with Logging {
+  val companion = Transform
+
+  def run(sc: SparkContext) {
+
+    val vcRdd = sc.loadVcf(args.vcfPath)
+
+    HBaseFunctions.saveVariantContextRDDToHBaseBulk(sc,
+      vcRdd,
+      args.hbaseTable,
+      sequenceDictionaryId = args.seqDictId,
+      numPartitions = args.repartitionNum,
+      stagingFolder = args.stagingFolder)
+
+  }
+
+}
--- a/adam-core/pom.xml
+++ b/adam-core/pom.xml
@@ -137,6 +137,10 @@
       <artifactId>spark-core_${scala.version.prefix}</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.apache.spark</groupId>
+      <artifactId>spark-streaming_2.10</artifactId>
+    </dependency>
+    <dependency>
       <groupId>it.unimi.dsi</groupId>
       <artifactId>fastutil</artifactId>
     </dependency>
@@ -185,5 +189,37 @@
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
     </dependency>
+
+    <dependency>
+      <groupId>org.apache.hbase</groupId>
+      <artifactId>hbase-hadoop2-compat</artifactId>
+
+    </dependency>
+
+    <dependency>
+      <groupId>org.apache.hbase</groupId>
+      <artifactId>hbase-server</artifactId>
+
+    </dependency>
+
+    <dependency>
+      <groupId>org.apache.hbase</groupId>
+      <artifactId>hbase-it</artifactId>
+
+    </dependency>
+
+    <dependency>
+      <groupId>org.apache.hbase</groupId>
+      <artifactId>hbase-client</artifactId>
+
+    </dependency>
+
+    <dependency>
+      <groupId>org.apache.hbase</groupId>
+      <artifactId>hbase-spark</artifactId>
+
+    </dependency>
+
+
   </dependencies>
 </project>
--- a/adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
@@ -1,0 +1,525 @@
+/**
+ * Licensed to Big Data Genomics (BDG) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The BDG licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.bdgenomics.adam.hbase
+
+import java.util
+import org.apache.hadoop.hbase.io.ImmutableBytesWritable
+import org.apache.hadoop.hbase.io.encoding.DataBlockEncoding
+import org.apache.spark.SparkContext
+import org.apache.spark.rdd.RDD
+import org.apache.hadoop.hbase.client._
+import org.apache.hadoop.hbase._
+import org.apache.hadoop.hbase.spark._
+import org.bdgenomics.adam.rdd.variation.{ GenotypeRDD, VariantContextRDD }
+import org.bdgenomics.adam.rich.RichVariant
+import org.bdgenomics.formats.avro._
+import org.apache.hadoop.hbase.spark.HBaseRDDFunctions._
+import org.apache.hadoop.hbase.util.Bytes
+import org.apache.hadoop.hbase.{ HBaseConfiguration, TableName }
+import org.apache.avro.specific.SpecificDatumWriter
+import org.apache.avro.io._
+import java.io.ByteArrayOutputStream
+import org.apache.hadoop.hbase.io.compress.Compression.Algorithm
+import org.apache.avro.specific.SpecificDatumReader
+import org.apache.hadoop.hbase.mapreduce.LoadIncrementalHFiles
+import org.apache.hadoop.fs.{ FSDataInputStream, FileSystem, Path }
+import org.bdgenomics.adam.models.{ ReferenceRegion, ReferencePosition, SequenceDictionary, VariantContext }
+import scala.io.Source
+
+import scala.collection.mutable.ListBuffer
+import sys.process._
+
+import scala.collection.JavaConverters._
+
+/**
+ * Created by jp on 7/23/16.
+ */
+
+object HBaseFunctions {
+
+  def saveSequenceDictionaryToHBase(hbaseTableName: String,
+                                    sequences: SequenceDictionary,
+                                    sequenceDictionaryId: String): Unit = {
+
+    val conf = HBaseConfiguration.create()
+    val connection = ConnectionFactory.createConnection(conf)
+    val table: Table = connection.getTable(TableName.valueOf(hbaseTableName))
+    val contigs = sequences.toAvro
+
+    val baos: java.io.ByteArrayOutputStream = new ByteArrayOutputStream()
+    val encoder = EncoderFactory.get().binaryEncoder(baos, null)
+
+    contigs.foreach((x) => {
+      val contigDatumWriter: DatumWriter[Contig] =
+        new SpecificDatumWriter[Contig](scala.reflect.classTag[Contig].runtimeClass.asInstanceOf[Class[Contig]])
+      contigDatumWriter.write(x, encoder)
+    })
+
+    encoder.flush()
+    baos.flush()
+
+    val put: Put = new Put(Bytes.toBytes(sequenceDictionaryId))
+    put.addColumn(Bytes.toBytes("meta"), Bytes.toBytes("contigdata"), baos.toByteArray)
+    table.put(put)
+  }
+
+  def loadSequenceDictionaryFromHBase(hbaseTableName: String,
+                                      sequenceDictionaryId: String): SequenceDictionary = {
+
+    val conf = HBaseConfiguration.create()
+    val connection = ConnectionFactory.createConnection(conf)
+    val table = connection.getTable(TableName.valueOf(hbaseTableName))
+
+    val myGet = new Get(Bytes.toBytes(sequenceDictionaryId))
+    val result = table.get(myGet)
+    val dataBytes = result.getValue(Bytes.toBytes("meta"), Bytes.toBytes("contigdata"))
+
+    val contigDatumReader: DatumReader[Contig] =
+      new SpecificDatumReader[Contig](scala.reflect.classTag[Contig]
+        .runtimeClass
+        .asInstanceOf[Class[Contig]])
+
+    val decoder = DecoderFactory.get().binaryDecoder(dataBytes, null)
+    var resultList = new ListBuffer[Contig]
+    while (!decoder.isEnd) {
+      resultList += contigDatumReader.read(null, decoder)
+    }
+
+    SequenceDictionary.fromAvro(resultList)
+
+  }
+
+  def saveSampleMetadataToHBase(hbaseTableName: String,
+                                samples: Seq[Sample]): Unit = {
+
+    val conf = HBaseConfiguration.create()
+    val connection = ConnectionFactory.createConnection(conf)
+    val table = connection.getTable(TableName.valueOf(hbaseTableName))
+
+    samples.foreach((x) => {
+      val baos: java.io.ByteArrayOutputStream = new ByteArrayOutputStream()
+      val encoder = EncoderFactory.get().binaryEncoder(baos, null)
+
+      val sampleDatumWriter: DatumWriter[Sample] =
+        new SpecificDatumWriter[Sample](scala.reflect.classTag[Sample].runtimeClass.asInstanceOf[Class[Sample]])
+      sampleDatumWriter.write(x, encoder)
+      encoder.flush()
+      baos.flush()
+
+      val curr_sampleid = x.getSampleId
+
+      val put = new Put(Bytes.toBytes(curr_sampleid))
+      put.addColumn(Bytes.toBytes("meta"), Bytes.toBytes("sampledata"), baos.toByteArray)
+      table.put(put)
+
+    })
+  }
+
+  def loadSampleMetadataFromHBase(hbaseTableName: String,
+                                  sampleIds: List[String]): Seq[Sample] = {
+
+    val conf = HBaseConfiguration.create()
+    val connection = ConnectionFactory.createConnection(conf)
+    val table = connection.getTable(TableName.valueOf(hbaseTableName))
+
+    var resultList = new ListBuffer[Sample]
+
+    sampleIds.foreach((sampleId) => {
+      val myGet = new Get(Bytes.toBytes(sampleId))
+      val result = table.get(myGet)
+      val dataBytes = result.getValue(Bytes.toBytes("meta"), Bytes.toBytes("sampledata"))
+
+      val sampleDatumReader: DatumReader[Sample] =
+        new SpecificDatumReader[Sample](scala.reflect.classTag[Sample]
+          .runtimeClass
+          .asInstanceOf[Class[Sample]])
+
+      val decoder = DecoderFactory.get().binaryDecoder(dataBytes, null)
+
+      while (!decoder.isEnd) {
+        resultList += sampleDatumReader.read(null, decoder)
+      }
+
+    })
+
+    resultList
+
+  }
+
+  def loadRDDofGenotypeFromHBase(sc: SparkContext,
+                                 hbaseTableName: String,
+                                 sampleIds: List[String],
+                                 queryRegion: ReferenceRegion = null,
+                                 numPartitions: Int = 0): RDD[Genotype] = {
+
+    val scan = new Scan()
+    scan.setCaching(100)
+    scan.setMaxVersions(1)
+
+    if (queryRegion != null) {
+      val start = queryRegion.referenceName + "_" + String.format("%10s", queryRegion.start.toString).replace(' ', '0')
+      val stop = queryRegion.referenceName + "_" + String.format("%10s", queryRegion.end.toString).replace(' ', '0')
+      scan.setStartRow(Bytes.toBytes(start))
+      scan.setStopRow(Bytes.toBytes(stop))
+    }
+
+    val conf = HBaseConfiguration.create()
+    val hbaseContext = new HBaseContext(sc, conf)
+
+    sampleIds.foreach(sampleId => {
+      scan.addColumn(Bytes.toBytes("g"), Bytes.toBytes(sampleId))
+    })
+
+    val resultHBaseRDD = hbaseContext.hbaseRDD(TableName.valueOf(hbaseTableName), scan)
+
+    val resultHBaseRDDrepar = if (numPartitions > 0) resultHBaseRDD.repartition(numPartitions)
+    else resultHBaseRDD
+
+    val result: RDD[Genotype] = resultHBaseRDDrepar.mapPartitions((iterator) => {
+
+      val genotypeDatumReader: DatumReader[Genotype] = new SpecificDatumReader[Genotype](scala.reflect.classTag[Genotype]
+        .runtimeClass
+        .asInstanceOf[Class[Genotype]])
+
+      iterator.flatMap((curr) => {
+        var resultList = new ListBuffer[Genotype]
+        sampleIds.foreach((sample) => {
+          val myVal: util.List[Cell] = curr._2.getColumnCells(Bytes.toBytes("g"), Bytes.toBytes(sample))
+          val decoder = DecoderFactory.get().binaryDecoder(CellUtil.cloneValue(myVal.get(0)), null)
+          while (!decoder.isEnd) {
+            resultList += genotypeDatumReader.read(null, decoder)
+          }
+        })
+        resultList
+      })
+
+    })
+    result
+  }
+
+  def loadRDDofVariantContextFromHBase(sc: SparkContext,
+                                       hbaseTableName: String,
+                                       inputsampleIds: List[String] = null,
+                                       sampleListFile: String = null,
+                                       queryRegion: ReferenceRegion = null,
+                                       numPartitions: Int = 0): RDD[VariantContext] = {
+
+    val scan = new Scan()
+    scan.setCaching(100)
+    scan.setMaxVersions(1)
+
+    if (queryRegion != null) {
+      val start = queryRegion.referenceName + "_" + String.format("%10s", queryRegion.start.toString).replace(' ', '0')
+      val stop = queryRegion.referenceName + "_" + String.format("%10s", queryRegion.end.toString).replace(' ', '0')
+      scan.setStartRow(Bytes.toBytes(start))
+      scan.setStopRow(Bytes.toBytes(stop))
+    }
+
+    val conf = HBaseConfiguration.create()
+    val hbaseContext = new HBaseContext(sc, conf)
+
+    var sampleIds: List[String] = inputsampleIds
+    if (sampleListFile != null) {
+      sampleIds = Source.fromFile(sampleListFile).getLines.toList
+    }
+
+    sampleIds.foreach(sampleId => {
+      scan.addColumn(Bytes.toBytes("g"), Bytes.toBytes(sampleId))
+    })
+
+    val resultHBaseRDD = hbaseContext.hbaseRDD(TableName.valueOf(hbaseTableName), scan)
+
+    val resultHBaseRDDrepar = if (numPartitions > 0) resultHBaseRDD.repartition(numPartitions)
+    else resultHBaseRDD
+
+    val result: RDD[VariantContext] = resultHBaseRDDrepar.mapPartitions((iterator) => {
+
+      val genotypeDatumReader: DatumReader[Genotype] = new SpecificDatumReader[Genotype](scala.reflect.classTag[Genotype]
+        .runtimeClass
+        .asInstanceOf[Class[Genotype]])
+
+      iterator.map((curr) => {
+        var resultList = new ListBuffer[Genotype]
+        sampleIds.foreach((sample) => {
+          val myVal = curr._2.getColumnCells(Bytes.toBytes("g"), Bytes.toBytes(sample))
+          val decoder = DecoderFactory.get().binaryDecoder(CellUtil.cloneValue(myVal.get(0)), null)
+          while (!decoder.isEnd) {
+            resultList += genotypeDatumReader.read(null, decoder)
+          }
+        })
+
+        resultList
+        val x = resultList.toList
+
+        val firstGenotype: Genotype = x.head
+        val currVar: RichVariant = RichVariant.genotypeToRichVariant(firstGenotype)
+        val currRefPos = ReferencePosition(currVar.variant.getContigName, currVar.variant.getStart)
+
+        val currVariantContext: VariantContext = new VariantContext(currRefPos, currVar, x)
+        currVariantContext
+      })
+
+    })
+
+    result
+
+  }
+
+  def createHBaseGenotypeTable(hbaseTableName: String, splitsFileName: String): Unit = {
+    val conf = HBaseConfiguration.create()
+
+    val connection = ConnectionFactory.createConnection(conf)
+
+    val admin = connection.getAdmin
+
+    val tableDescriptor = new HTableDescriptor(TableName.valueOf(hbaseTableName))
+    tableDescriptor.addFamily(new HColumnDescriptor("g".getBytes()).setCompressionType(Algorithm.GZ)
+      .setDataBlockEncoding(DataBlockEncoding.FAST_DIFF)
+      .setMaxVersions(1))
+
+    val splits: Array[Array[Byte]] = Source.fromFile(splitsFileName).getLines.toArray
+      .map(x => x.split(" "))
+      .map(g => Bytes.toBytes(g(0) + "_"
+        + String.format("%10s", g(1).toString).replace(' ', '0')))
+
+    admin.createTable(tableDescriptor, splits)
+
+    val hbaseTableName_meta = hbaseTableName + "_meta"
+
+    val tableDescriptor_meta = new HTableDescriptor(TableName.valueOf(hbaseTableName_meta))
+    tableDescriptor_meta.addFamily(new HColumnDescriptor("meta".getBytes())
+      .setCompressionType(Algorithm.GZ)
+      .setDataBlockEncoding(DataBlockEncoding.FAST_DIFF)
+      .setMaxVersions(1))
+    admin.createTable(tableDescriptor_meta)
+  }
+
+  def saveVariantContextRDDToHBase(sc: SparkContext,
+                                   vcRdd: VariantContextRDD,
+                                   hbaseTableName: String,
+                                   sequenceDictionaryId: String = null,
+                                   saveSequenceDictionary: Boolean = true): Unit = {
+
+    val conf = HBaseConfiguration.create()
+
+    val hbaseContext = new HBaseContext(sc, conf)
+
+    saveSampleMetadataToHBase(hbaseTableName + "_meta", vcRdd.samples)
+
+    if (saveSequenceDictionary) saveSequenceDictionaryToHBase(hbaseTableName + "_meta", vcRdd.sequences, sequenceDictionaryId)
+
+    val data: RDD[VariantContext] = vcRdd.rdd
+
+    val genodata = vcRdd.rdd.mapPartitions((iterator) => {
+      val genotypebaos: java.io.ByteArrayOutputStream = new ByteArrayOutputStream()
+      val genotypeEncoder = EncoderFactory.get().binaryEncoder(genotypebaos, null)
+
+      val genotypeDatumWriter: DatumWriter[Genotype] = new SpecificDatumWriter[Genotype](
+        scala.reflect.classTag[Genotype]
+          .runtimeClass
+          .asInstanceOf[Class[Genotype]])
+
+      val genotypesForHbase: Iterator[(Array[Byte], List[(String, Array[Byte])])] = iterator.map((putRecord) => {
+        val myRowKey = Bytes.toBytes(putRecord.variant.variant.getContigName + "_" + String.format("%10s", putRecord.variant.variant.getStart.toString).replace(' ', '0') + "_" +
+          putRecord.variant.variant.getReferenceAllele + "_" + putRecord.variant.variant.getAlternateAllele + "_" + putRecord.variant.variant.getEnd)
+
+        val genotypes: List[(String, Array[Byte])] = putRecord.genotypes.map((geno) => {
+          genotypebaos.reset()
+          genotypeDatumWriter.write(geno, genotypeEncoder)
+          genotypeEncoder.flush()
+          genotypebaos.flush()
+          (geno.getSampleId, genotypebaos.toByteArray)
+        }).toList
+
+        (myRowKey, genotypes)
+
+      })
+
+      genotypesForHbase
+
+    })
+
+    genodata.hbaseBulkPut(hbaseContext,
+      TableName.valueOf(hbaseTableName),
+      (putRecord) => {
+        val put = new Put(putRecord._1)
+
+        putRecord._2.foreach((x) => {
+          val sampleId: String = x._1
+          val genoBytes: Array[Byte] = x._2
+          put.addColumn(Bytes.toBytes("g"), Bytes.toBytes(sampleId), genoBytes)
+        })
+        put
+      })
+  }
+
+  def saveVariantContextRDDToHBaseBulk(sc: SparkContext,
+                                       vcRdd: VariantContextRDD,
+                                       hbaseTableName: String,
+                                       sequenceDictionaryId: String,
+                                       saveSequenceDictionary: Boolean = true,
+                                       numPartitions: Int = 0,
+                                       stagingFolder: String): Unit = {
+
+    val conf = HBaseConfiguration.create()
+    val hbaseContext = new HBaseContext(sc, conf)
+
+    saveSampleMetadataToHBase(hbaseTableName + "_meta", vcRdd.samples)
+
+    if (saveSequenceDictionary) saveSequenceDictionaryToHBase(hbaseTableName + "_meta", vcRdd.sequences, sequenceDictionaryId)
+
+    val rddMaybeRepar = if (numPartitions > 0) vcRdd.rdd.repartition(numPartitions)
+    else vcRdd.rdd
+
+    val data: RDD[VariantContext] = vcRdd.rdd
+
+    val genodata = rddMaybeRepar.mapPartitions((iterator) => {
+      val genotypebaos: java.io.ByteArrayOutputStream = new ByteArrayOutputStream()
+      val genotypeEncoder = EncoderFactory.get().binaryEncoder(genotypebaos, null)
+
+      val genotypeDatumWriter: DatumWriter[Genotype] = new SpecificDatumWriter[Genotype](
+        scala.reflect.classTag[Genotype]
+          .runtimeClass
+          .asInstanceOf[Class[Genotype]])
+
+      val genotypesForHbase: Iterator[(Array[Byte], List[(String, Array[Byte])])] = iterator.map((putRecord) => {
+        val myRowKey = Bytes.toBytes(putRecord.variant.variant.getContigName + "_" +
+          String.format("%10s", putRecord.variant.variant.getStart.toString).replace(' ', '0') + "_" +
+          putRecord.variant.variant.getReferenceAllele + "_" +
+          putRecord.variant.variant.getAlternateAllele +
+          "_" + (putRecord.variant.variant.getEnd - putRecord.variant.variant.getStart))
+
+        val genotypes: List[(String, Array[Byte])] = putRecord.genotypes.map((geno) => {
+          genotypebaos.reset()
+          genotypeDatumWriter.write(geno, genotypeEncoder)
+          genotypeEncoder.flush()
+          genotypebaos.flush()
+          (geno.getSampleId, genotypebaos.toByteArray)
+        }).toList
+
+        (myRowKey, genotypes)
+      })
+      genotypesForHbase
+    })
+
+    val familyHBaseWriterOptions = new java.util.HashMap[Array[Byte], FamilyHFileWriteOptions]
+
+    val f1Options = new FamilyHFileWriteOptions("GZ", "ROW", 128, "FAST_DIFF")
+
+    familyHBaseWriterOptions.put(Bytes.toBytes("g"), f1Options)
+
+    genodata.hbaseBulkLoad(hbaseContext,
+      TableName.valueOf(hbaseTableName),
+      t => {
+        val data = new ListBuffer[(KeyFamilyQualifier, Array[Byte])]
+        val rowKey = t._1
+        val family = Bytes.toBytes("g")
+
+        t._2.foreach((x) => {
+
+          val qualifier: String = x._1
+          val value: Array[Byte] = x._2
+
+          val mykeyFamilyQualifier = new KeyFamilyQualifier(rowKey, family, Bytes.toBytes(qualifier))
+
+          data += Tuple2(mykeyFamilyQualifier, value)
+
+        })
+
+        data.iterator
+      },
+      stagingFolder, familyHBaseWriterOptions,
+      compactionExclude = false,
+      HConstants.DEFAULT_MAX_FILE_SIZE)
+    ("hadoop fs -chmod -R 777 " + stagingFolder) !
+    val conn = ConnectionFactory.createConnection(conf)
+    val load = new LoadIncrementalHFiles(conf)
+    load.doBulkLoad(new Path(stagingFolder), conn.getAdmin, conn.getTable(TableName.valueOf(hbaseTableName)), conn.getRegionLocator(TableName.valueOf(hbaseTableName)))
+
+  }
+
+  def deleteGenotypeSamplesFromHBase(sc: SparkContext,
+                                     hbaseTableName: String,
+                                     inputsampleIds: List[String] = null,
+                                     sampleListFile: String = null,
+                                     numPartitions: Int = 0): Unit = {
+
+    val scan = new Scan()
+    scan.setCaching(100)
+    scan.setMaxVersions(1)
+
+    val conf = HBaseConfiguration.create()
+    val hbaseContext = new HBaseContext(sc, conf)
+
+    var sampleIds: List[String] = inputsampleIds
+    if (sampleListFile != null) {
+      sampleIds = Source.fromFile(sampleListFile).getLines.toList
+    }
+
+    sampleIds.foreach(sampleId => {
+      scan.addColumn(Bytes.toBytes("g"), Bytes.toBytes(sampleId))
+    })
+
+    val resultHBaseRDD: RDD[(ImmutableBytesWritable, Result)] = hbaseContext.hbaseRDD(TableName.valueOf(hbaseTableName), scan)
+
+    hbaseContext.bulkDelete[(ImmutableBytesWritable, Result)](resultHBaseRDD,
+      TableName.valueOf(hbaseTableName),
+      putRecord => {
+        val currDelete = new Delete(putRecord._2.getRow)
+        sampleIds.foreach((sample) => {
+          currDelete.addColumns(Bytes.toBytes("g"), Bytes.toBytes(sample))
+        })
+        currDelete
+      },
+      4)
+
+  }
+
+  def loadGenotypesFromHBaseToGenotypeRDD(sc: SparkContext,
+                                          hbaseTableName: String,
+                                          sampleIds: List[String],
+                                          sequenceDictionaryId: String,
+                                          queryRegion: ReferenceRegion = null, //one-based
+                                          numPartitions: Int = 0): GenotypeRDD = {
+
+    val sequenceDictionary = loadSequenceDictionaryFromHBase(hbaseTableName + "_meta", sequenceDictionaryId)
+    val sampleMetadata = loadSampleMetadataFromHBase(hbaseTableName + "_meta", sampleIds)
+    val genotypes = loadRDDofGenotypeFromHBase(sc, hbaseTableName, sampleIds, queryRegion, numPartitions)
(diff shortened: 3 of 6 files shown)


adam-cli/src/main/scala/org/bdgenomics/adam/cli/Vcf2HBase.scala
```diff
@@ -0,0 +1,82 @@
...
+  @Args4jOption(required = true, name = "-hbase_table", usage = "HBase table name in which to load VCF file")
+  var hbaseTable: String = null
+
+  @Args4jOption(required = true, name = "-seq_dict_id", usage = "User defined name to apply to the sequence dictionary create from this VCF, or name of existing sequence dictionary to be used")
+  var seqDictId: String = null
+
+  @Args4jOption(required = false, name = "-use_existing_seqdict", usage = "Use an existing sequence dictionary, don't write a new one")
+  var useExistingSeqDict: Boolean = false
+
+  @Args4jOption(required = false, name = "-repartition_num", usage = "Reparition into N partitions prior to writing to HBase")
```
heuermh: typo in usage `Reparition` &rarr; `Repartition`.  Also, might `-partitions` be a better name for this argument?

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Vcf2HBase.scala
```diff
@@ -0,0 +1,82 @@
...
+  var repartitionNum: Int = 0
+
+  @Args4jOption(required = true, name = "-staging_folder", usage = "location for temporary files during bulk load")
```
heuermh: `location` &rarr; `Location` in usage

adam-core/pom.xml
```diff
@@ -137,6 +137,10 @@
       <artifactId>spark-core_${scala.version.prefix}</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.apache.spark</groupId>
+      <artifactId>spark-streaming_2.10</artifactId>
```
heuermh: use `${scala.version.prefix}`

adam-core/pom.xml
```diff
@@ -185,5 +189,37 @@
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
     </dependency>
+
```
heuermh: remove extra blank lines

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
+package org.bdgenomics.adam.hbase
```
heuermh: missing license header

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+import org.bdgenomics.adam.models.{ ReferenceRegion, ReferencePosition, SequenceDictionary, VariantContext }
+import scala.io.Source
+
+import scala.collection.mutable.ListBuffer
+import sys.process._
+
+import scala.collection.JavaConverters._
+
+/**
+ * Created by jp on 7/23/16.
```
heuermh: improve doc comment

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+ */
+
+object HBaseFunctions {
+
+  def saveSequenceDictionaryToHBase(hbaseTableName: String,
```
heuermh: add method doc

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+
+    encoder.flush()
+    baos.flush()
+
+    val put: Put = new Put(Bytes.toBytes(sequenceDictionaryId))
+    put.addColumn(Bytes.toBytes("meta"), Bytes.toBytes("contigdata"), baos.toByteArray)
+    table.put(put)
+  }
+
+  def loadSequenceDictionaryFromHBase(hbaseTableName: String,
```
heuermh: add method doc

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+                                      sequenceDictionaryId: String): SequenceDictionary = {
+
+    val conf = HBaseConfiguration.create()
+    val connection = ConnectionFactory.createConnection(conf)
```
heuermh: is there much penalty to creating a new connection in each method?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+    var resultList = new ListBuffer[Contig]
+    while (!decoder.isEnd) {
+      resultList += contigDatumReader.read(null, decoder)
+    }
+
+    SequenceDictionary.fromAvro(resultList)
+
+  }
+
+  def saveSampleMetadataToHBase(hbaseTableName: String,
```
heuermh: add method doc

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+      val curr_sampleid = x.getSampleId
+
+      val put = new Put(Bytes.toBytes(curr_sampleid))
+      put.addColumn(Bytes.toBytes("meta"), Bytes.toBytes("sampledata"), baos.toByteArray)
+      table.put(put)
+
+    })
+  }
+
+  def loadSampleMetadataFromHBase(hbaseTableName: String,
```
heuermh: add method doc; change method name to `loadSamplesFromHBase`

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+      val contigDatumWriter: DatumWriter[Contig] =
+        new SpecificDatumWriter[Contig](scala.reflect.classTag[Contig].runtimeClass.asInstanceOf[Class[Contig]])
+      contigDatumWriter.write(x, encoder)
+    })
+
+    encoder.flush()
+    baos.flush()
+
+    val put: Put = new Put(Bytes.toBytes(sequenceDictionaryId))
+    put.addColumn(Bytes.toBytes("meta"), Bytes.toBytes("contigdata"), baos.toByteArray)
```
heuermh: here and elsewhere, these strings are column names?  If so I would drop `data` and name the columns more simply, `meta`, `contig`, `sample`, etc.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+                                 sampleIds: List[String],
+                                 queryRegion: ReferenceRegion = null,
+                                 numPartitions: Int = 0): RDD[Genotype] = {
+
+    val scan = new Scan()
+    scan.setCaching(100)
+    scan.setMaxVersions(1)
+
+    if (queryRegion != null) {
+      val start = queryRegion.referenceName + "_" + String.format("%10s", queryRegion.start.toString).replace(' ', '0')
```
heuermh: are we wheat-safe? https://www.biostars.org/p/12560/  :)

This should be extracted out using the strategy pattern to allow other key definitions.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+
+    resultList
+
+  }
+
+  def loadRDDofGenotypeFromHBase(sc: SparkContext,
+                                 hbaseTableName: String,
+                                 sampleIds: List[String],
+                                 queryRegion: ReferenceRegion = null,
+                                 numPartitions: Int = 0): RDD[Genotype] = {
```
heuermh: `numPartitions` &rarr; `partitions`

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+        resultList += sampleDatumReader.read(null, decoder)
+      }
+
+    })
+
+    resultList
+
+  }
+
+  def loadRDDofGenotypeFromHBase(sc: SparkContext,
```
heuermh: add method doc; change method name to `loadGenotypesFromHBase`

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+          }
+        })
+        resultList
+      })
+
+    })
+    result
+  }
+
+  def loadRDDofVariantContextFromHBase(sc: SparkContext,
```
heuermh: add method doc; change method name to `loadVariantContextsFromHBase`; possibly limit visibility to package-private

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+                                       hbaseTableName: String,
+                                       inputsampleIds: List[String] = null,
```
heuermh: `inputsampleIds` &rarr; `sampleIds`

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+                                       sampleListFile: String = null,
+                                       queryRegion: ReferenceRegion = null,
+                                       numPartitions: Int = 0): RDD[VariantContext] = {
```
heuermh: `numPartitions` &rarr; `partitions`

jpdna:
Can someone help point me to which log I should be looking at to diagnose this build failure? Thx!

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+    val result: RDD[VariantContext] = resultHBaseRDDrepar.mapPartitions((iterator) => {
+
+      val genotypeDatumReader: DatumReader[Genotype] = new SpecificDatumReader[Genotype](scala.reflect.classTag[Genotype]
+        .runtimeClass
+        .asInstanceOf[Class[Genotype]])
+
+      iterator.map((curr) => {
+        var resultList = new ListBuffer[Genotype]
+        sampleIds.foreach((sample) => {
+          val myVal = curr._2.getColumnCells(Bytes.toBytes("g"), Bytes.toBytes(sample))
```
heuermh: myXxx is not a very good variable name style

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+        currVariantContext
+      })
+
+    })
+
+    result
+
+  }
+
+  def createHBaseGenotypeTable(hbaseTableName: String, splitsFileName: String): Unit = {
```
heuermh: add method doc

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+      .setMaxVersions(1))
+
+    val splits: Array[Array[Byte]] = Source.fromFile(splitsFileName).getLines.toArray
+      .map(x => x.split(" "))
+      .map(g => Bytes.toBytes(g(0) + "_"
+        + String.format("%10s", g(1).toString).replace(' ', '0')))
+
+    admin.createTable(tableDescriptor, splits)
+
+    val hbaseTableName_meta = hbaseTableName + "_meta"
```
heuermh: keep to camelCase

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+
+    val tableDescriptor_meta = new HTableDescriptor(TableName.valueOf(hbaseTableName_meta))
+    tableDescriptor_meta.addFamily(new HColumnDescriptor("meta".getBytes())
+      .setCompressionType(Algorithm.GZ)
+      .setDataBlockEncoding(DataBlockEncoding.FAST_DIFF)
+      .setMaxVersions(1))
+    admin.createTable(tableDescriptor_meta)
+  }
+
+  def saveVariantContextRDDToHBase(sc: SparkContext,
```
heuermh: add method doc; change method name to `saveVariantContextsToHBase`; possibly limit visibility to package-private

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+    val hbaseContext = new HBaseContext(sc, conf)
+
+    saveSampleMetadataToHBase(hbaseTableName + "_meta", vcRdd.samples)
+
+    if (saveSequenceDictionary) saveSequenceDictionaryToHBase(hbaseTableName + "_meta", vcRdd.sequences, sequenceDictionaryId)
+
+    val data: RDD[VariantContext] = vcRdd.rdd
+
+    val genodata = vcRdd.rdd.mapPartitions((iterator) => {
+      val genotypebaos: java.io.ByteArrayOutputStream = new ByteArrayOutputStream()
```
heuermh: use camelCase

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+      val genotypeEncoder = EncoderFactory.get().binaryEncoder(genotypebaos, null)
+
+      val genotypeDatumWriter: DatumWriter[Genotype] = new SpecificDatumWriter[Genotype](
+        scala.reflect.classTag[Genotype]
+          .runtimeClass
+          .asInstanceOf[Class[Genotype]])
+
+      val genotypesForHbase: Iterator[(Array[Byte], List[(String, Array[Byte])])] = iterator.map((putRecord) => {
+        val myRowKey = Bytes.toBytes(putRecord.variant.variant.getContigName + "_" + String.format("%10s", putRecord.variant.variant.getStart.toString).replace(' ', '0') + "_" +
```
heuermh: `myRowKey` &rarr; `rowKey`, refactor key building out

fnothaft:
See https://amplab.cs.berkeley.edu/jenkins/job/ADAM-prb/1558/HADOOP_VERSION=2.6.0,SCALAVER=2.10,SPARK_VERSION=1.5.2,label=centos/console. Specifically:

```
++ git status --porcelain
+ test -n ' M adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala'
+ echo 'Please run '\''./scripts/format-source'\'''
Please run './scripts/format-source'
+ exit 1
Build step 'Execute shell' marked build as failure
```

A nice easy one to fix ;).

I'll make a review pass later today.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+
+    val hbaseContext = new HBaseContext(sc, conf)
+
+    saveSampleMetadataToHBase(hbaseTableName + "_meta", vcRdd.samples)
+
+    if (saveSequenceDictionary) saveSequenceDictionaryToHBase(hbaseTableName + "_meta", vcRdd.sequences, sequenceDictionaryId)
+
+    val data: RDD[VariantContext] = vcRdd.rdd
+
+    val genodata = vcRdd.rdd.mapPartitions((iterator) => {
```
heuermh: `genodata` &rarr; `genotypePartitions` ?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+        putRecord._2.foreach((x) => {
+          val sampleId: String = x._1
+          val genoBytes: Array[Byte] = x._2
+          put.addColumn(Bytes.toBytes("g"), Bytes.toBytes(sampleId), genoBytes)
+        })
+        put
+      })
+  }
+
+  def saveVariantContextRDDToHBaseBulk(sc: SparkContext,
```
heuermh: add method doc; change method name to `saveVariantContextsToHBaseBulk`, possibly limit visibility to package-private

Why are these *Bulk methods necessary?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+    val hbaseContext = new HBaseContext(sc, conf)
+
+    saveSampleMetadataToHBase(hbaseTableName + "_meta", vcRdd.samples)
+
+    if (saveSequenceDictionary) saveSequenceDictionaryToHBase(hbaseTableName + "_meta", vcRdd.sequences, sequenceDictionaryId)
+
+    val rddMaybeRepar = if (numPartitions > 0) vcRdd.rdd.repartition(numPartitions)
+    else vcRdd.rdd
+
+    val data: RDD[VariantContext] = vcRdd.rdd
```
heuermh: `data` &rarr; `variantContextRdd`

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+
+    val genodata = rddMaybeRepar.mapPartitions((iterator) => {
```
heuermh: this block looks similar to code above

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+      val genotypebaos: java.io.ByteArrayOutputStream = new ByteArrayOutputStream()
+      val genotypeEncoder = EncoderFactory.get().binaryEncoder(genotypebaos, null)
+
+      val genotypeDatumWriter: DatumWriter[Genotype] = new SpecificDatumWriter[Genotype](
+        scala.reflect.classTag[Genotype]
+          .runtimeClass
+          .asInstanceOf[Class[Genotype]])
+
+      val genotypesForHbase: Iterator[(Array[Byte], List[(String, Array[Byte])])] = iterator.map((putRecord) => {
+        val myRowKey = Bytes.toBytes(putRecord.variant.variant.getContigName + "_" +
```
heuermh: refactor out key building

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+      compactionExclude = false,
+      HConstants.DEFAULT_MAX_FILE_SIZE)
+    ("hadoop fs -chmod -R 777 " + stagingFolder) !
+    val conn = ConnectionFactory.createConnection(conf)
+    val load = new LoadIncrementalHFiles(conf)
+    load.doBulkLoad(new Path(stagingFolder), conn.getAdmin, conn.getTable(TableName.valueOf(hbaseTableName)), conn.getRegionLocator(TableName.valueOf(hbaseTableName)))
+
+  }
+
+  def deleteGenotypeSamplesFromHBase(sc: SparkContext,
```
heuermh: is deleting an operation we need to support?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+        sampleIds.foreach((sample) => {
+          currDelete.addColumns(Bytes.toBytes("g"), Bytes.toBytes(sample))
+        })
+        currDelete
+      },
+      4)
+
+  }
+
+  def loadGenotypesFromHBaseToGenotypeRDD(sc: SparkContext,
```
heuermh: add method doc; change method name to `loadGenotypesFromHBase`

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,508 @@
...
+                                          queryRegion: ReferenceRegion = null, //one-based
+                                          numPartitions: Int = 0): GenotypeRDD = {
+
+    val sequenceDictionary = loadSequenceDictionaryFromHBase(hbaseTableName + "_meta", sequenceDictionaryId)
+    val sampleMetadata = loadSampleMetadataFromHBase(hbaseTableName + "_meta", sampleIds)
+    val genotypes = loadRDDofGenotypeFromHBase(sc, hbaseTableName, sampleIds, queryRegion, numPartitions)
+    GenotypeRDD(genotypes, sequenceDictionary, sampleMetadata)
+  }
+
+  def loadGenotypesFromHBaseToVariantContextRDD(sc: SparkContext,
```
heuermh: I'm not sure we need/want this method.

The way I've been interpreting the current API design is that `VariantContext` is only an intermediate `o.b.a.models` class for converting to/from htsjdk `VariantContext`, not really something that we want users to focus on.

pom.xml
```diff
@@ -1,5 +1,15 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+
+  <repositories>
+    <repository>
+      <id>cloudera</id>
+      <url>https://repository.cloudera.com/artifactory/cloudera-repos/</url>
```
heuermh: Oh, this could be a major problem.  We can't* deploy to Maven Central with non-Maven Central repository dependencies.

* - "can't" here may strictly be too strong a word; I'll have to investigate

pom.xml
```diff
@@ -1,5 +1,15 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+
```
heuermh: remove extra blank lines

pom.xml
```diff
@@ -464,6 +474,22 @@
         </exclusions>
       </dependency>
       <dependency>
+        <groupId>org.apache.spark</groupId>
+        <artifactId>spark-streaming_2.10</artifactId>
```
heuermh: use `${scala.version.prefix}`

pom.xml
```diff
@@ -564,6 +590,74 @@
         <artifactId>scala-guice_${scala.version.prefix}</artifactId>
         <version>4.0.1</version>
       </dependency>
+
```
heuermh: remove commented out code, create an issue to add these back in later

pom.xml
```diff
@@ -564,6 +590,74 @@
...
+        <groupId>org.apache.hbase</groupId>
+        <artifactId>hbase-common</artifactId>
+        <version>2.0.0-SNAPSHOT</version>
+      </dependency>  -->
+
+
+      <dependency>
+        <groupId>org.apache.hbase</groupId>
+        <artifactId>hbase-hadoop2-compat</artifactId>
+        <version>1.2.0-cdh5.8.0</version>
```
heuermh: refactor version to a variable `${hbase.version}` here and below

heuermh:
Nice work!  I made a review pass focused on API and style.

If the external Maven repository is a problem, we may have to stick with Apache HBase rather than CDH versions, even if that means waiting for them to release Spark support.

In order to test & validate the code changes, I'll need some doc on how to run and test it.  :)

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Vcf2HBase.scala
```diff
@@ -0,0 +1,82 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.bdgenomics.adam.cli
+
+import org.apache.spark.SparkContext
+
```
fnothaft: Nit: remove whitespace.

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Vcf2HBase.scala
```diff
@@ -0,0 +1,82 @@
...
+    new Vcf2HBase(Args4j[Vcf2HBaseArgs](cmdLine))
+  }
+}
+
+class Vcf2HBaseArgs extends Args4jBase {
+
+  @Argument(required = true, metaVar = "VCF", usage = "The VCF file to convert", index = 0)
+  var vcfPath: String = _
+
+  @Args4jOption(required = true, name = "-hbase_table", usage = "HBase table name in which to load VCF file")
```
fnothaft: By convention, we typically make `required = true` options as `@Argument`s.

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Vcf2HBase.scala
fnothaft: IIRC, we've used `-repartition` across other commands, e.g., `Transform` (which has both `-coalesce` and `-repartition`). I like `-repartition` because it implies the cost of the operation: we take a shuffle.

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Vcf2HBase.scala
```diff
@@ -0,0 +1,82 @@
...
+  var hbaseTable: String = null
+
+  @Args4jOption(required = true, name = "-seq_dict_id", usage = "User defined name to apply to the sequence dictionary create from this VCF, or name of existing sequence dictionary to be used")
+  var seqDictId: String = null
+
+  @Args4jOption(required = false, name = "-use_existing_seqdict", usage = "Use an existing sequence dictionary, don't write a new one")
```
fnothaft: This argument seems to be unused?

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Vcf2HBase.scala
```diff
@@ -0,0 +1,82 @@
...
+  val companion = Transform
+
+  def run(sc: SparkContext) {
+
+    val vcRdd = sc.loadVcf(args.vcfPath)
+
+    HBaseFunctions.saveVariantContextRDDToHBaseBulk(sc,
+      vcRdd,
+      args.hbaseTable,
+      sequenceDictionaryId = args.seqDictId,
```
fnothaft: I'd prefer to wrap all of the optional arguments in `Option`s, as some of them will be null if not provided.

adam-core/pom.xml
fnothaft: OOC, why do we depend on Spark streaming here?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
fnothaft: The license header shows up as there for me?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.bdgenomics.adam.hbase
+
+import java.util
+import org.apache.hadoop.hbase.io.ImmutableBytesWritable
+import org.apache.hadoop.hbase.io.encoding.DataBlockEncoding
+import org.apache.spark.SparkContext
```
fnothaft: Ordering nit: `org.apache.spark` after `org.apache.hadoop`

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+import org.apache.spark.rdd.RDD
+import org.apache.hadoop.hbase.client._
```
fnothaft: I would prefer to not use `._` imports unless we're importing implicits or we are importing something like 8 classes from a package.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+import org.apache.hadoop.hbase._
+import org.apache.hadoop.hbase.spark._
+import org.bdgenomics.adam.rdd.variation.{ GenotypeRDD, VariantContextRDD }
+import org.bdgenomics.adam.rich.RichVariant
+import org.bdgenomics.formats.avro._
```
fnothaft: Ordering nit: `org.bdgenomics` after `org.apache`.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+import org.apache.hadoop.hbase.spark.HBaseRDDFunctions._
+import org.apache.hadoop.hbase.util.Bytes
+import org.apache.hadoop.hbase.{ HBaseConfiguration, TableName }
+import org.apache.avro.specific.SpecificDatumWriter
```
fnothaft: Ordering nit: `org.apache.avro` before `org.apache.hadoop`.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+import org.apache.avro.io._
+import java.io.ByteArrayOutputStream
```
fnothaft: Ordering nit: `org` after `java`

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+
+import scala.collection.mutable.ListBuffer
+import sys.process._
+
+import scala.collection.JavaConverters._
+
+/**
+ * Created by jp on 7/23/16.
+ */
+
```
fnothaft: Nit: no space

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
fnothaft: +1

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+  def saveSequenceDictionaryToHBase(hbaseTableName: String,
+                                    sequences: SequenceDictionary,
+                                    sequenceDictionaryId: String): Unit = {
+
+    val conf = HBaseConfiguration.create()
+    val connection = ConnectionFactory.createConnection(conf)
+    val table: Table = connection.getTable(TableName.valueOf(hbaseTableName))
+    val contigs = sequences.toAvro
+
+    val baos: java.io.ByteArrayOutputStream = new ByteArrayOutputStream()
```
fnothaft: Shouldn't need the type here.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+
+object HBaseFunctions {
+
+  def saveSequenceDictionaryToHBase(hbaseTableName: String,
+                                    sequences: SequenceDictionary,
+                                    sequenceDictionaryId: String): Unit = {
+
+    val conf = HBaseConfiguration.create()
+    val connection = ConnectionFactory.createConnection(conf)
+    val table: Table = connection.getTable(TableName.valueOf(hbaseTableName))
```
fnothaft: Shouldn't need the type here.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
fnothaft: Can you doc this section of the code? If I read this right, you are making an avro encoder that writes bytes into an array, so that you can then write Avro records into this array, which you then copy to HBase?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+    val contigs = sequences.toAvro
+
+    val baos: java.io.ByteArrayOutputStream = new ByteArrayOutputStream()
+    val encoder = EncoderFactory.get().binaryEncoder(baos, null)
+
+    contigs.foreach((x) => {
```
fnothaft: Stylistically, drop parens around `x` in `.foreach((x) => {`.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+      val contigDatumWriter: DatumWriter[Contig] =
```
fnothaft: I would factor the DatumWriter creation out of the for loop.

fnothaft: Also, shouldn't `scala.reflect.classTag[Contig].runtimeClass.asInstanceOf[Class[Contig]]` reduce down to `classOf[Contig]`?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
fnothaft: +1

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
fnothaft: +1

Also, I would factor the connection creation code out into a class/package private method.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+    val conf = HBaseConfiguration.create()
+    val connection = ConnectionFactory.createConnection(conf)
+    val table = connection.getTable(TableName.valueOf(hbaseTableName))
+
+    val myGet = new Get(Bytes.toBytes(sequenceDictionaryId))
+    val result = table.get(myGet)
+    val dataBytes = result.getValue(Bytes.toBytes("meta"), Bytes.toBytes("contigdata"))
+
+    val contigDatumReader: DatumReader[Contig] =
+      new SpecificDatumReader[Contig](scala.reflect.classTag[Contig]
```
fnothaft: Shouldn't `scala.reflect...` translate down to `classOf[Contig]`?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+        .asInstanceOf[Class[Contig]])
+
+    val decoder = DecoderFactory.get().binaryDecoder(dataBytes, null)
+    var resultList = new ListBuffer[Contig]
+    while (!decoder.isEnd) {
+      resultList += contigDatumReader.read(null, decoder)
+    }
+
+    SequenceDictionary.fromAvro(resultList)
+
```
fnothaft: Nit: extra space.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+  def saveSampleMetadataToHBase(hbaseTableName: String,
+                                samples: Seq[Sample]): Unit = {
+
+    val conf = HBaseConfiguration.create()
+    val connection = ConnectionFactory.createConnection(conf)
+    val table = connection.getTable(TableName.valueOf(hbaseTableName))
+
+    samples.foreach((x) => {
+      val baos: java.io.ByteArrayOutputStream = new ByteArrayOutputStream()
```
fnothaft: This `ByteArrayOutputStream` code is also repeated, let's factor it out.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
fnothaft: I kinda prefer the `loadSampleMetadataFromHBase` name, since to me, `loadSamplesFromHBase` sounds more like "load the genotypes for samples x,y,z from HBase", instead of "load the sample descriptions for samples x,y,z from HBase".

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+        new SpecificDatumReader[Sample](scala.reflect.classTag[Sample]
+          .runtimeClass
+          .asInstanceOf[Class[Sample]])
+
+      val decoder = DecoderFactory.get().binaryDecoder(dataBytes, null)
+
+      while (!decoder.isEnd) {
+        resultList += sampleDatumReader.read(null, decoder)
+      }
+
```
fnothaft: Nit: whitespace.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+    })
+
+    resultList
+
```
fnothaft: Nit: whitespace.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+    val table = connection.getTable(TableName.valueOf(hbaseTableName))
+
+    var resultList = new ListBuffer[Sample]
+
+    sampleIds.foreach((sampleId) => {
+      val myGet = new Get(Bytes.toBytes(sampleId))
+      val result = table.get(myGet)
+      val dataBytes = result.getValue(Bytes.toBytes("meta"), Bytes.toBytes("sampledata"))
+
+      val sampleDatumReader: DatumReader[Sample] =
```
fnothaft: Also, I think you can factor the code from `val sampleDatumReader`...`resultList` out into a function as well. It seems like it is replicated several places.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
fnothaft: +1

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+
+    })
+
+    resultList
+
+  }
+
+  def loadRDDofGenotypeFromHBase(sc: SparkContext,
+                                 hbaseTableName: String,
+                                 sampleIds: List[String],
```
fnothaft: Prefer either `Seq` or `Iterable` to `List`.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+                                 queryRegion: ReferenceRegion = null,
```
fnothaft: Please wrap in `Option` and don't pass nulls.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+                                 numPartitions: Int = 0): RDD[Genotype] = {
```
fnothaft: Ditto here. Pass an option instead of an uninformative 0 value.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+
+    val scan = new Scan()
+    scan.setCaching(100)
```
fnothaft: What are these parameters, and can we expose them so that they are tunable?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
fnothaft: Where possible, prefer using format: `"%s_%d".format(queryRegion.referenceName, queryRegion.start)`

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+      val stop = queryRegion.referenceName + "_" + String.format("%10s", queryRegion.end.toString).replace(' ', '0')
+      scan.setStartRow(Bytes.toBytes(start))
+      scan.setStopRow(Bytes.toBytes(stop))
+    }
+
+    val conf = HBaseConfiguration.create()
+    val hbaseContext = new HBaseContext(sc, conf)
+
+    sampleIds.foreach(sampleId => {
+      scan.addColumn(Bytes.toBytes("g"), Bytes.toBytes(sampleId))
```
fnothaft: What's the `"g"` come from?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+    })
+
+    val resultHBaseRDD = hbaseContext.hbaseRDD(TableName.valueOf(hbaseTableName), scan)
+
+    val resultHBaseRDDrepar = if (numPartitions > 0) resultHBaseRDD.repartition(numPartitions)
+    else resultHBaseRDD
+
+    val result: RDD[Genotype] = resultHBaseRDDrepar.mapPartitions((iterator) => {
```
fnothaft: What is going on here? Like, what is inside the RDD that we are repartitioning and then mapPartitionsing over?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
fnothaft: I would remove this method wholesale and prefer `loadGenotypesFromHBase` chained to `toVariantContextRDD`.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
fnothaft: +1, and as an Option.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
fnothaft: All arguments after `inputsampleIds` should be wrapped in `Option`s, we should avoid passing nulls wherever possible.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
fnothaft: +1

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+        var resultList = new ListBuffer[Genotype]
+        sampleIds.foreach((sample) => {
+          val myVal = curr._2.getColumnCells(Bytes.toBytes("g"), Bytes.toBytes(sample))
+          val decoder = DecoderFactory.get().binaryDecoder(CellUtil.cloneValue(myVal.get(0)), null)
+          while (!decoder.isEnd) {
+            resultList += genotypeDatumReader.read(null, decoder)
+          }
+        })
+
+        resultList
```
fnothaft: This line seems to be unused?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+        val x = resultList.toList
```
fnothaft: Should have a better name.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
fnothaft: Prefer `) {` to `): Unit = {`.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+      })
+
+    })
+
+    result
+
+  }
+
+  def createHBaseGenotypeTable(hbaseTableName: String, splitsFileName: String): Unit = {
+    val conf = HBaseConfiguration.create()
```
fnothaft: Please document what's going on here inline.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+
+    val connection = ConnectionFactory.createConnection(conf)
+
+    val admin = connection.getAdmin
+
+    val tableDescriptor = new HTableDescriptor(TableName.valueOf(hbaseTableName))
+    tableDescriptor.addFamily(new HColumnDescriptor("g".getBytes()).setCompressionType(Algorithm.GZ)
```
fnothaft: Can we expose these all as configurable?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+      .setDataBlockEncoding(DataBlockEncoding.FAST_DIFF)
+      .setMaxVersions(1))
+
+    val splits: Array[Array[Byte]] = Source.fromFile(splitsFileName).getLines.toArray
```
fnothaft: What's this code do? Please add docs inline.

fnothaft: Also, `Source.fromFile` won't work reliably in YARN cluster mode.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+      .map(g => Bytes.toBytes(g(0) + "_"
+        + String.format("%10s", g(1).toString).replace(' ', '0')))
+
+    admin.createTable(tableDescriptor, splits)
+
+    val hbaseTableName_meta = hbaseTableName + "_meta"
+
+    val tableDescriptor_meta = new HTableDescriptor(TableName.valueOf(hbaseTableName_meta))
+    tableDescriptor_meta.addFamily(new HColumnDescriptor("meta".getBytes())
+      .setCompressionType(Algorithm.GZ)
```
fnothaft: Can these be exposed as configurable?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
fnothaft: Again, I would remove this function and prefer chaining `VariantContextRDD.toGenotypeRDD.saveGenotypesToHBase`.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
fnothaft: Remove method, funnel calls through `toGenotypeRdd.saveGenotypesToHBaseBulk`.

> Why are these *Bulk methods necessary?

IIRC the bulk puts are much more efficient.
That being said, ideally, we should have one-or-the-other, not both.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+        }).toList
+
+        (myRowKey, genotypes)
+      })
+      genotypesForHbase
+    })
+
+    val familyHBaseWriterOptions = new java.util.HashMap[Array[Byte], FamilyHFileWriteOptions]
+
+    val f1Options = new FamilyHFileWriteOptions("GZ", "ROW", 128, "FAST_DIFF")
```
fnothaft: Make configurable.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+
+    familyHBaseWriterOptions.put(Bytes.toBytes("g"), f1Options)
+
+    genodata.hbaseBulkLoad(hbaseContext,
+      TableName.valueOf(hbaseTableName),
+      t => {
```
fnothaft: What is going on here? Please add inline docs.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,525 @@
...
+          data += Tuple2(mykeyFamilyQualifier, value)
+
+        })
+
+        data.iterator
+      },
+      stagingFolder, familyHBaseWriterOptions,
+      compactionExclude = false,
+      HConstants.DEFAULT_MAX_FILE_SIZE)
+    ("hadoop fs -chmod -R 777 " + stagingFolder) !
```
fnothaft: Nonononono! Why?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
fnothaft: Agreed; can't we just rely on the HBase CLI for this?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
fnothaft: +1

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
fnothaft: +1, it is purely for going to/from `htsjdk.variant.VariantContext`.

pom.xml
fnothaft: +1

pom.xml
fnothaft: +1, and remove spaces between dependencies

jpdna:
If I add `<scope>provided</scope>`
to the hbase dependencies in the project root pom.xml
it compiles fine, but at runtime on the bdg CDH cluster when trying to connect to hbase I get:
`java.lang.NoClassDefFoundError: org/apache/hadoop/hbase/HBaseConfiguration`

The needed jars do exist on cluster like at:
`/opt/cloudera/parcels/CDH-5.8.0-1.cdh5.8.0.p0.42/lib/hbase/hbase-hadoop2-compat.jar`

I am not sure what is the correct way to get these included in the classpath such that `provided` would work, and how we would explain what is needed to users.

fnothaft:
I think you'd just provide a `--conf spark.jars=/opt/cloudera/parcels/CDH-5.8.0-1.cdh5.8.0.p0.42/lib/hbase/hbase-hadoop2-compat.jar` before `--` in the adam-submit command line. If this works, we should roll this in to the #493 docs.

jpdna:
for me the `--conf spark.jars=....` appeared to not actually be changing the spark.jars parameter when checking the spark GUI environment page.

I got closer with the following, but then ran into a problem I can't resolve:

I start the adam-shell with:
```
../adam/bin/adam-shell --master yarn-client --num-executors 1 \
 --executor-cores 2 --executor-memory 20g --jars /spark/opt/cloudera/parcels/CDH-5.8.0-1.cdh5.8.0.p0.42/lib/hbase/hbase-hadoop2-compat-1.2.0-cdh5.8.0.jar,\
/opt/cloudera/parcels/CDH-5.8.0-1.cdh5.8.0.p0.42/lib/hbase/hbase-server-1.2.0-cdh5.8.0.jar, 
\
/opt/cloudera/parcels/CDH-5.8.0-1.cdh5.8.0.p0.42/lib/hbase/hbase-it-1.2.0-cdh5.8.0.jar,/opt/cloudera/parcels/CDH-5.8.0-1.cdh5.8.0.p0.42/lib/hbase/hbase-it-1.2.0-cdh5.8.0.jar, \ 
/opt/cloudera/parcels/CDH-5.8.0-1.cdh5.8.0.p0.42/lib/hbase/hbase-client-1.2.0-cdh5.8.0.jar, \ 
/opt/cloudera/parcels/CDH-5.8.0-1.cdh5.8.0.p0.42/lib/hbase/hbase-spark-1.2.0-cdh5.8.0.jar,/home/eecs/jpaschall/work/hbase/v12/adam/adam-assembly/target/adam_2.10-0.20.1-SNAPSHOT.jar, \
/opt/cloudera/parcels/CDH-5.8.0-1.cdh5.8.0.p0.42/lib/hbase/lib/hbase-common-1.2.0-cdh5.8.0.jar,/ \
opt/cloudera/parcels/CDH-5.8.0-1.cdh5.8.0.p0.42/lib/hbase/lib/protobuf-java-2.5.0.jar
```

and looking in the UI spark.jars does contain all these paths.

I am even able to:
```
import org.bdgenomics.adam.rdd.ADAMContext
import org.bdgenomics.adam.hbase.HBaseFunctions
import org.apache.hadoop.hbase.HBaseConfiguration
```
the last of which wouldn't work until adding those jars for hbase
However when I then run the first function that uses hbase:
```
HBaseFunctions.createHBaseGenotypeTable("nov8_2", "/home/eecs/jpaschall/work/hbase/v3/run1/splits.chr22")

```
I get error below:
```
scala> HBaseFunctions.createHBaseGenotypeTable("nov8_2", "/home/eecs/jpaschall/work/hbase/v3/run1/splits.chr22")
java.lang.NoClassDefFoundError: org/apache/hadoop/hbase/protobuf/generated/MasterProtos$MasterService$BlockingInterface
	at java.lang.Class.forName0(Native Method)
```
which despite including that jar `/opt/cloudera/parcels/CDH-5.8.0-1.cdh5.8.0.p0.42/lib/hbase/lib/protobuf-java-2.5.0.jar`

I can't seem to resolve.

For the moment I think I will work on other aspects of the PR and see if others have a suggestion, as it works fine when  dependencies are included without `<scope>provided</scope>`
I know it is unfortunate to package these dependencies like this when the should be able to be provided, and may motivate us to keep the hbase work a separate module or repo as I understand if these dependencies must be included.

fnothaft:
@jpdna The class you are looking for is in /opt/cloudera/parcels/CDH-5.8.0-1.cdh5.8.0.p0.42/lib/hbase/lib/hbase-protocol-1.2.0-cdh5.8.0.jar.

fnothaft:
Also the way I found the missing class was using the command:

```
find /opt/cloudera/parcels/CDH-5.8.0-1.cdh5.8.0.p0.42/lib/hbase/lib/ -name "*.jar" -print -exec jar Ptvf {} \; | grep -e /opt/ -e <classname>
```

This command is terrible, but if it finds the class in one of the jarfiles, the path to the jarfile will immediately precede the output of the jar command. Alas.

jpdna:
I get closer with /hbase-protocol-1.2.0-cdh5.8.0.jar
and also adding  htrace-core-3.2.0-incubating.jar 
as 
```
 ../adam/bin/adam-shell --master yarn-client --num-executors 1 --executor-cores 2 --executor-memory 20g --jars /opt/cloudera/parcels/CDH/jars/hbase-hadoop2-compat-1.2.0-cdh5.8.0.jar,/opt/cloudera/parcels/CDH/jars/hbase-server-1.2.0-cdh5.8.0.jar,/opt/cloudera/parcels/CDH/jars/hbase-it-1.2.0-cdh5.8.0.jar,/opt/cloudera/parcels/CDH/jars/hbase-client-1.2.0-cdh5.8.0.jar,/opt/cloudera/parcels/CDH/jars/hbase-spark-1.2.0-cdh5.8.0.jar,/home/eecs/jpaschall/work/hbase/v12/adam/adam-assembly/target/adam_2.10-0.20.1-SNAPSHOT.jar,/opt/cloudera/parcels/CDH/jars/hbase-common-1.2.0-cdh5.8.0.jar,/opt/cloudera/parcels/CDH/jars/hbase-protocol-1.2.0-cdh5.8.0.jar,/opt/cloudera/parcels/CDH/jars/htrace-core-3.2.0-incubating.jar
```

however, now when running the create table, and presumably other hbase functions, spark seems to not be able to connect to hbase server, and times out.   However, it works properly when the hbase library dependencies are built with the project as before, just doesn't work with scope provided.

as discussed, I won't block on this as we have the WAR of not using `scope provided` for now.

adam-core/pom.xml
jpdna: Not sure, but commenting it out of pom causes build to fail.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
jpdna: Genotypes come out of HBase already "row per variant" shape like a VCF file or a VariantContext.   Thus, it is very efficient/fast to get a VariantContext directly from HBase, and useful for per variant operations like calculating allele frequencies on the fly over a set of samples.  This potential utility of directly using "variant-major" shaped data from HBase motivates me to leave in a `loadVariantContextFromHBase`

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
jpdna: Its the genotypes data read in from HBase.
The re partitioning here will have the effect of doing a shuffle prior to deserializing the raw bytes that come back as values from HBase.  Motivation to do so is to increase parallelism at the earliest point possible when, as seems to be the case, bottleneck is not reading data from HBase, but in deserializing it and making it into an ADAM RDD.

jpdna:
superseded by #1335

PR #1144, opened 2016-09-01
Merge VariantAnnotation and DatabaseVariantAnnotation records

heuermh:
Work in progress.  Depends on https://github.com/bigdatagenomics/bdg-formats/pull/105.

It appears our test coverage for `DatabaseVariantAnnotation` and related was insufficient, because this pull request passes all unit tests without any changes to `VariantAnnotationConverter`.

Still left to do:
- [ ] Update `VariantAnnotationConverter` for new schema fields
- [ ] Merge code from `VariantAnnotations` into `VariantAnnotationConverter`
- [ ] Code cleanup/refactoring in `VariantContextConverter`, `VariantAnnotationConverter`
- [ ] Write new unit tests where coverage was lacking

Issue #1216, opened 2016-10-17
Maintaining sorted/partitioned knowledge


devin-petersohn:
It would be beneficial if we had a way to know that a given file was sorted when it was written. As is, RDDs are force-sorted and partitioned before operations that require sorted RDDs. 

Implementing Issue #1175 is a part of this work.

heuermh:
See also https://github.com/bigdatagenomics/adam/issues/188

devin-petersohn:
@fnothaft Following up from our conversation this morning, the mix-in implementation is finally debugged. I am going to move toward working on splitting partition and sort.

Here is the working repo for those interested: https://github.com/devin-petersohn/adam/tree/partitioner

jpdna:
@devin-petersohn - just want to make sure you have seen this thread: #651

jpdna:
It may be useful to look at:
https://issues.apache.org/jira/browse/SPARK-8836
https://issues.apache.org/jira/browse/SPARK-2213

devin-petersohn:
Resolved with #1324.

PR #1250, opened 2016-11-08
Merge VariantAnnotation and DatabaseVariantAnnotation records

heuermh:
Supercedes #1144


adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -267,7 +267,7 @@ class ADAMContext private (@transient val sc: SparkContext) extends Serializable
    * @tparam T The type of records to return
    * @return An RDD with records of the specified type
    */
-  private[rdd] def loadParquet[T](
+  def loadParquet[T](
```
heuermh: I had to make this public again for unit tests in package `o.b.a.projections`.  It also allows for loading user-defined schema (e.g. extensions to bdg-formats) from Avro-in-Parquet files.

adam-core/src/main/scala/org/bdgenomics/adam/converters/TranscriptEffectConverter.scala
```diff
@@ -110,6 +147,7 @@ object VariantAnnotations extends Serializable with Logging {
 
     val te = TranscriptEffect.newBuilder()
     setIfNotEmpty(alternateAllele, te.setAlternateAllele(_))
+    // note: annotationImpact is not mapped
```
fnothaft: I don't get this comment; can you flesh it out more?

adam-core/src/main/scala/org/bdgenomics/adam/converters/TranscriptEffectConverter.scala
```diff
@@ -110,6 +147,7 @@ object VariantAnnotations extends Serializable with Logging {
...
     if (!effects.isEmpty) te.setEffects(effects.asJava)
```
fnothaft: `effects.nonEmpty`

adam-core/src/main/scala/org/bdgenomics/adam/converters/TranscriptEffectConverter.scala
```diff
@@ -132,26 +170,98 @@ object VariantAnnotations extends Serializable with Logging {
     Seq(te.build())
```
fnothaft: Unrelated to this PR, as this line is unchanged, but whenever possible, I prefer `Iterable` to `Seq` unless you need random lookup by index.

adam-core/src/main/scala/org/bdgenomics/adam/converters/TranscriptEffectConverter.scala
```diff
@@ -132,26 +170,98 @@ object VariantAnnotations extends Serializable with Logging {
...
+   * @param vc htsjdk variant context
+   * @param stringency validation stringency, defaults to strict
+   * @return the htsjdk VCF INFO reserved key "ANN" value converted into zero or more
+   *    TranscriptEffects, matching on alternate allele, and wrapped in an option
+   */
+  def convertToTranscriptEffects(
     variant: Variant,
     vc: VariantContext,
-    stringency: ValidationStringency = ValidationStringency.STRICT): VariantAnnotation = {
+    stringency: ValidationStringency = ValidationStringency.STRICT): Option[List[TranscriptEffect]] = {
```
fnothaft: Instead of returning `Option[List[TranscriptEffect]]` I would just return `List[TranscriptEffect]`. If you would return a `None`, I would just return a `List.empty` instead.

adam-core/src/main/scala/org/bdgenomics/adam/converters/TranscriptEffectConverter.scala
```diff
@@ -132,26 +170,98 @@ object VariantAnnotations extends Serializable with Logging {
...
+
+    def parseAndFilter(attr: String): Option[List[TranscriptEffect]] = {
+      if (attr == VCFConstants.MISSING_VALUE_v4) {
+        None
+      } else {
+        val filtered = parseAnn(attr, stringency).filter(_.getAlternateAllele == variant.getAlternateAllele)
```
fnothaft: If you make the above change, then the if-else clause here just becomes:

```
if (attr == VCFConstants.MISSING_VALUE_v4) {
  List.empty
} else {
  parseAnn(attr, stringency)
    .filter(_.getAlternateAllele == variant.getAlternateAllele)
}
```

Also, I would break at the `.filter`, because that line is a bit long.

adam-core/src/main/scala/org/bdgenomics/adam/converters/TranscriptEffectConverter.scala
```diff
@@ -132,26 +170,98 @@ object VariantAnnotations extends Serializable with Logging {
...
+   * Convert the specified transcript effects into a string suitable for a VCF INFO reserved
+   * key "ANN" value.
+   *
+   * @param effects zero or more transcript effects
+   * @return the specified transcript effects converted into a string suitable for a VCF INFO
+   *    reserved key "ANN" value
+   */
+  def convertToVcfInfoAnnValue(effects: Seq[TranscriptEffect]): String = {
+    def toFraction(numerator: java.lang.Integer, denominator: java.lang.Integer): String = {
+      val numOpt = Option(numerator)
```
fnothaft: I've seen `Option` do odd things with integers before. I think you're fine here since you're using `java.lang.Integer` which can be null (scala's Int type "cannot", IIRC, which triggers the odd behavior), but I would make sure to have a unit test covering the null integer case just to be safe.

adam-core/src/main/scala/org/bdgenomics/adam/converters/TranscriptEffectConverter.scala
```diff
@@ -132,26 +170,98 @@ object VariantAnnotations extends Serializable with Logging {
...
+      val denomOpt = Option(denominator)
+
+      val sb = StringBuilder.newBuilder
```
fnothaft: I think this code would be a bit cleaner with a match:

```
(numOpt, denomOpt) match {
  case (Some(n), Some(d)) => {
    "%d/%d".format(n, d)
  }
  case (None, None) => {
    ""
  }
  case _ => {
     // validate/throw?
     if (validationStringency == ValidationStringency.STRICT) {
       throw new IllegalArgumentException("Incorrect fractional value in %s.".format(te))
     } else if (validationStringency == ValidationStringency.LENIENT) {
       log.warn("Incorrect fractional value in %s.".format(te))
     }
     ""
  }
}
```

Also, I would either make this package private/private, or move it inside of `toAnn`, which I _think_ is the only place it is used.

adam-core/src/main/scala/org/bdgenomics/adam/projections/ContigField.scala
```diff
@@ -17,18 +17,12 @@
  */
 package org.bdgenomics.adam.projections
 
-import org.bdgenomics.formats.avro.DatabaseVariantAnnotation
+import org.bdgenomics.formats.avro.Contig
```
fnothaft: OOC, why does this show up as a file move? Any thoughts? May be just github being funky.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
fnothaft: Perhaps let's make it `private[adam]`?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/variation/VariantContextRDD.scala
```diff
@@ -60,20 +60,19 @@ case class VariantContextRDD(rdd: RDD[VariantContext],
    * @param ann Annotation RDD to join against.
    * @return Returns a VariantContextRDD where annotations have been filled in.
    */
-  def joinDatabaseVariantAnnotation(ann: DatabaseVariantAnnotationRDD): VariantContextRDD = {
+  def joinVariantAnnotations(ann: VariantAnnotationRDD): VariantContextRDD = {
     replaceRdd(rdd.keyBy(_.variant)
```
fnothaft: We might want to open a ticket for this, but after #1216 this should probably be implemented using a region join instead of a Spark core `leftOuterJoin`.

adam-core/src/test/scala/org/bdgenomics/adam/converters/TranscriptEffectConverterSuite.scala
```diff
@@ -0,0 +1,245 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.bdgenomics.adam.converters
+
+import com.google.common.collect.ImmutableList
+import htsjdk.samtools.ValidationStringency
+import htsjdk.variant.vcf.VCFConstants
+import htsjdk.variant.variantcontext.VariantContext
```
fnothaft: Nit: `htsjdk.variant.vcf` after `htsjdk.variant.variantcontext`

adam-core/src/main/scala/org/bdgenomics/adam/converters/TranscriptEffectConverter.scala
heuermh: The `annotationImpact` field (and variable above) is output by SnpEff version 4.2 but is not part of the VCF ANN specification, so I did not include it in our `TranscriptEffect` schema.

adam-core/src/main/scala/org/bdgenomics/adam/converters/TranscriptEffectConverter.scala
heuermh: That would make my brain hurt less.  The thought is elsewhere it matters whether this field has been set, so checking Option seemed more correct than checking for an empty list.

adam-core/src/main/scala/org/bdgenomics/adam/converters/TranscriptEffectConverter.scala
heuermh: Will be adding try catch with validation stringency here shortly...

adam-core/src/main/scala/org/bdgenomics/adam/converters/TranscriptEffectConverter.scala
heuermh: Yes, with `Int` unit tests fail with NPE before they can be wrapped in `Option`.

adam-core/src/main/scala/org/bdgenomics/adam/converters/TranscriptEffectConverter.scala
heuermh: I thought it was already private since it is nested in `convertToVcfInfoAnnValue`?  Still have some to learn about visibility in Scala.  The tuple of options is cleaner.  (I can't believe I just said that)

fnothaft: Ah yes, you are right RE: protection; I had missed the nesting.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
heuermh: It seems like this would be useful outside of ADAM.  I haven't fully thought through the use case though: someone wants to add a new schema record `Foo`, they extend `ADAMKyroRegistrator` to register, then extend `ADAMContext` to add their new `loadFoo` method, which presumably would call `loadParquet` for `foo.adam` paths.  If `Foo` has a sequence dictionary or samples, those avro metadata methods would also be useful, and extending from `GenomicRDD` and friends needs to be possible.

fnothaft: I don't disagree, but I'd rather keep these private until someone knocks on our door asking to make them public. My philosophy here is simply that it is easier to make private interfaces public than it is to make public interfaces private. That being said, this is a weak preference: if you feel strongly about it, I'm OK with making it public, esp. since `loadParquet` has been public previously.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/variation/VariantContextRDD.scala
heuermh: Created new issue #1259

jpdna:
@heuermh - is it reasonable / useful  for me to try to build the code in this PR locally to test it out at this point?   I tried to compile it, but can't seem to find a version bdg-formats that works with it.  

I tried both 
https://github.com/heuermh/bdg-formats/tree/master as it was a 0.9.1-SNAPSHOT version and then tried changing to current bdg-formats  0.10.1-SNAPSHOT, but bdg-formats compile problems with both.

heuermh:
@jpdna As is this branch will not compile due to the filter-related changes in bdg-formats.  I've made the code changes locally but they need more unit tests.  I'll push these in a commit tomorrow morning.

adam-core/src/main/scala/org/bdgenomics/adam/models/VariantContext.scala
```diff
@@ -143,6 +143,5 @@ class VariantContext(
     val position: ReferencePosition,
     val variant: RichVariant,
     val genotypes: Iterable[Genotype],
-    val databases: Option[DatabaseVariantAnnotation] = None) {
+    val databases: Option[VariantAnnotation] = None) {
```
heuermh: nice catch! fixed

heuermh:
Pushed new commits that fixes the separate variant and genotype filters issue and updates bdg-formats to the release version 0.10.0.  I implemented the filter stuff to the best that htsjdk makes available to us; I could either continue to hack on it so that `Genotype.filtersApplied` is set correctly or punt until a later issue.

heuermh:
Fixes #194

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantContextConverter.scala
```diff
@@ -418,11 +426,22 @@ private[adam] class VariantContextConverter(dict: Option[SequenceDictionary] = N
...
           .setSampleId(g.getSampleName)
           .setAlleles(g.getAlleles.map(VariantContextConverter.convertAllele(vc, _)))
           .setPhased(g.isPhased)
 
+        // copy variant calling annotations to update filter attributes
+        // (because the htsjdk Genotype is not available when build is called upstream)
+        val copy = VariantCallingAnnotations.newBuilder(annotations)
+        // htsjdk does not provide a field filtersWereApplied for genotype as it does in VariantContext
+        // we might be able to calculate it by querying the FT FORMAT field value directly
+        copy.setFiltersApplied(true)
```
fnothaft: I think this would work:

```
g.getAnyAttribute("FT") != null
```

jpdna:
I tried to load this VCF:
https://drive.google.com/open?id=0B6jh69UgixwpTDlSemtreklDNUU
which is the ClinVar VCF with ANN column added by SNPEff

It seems to fail, but without an error message
```
val x = sc.loadVariantAnnotations("/home/paschallj/ADAM/nov11_annot/run1/test1.200.vcf")

scala> x.rdd.count
res13: Long = 0
```
This VCF does seem to load into a VariantRDD with loadVCF fine.

I suspect there is something unexpected about the format of my VCF file ANN field, but if this is current snpeff output then it could be problem for some users.

Can you point me to a test VCF with a ANN field that is working properly that I can compare to?

fnothaft:
@jpdna do you get any error/warning messages in the logs? If you have `ValidationStringency.LENIENT` set, I'd expect to see error messages there.

jpdna:
Where do I set `ValidationStringency.LENIENT"  ?
currently I don't see any error messages in adam.log when I run this in adam-shell

heuermh:
@jpdna It might be hard to follow, since things are spread over several issues, but this pull request does not yet support populating `VariantAnnotation.transcriptEffects` from VCF INFO reserved key ANN values.  See https://github.com/bigdatagenomics/adam/issues/1044#issuecomment-260015756

fnothaft:
Sure, but even then @jpdna should be getting one VariantAnnotation record per Variant, no?

heuermh:
Maybe, I don't know how well that part of the code works. Based on recent conversations on gitter, not too well?

jpdna:
> this pull request does not yet support populating VariantAnnotation.transcriptEffects

ah, thanks for clarifying @heuermh  - I'll plan to watch this PR then for the further commits and try my test  again when you ping that reading ANN field into transcriptEffects is ready.   Perhaps some rows of the  VCF I linked to above can be a useful in the test suite - both a VEP and SNPeff derived example annotated VCF would be good.

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantContextConverter.scala
heuermh: Unfortunately it does not
https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/variant/variantcontext/Genotype.java#L560

And careful, `FT` is considered a forbidden key  :)
https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/variant/variantcontext/Genotype.java#L660

adam-core/src/main/scala/org/bdgenomics/adam/converters/TranscriptEffectConverter.scala
fnothaft: This NPE with Option types still needs to be fixed.

adam-core/src/main/scala/org/bdgenomics/adam/converters/TranscriptEffectConverter.scala
```diff
@@ -132,26 +170,115 @@ object VariantAnnotations extends Serializable with Logging {
...
+   * @return the htsjdk VCF INFO reserved key "ANN" value converted into zero or more
+   *    TranscriptEffects, matching on alternate allele, and wrapped in an option
+   */
+  def convertToTranscriptEffects(
     variant: Variant,
     vc: VariantContext,
-    stringency: ValidationStringency = ValidationStringency.STRICT): VariantAnnotation = {
+    stringency: ValidationStringency = ValidationStringency.STRICT): Option[List[TranscriptEffect]] = {
+
+    def parseAndFilter(attr: String): Option[List[TranscriptEffect]] = {
```
fnothaft: This method should be private.

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantContextConverter.scala
fnothaft: Can we create an issue to track the upstream htsjdk issue?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -1172,37 +1172,33 @@ class ADAMContext private (@transient val sc: SparkContext) extends Serializable
...
-   *
    * @param filePath The path to the VCF file(s) to load annotations from.
-   * @return Returns DatabaseVariantAnnotationRDD.
+   * @return Returns VariantAnnotationRDD.
    */
-  def loadVcfAnnotations(filePath: String): DatabaseVariantAnnotationRDD = {
-    loadVcf(filePath).toDatabaseVariantAnnotationRDD
+  def loadVcfAnnotations(
+    filePath: String): VariantAnnotationRDD = {
+    loadVcf(filePath).toVariantAnnotationRDD
```
fnothaft: Just for tracking RE @jpdna's [comment about not getting any annotations from a VCF](https://github.com/bigdatagenomics/adam/pull/1250#issuecomment-260020874), this line is the culprit. Specifically, `loadVcf` right now just parses the `Genotype`s. We should make the `VariantContextConverter` parse out the annotations by default in the follow on PR.

fnothaft: No changes necessary in this PR, just an FYI.

fnothaft: Also, @jpdna it'd be great to add some unit tests that use that file and try to load a few ANN fields. That should be an acceptance test for the release. Would you be able to do that?

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantContextConverter.scala
heuermh: Created new issue #1269

adam-core/src/main/scala/org/bdgenomics/adam/converters/TranscriptEffectConverter.scala
heuermh: As far as I can see, this is fine with `java.lang.Integer` parameters.

The unit test `"convert transcript effect to VCF ANN attribute value"` in `TranscriptEffectConverterSuite` covers nulls in various places (cdnaPosition, cdnaLength, cdsLength).

adam-core/src/main/scala/org/bdgenomics/adam/converters/TranscriptEffectConverter.scala
heuermh: `error: illegal start of statement (no modifiers allowed here)
[ERROR]     private def parseAndFilter(attr: String): Option[List[TranscriptEffect]] = {`

heuermh:
Pushed commit with some additional unit tests.  Let me know if I've addressed all the review comments, and thank you for volunteering to merge this manually.

adam-core/src/main/scala/org/bdgenomics/adam/converters/TranscriptEffectConverter.scala
fnothaft: OK, SGTM!

adam-core/src/main/scala/org/bdgenomics/adam/converters/TranscriptEffectConverter.scala
fnothaft: Ah, sorry, I misread this and didn't notice that it is nested inside another function.

fnothaft:
LGTM now! I will merge this manually shortly.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
heuermh: Running SnpEff on the VCF files we are already using for unit tests ends up being not too interesting, with all intragenic variants.  It might take a little thinking to generate a more useful VCF file, say with variants right at intron/exon boundaries of a gene with a lot of splice variants, for example.

fnothaft:
Merged into upgrade-to-0.10.0 as c750830d0f4da906048d76edc7201c4845cdcc70 and c06143be29a0937a803e3aeef3fb7ec256438331. Thanks @heuermh! It is fantastic to get this refactor in. I will merge in the upgrade-to-0.10.0 branch on tests rerunning and passing.

heuermh:
Woot!  Thank you, @fnothaft!

PR #1252, opened 2016-11-08
[ADAM-1196] Add support for open ReferenceRegions.

### LINKED ISSUE #1196: Allow open-ended intervals in loadIndexedBam
Intervals' start and end has to be specified in a ReferenceRegion [before being passed to `loadIndexedBam`](https://github.com/bigdatagenomics/adam/blob/b903cfde5740186888998c2bce7e29b8ab310405/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala#L479).

It would be nice to allow intervals of the form `[N,∞)` without resorting to making callers use `Long.MaxValue` or the like.

The right model would be a sort of `ReferenceRegion`-precursor, shaped like `(String, Long, Option[Long])`, that basically only exists ephemerally in the course of loading subsets of indexed BAMs.

Guacamole has some code that does basically this [here](https://github.com/hammerlab/guacamole/blob/master/src/main/scala/org/hammerlab/guacamole/loci/set/LociParser.scala#L57-L95), allowing open-ended intervals, whole contigs, and "all" and "none" keywords, then resolving things into concrete `ReferenceRegion`s once contig-lengths are available, after parsing a BAM header.

fnothaft:
Resolves #1196.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
@@ -73,6 +73,44 @@ object ReferenceRegion {
   implicit def orderingForOptionalPositions = OptionalRegionOrdering
 
   /**
+   * Creates a reference region that starts at the beginning of a contig.
+   *
+   * @param referenceName The name of the reference contig that this region is
+   *   on.
+   * @param end The end position for this region.
+   * @return Returns a reference region that goes from the start of a contig to
+   *   a user provided end point.
+   */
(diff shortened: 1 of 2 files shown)


heuermh:
LGTM, is this really all that is necessary?

fnothaft:
I think so, but ideally @ryan-williams can comment. It looks like the code he referenced earlier in Guacamole disappeared, so I'm not 100% sure what he was thinking the interface would look like.

heuermh:
Thank you, @fnothaft!

Issue #1222, opened 2016-10-23
Instrumentation is lost in ShuffleRegionJoin


fnothaft:
Not sure why this happens, but I ran into it. TBD whether the problem is in ADAM or upstream.

PR #1253, opened 2016-11-08
Misc shuffle join fixes.

### LINKED ISSUE #1220: GenomicRDD shuffle region join passes partition count to partition size
Leads to joined RDDs having a hilarious number of partitions. See https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ShuffleRegionJoin.scala#L125 vs https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala#L362. Interestingly, this seems to work, but with bad performance. Fortunately, there is a WAR since you can override the value passed to `partitionLength` through the `optPartitions` parameter. This'll be a simple fix.

fnothaft:
Resolves #1220, #1222. Could use some more unit test coverage; I'll make a pass back and get that.


heuermh:
`<ping>`   Ready to go or wait for more unit tests?

fnothaft:
Not quite ready to go yet, will get to the tests today.

fnothaft:
Added commit c01b8a7, which adds a LOT of tests (and fixes some bugs that the tests uncovered). This is good to go now from my side. I think c01b8a7 should get squashed into 11725a8 before merging, though.

heuermh:
Could you squash these commits as you'd like?  Then I'll merge using the Rebase and Merge button.

fnothaft:
@heuermh I decided not to squash any of the commits down, but I amended the last commit message to make it more clear about what the last commit does. I've just pushed.

heuermh:
Thank you, @fnothaft!

PR #1254, opened 2016-11-08
[ADAM-1221] Bump Spark/Hadoop versions.

fnothaft:
Resolves #1221.

PR #1258, opened 2016-11-10
[ADAM-1221] Bump Spark/Hadoop versions.

fnothaft:
Reverts #1254, then reapplies commit but while fixing sed replacement line.

### DIFF
--- a/scripts/move_to_spark_2.sh
+++ b/scripts/move_to_spark_2.sh
@@ -2,7 +2,7 @@
 
 set +x
 
-svp="\${scala\.version\.prefix}"
+svp="\${scala.version.prefix}"
 substitution_cmd="s/_$svp/-spark2_$svp/g"
 
 find . -name "pom.xml" -exec sed \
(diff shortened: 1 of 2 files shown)


fnothaft:
This is still wrong!

scripts/move_to_spark_2.sh
```diff
@@ -2,10 +2,12 @@
 
 set +x
 
-svp="\${scala\.version\.prefix}"
+svp="\${scala.version.prefix}"
 substitution_cmd="s/_$svp/-spark2_$svp/g"
 
 find . -name "pom.xml" -exec sed \
+    -e "/utils-/ s/_2\.10/-spark2_2.10/g" \
+    -e "/adam-/ s/_2\.10/-spark2_2.10/g" \
```
heuermh: Does this mean we're missing `${scala.version.prefix}` somewhere, or are we intentionally not using it somewhere, in artifactId perhaps?

fnothaft: That line of thought was what led me to delete it originally. However, yeah, we're intentionally not using SVP in the artifact names. That is what #1254 broke. This looks to be passing but I'd like to go over the CI logs before we merge.

heuermh:
Thank you, @fnothaft!

PR #1260, opened 2016-11-11
[ADAM-1233] Expose header lines in Variant-related GenomicRDDs

### LINKED ISSUE #1233: Need a way for users to add VCF header lines
I've kinda known this to be a problem for a while, but I was just using a VCF I saved with ADAM that had genotype filters (`FT` field) in a downstream tool, and I got an exception in the downstream tool because my VCF did not have a `##FILTER=<ID=ID,Description="description">` line in the header. I have a general idea for how to do this; will post next week for discussion when I'm not bandwidth limited. I think this approach would also work for SAM program fields, which would be a plus.

fnothaft:
Resolves #1233. Adds a `headerLines` field to all Variant-related GenomicRDDs. In the current implementation, this field is populated by the set of all header lines for all INFO/FORMAT fields that we support. In a future patch, we will also store the header lines that correspond to fields that do not map to a specific field in the Variant/Genotype schema, and that are stored in an
attribute map.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/converters/SupportedHeaderLines.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/converters/SupportedHeaderLines.scala
@@ -67,4 +67,9 @@ private[adam] object SupportedHeaderLines {
    * All format lines in VCF format.
    */
   lazy val formatHeaderLines = VariantAnnotationConverter.FORMAT_KEYS.map(_.hdrLine)
+
+  /**
+   * All supported header lines in VCF format.
+   */
+  lazy val allHeaderLines = infoHeaderLines ++ formatHeaderLines
 }
--- a/adam-core/src/main/scala/org/bdgenomics/adam/converters/SupportedHeaderLines.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/converters/SupportedHeaderLines.scala
@@ -29,7 +29,7 @@ import htsjdk.variant.vcf.{
 private[adam] object SupportedHeaderLines {
 
   lazy val ancestralAllele = new VCFInfoHeaderLine("AA", 1, VCFHeaderLineType.String, "Ancestral allele");
-  lazy val alleleCount = new VCFInfoHeaderLine("AC", VCFHeaderLineCount.A, VCFHeaderLineType.String, "Allele count");
+  lazy val alleleCount = new VCFInfoHeaderLine("AC", VCFHeaderLineCount.A, VCFHeaderLineType.Integer, "Allele count");
   lazy val readDepth = new VCFInfoHeaderLine("AD", VCFHeaderLineCount.R, VCFHeaderLineType.Integer, "Total read depths for each allele");
   lazy val forwardReadDepth = new VCFInfoHeaderLine("ADF", VCFHeaderLineCount.R, VCFHeaderLineType.Integer, "Read depths for each allele on the forward strand");
   lazy val reverseReadDepth = new VCFInfoHeaderLine("ADR", VCFHeaderLineCount.R, VCFHeaderLineType.Integer, "Read depths for each allele on the reverse strand");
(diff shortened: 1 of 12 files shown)


jpdna:
LGTM

heuermh:
I don't see a way for users to add to the set of headers (short of modifying the `headerLines` field directly) and I don't see the code for persisting them out to disk or using them to write out the VCF header.  Future commits to this PR or wait to a later PR?

fnothaft:
> I don't see a way for users to add to the set of headers (short of modifying the headerLines field directly)

Since these are all case classes, that's a one-liner via the copy method, but I can make a pass to update this if you'd like.

> I don't see the code for <snip /> using them to write out the VCF header

That change is https://github.com/bigdatagenomics/adam/pull/1260/files#diff-b8227780b230d68bbf54dc946ef1907eR145.

> I don't see the code for persisting them out to disk

You know, I'd thought that change would require an addition and a bump to BDG formats, but I just realized it doesn't. I'll get that today.

heuermh:
>> I don't see a way for users to add to the set of headers (short of modifying the headerLines field directly)

>Since these are all case classes, that's a one-liner via the copy method, but I can make a pass to update this if you'd like.

I can check learn something new off my todo list for the day, thanks.  :)

fnothaft:
> I can check learn something new off my todo list for the day, thanks. :)

Yeah, it's a poorly documented feature of case classes. Sigh!

adam-core/src/main/scala/org/bdgenomics/adam/converters/SupportedHeaderLines.scala
```diff
@@ -41,7 +41,7 @@ private[adam] object SupportedHeaderLines {
   lazy val validated = new VCFInfoHeaderLine("VALIDATED", 0, VCFHeaderLineType.Flag, "Validated by follow-up experiment");
   lazy val thousandGenomes = new VCFInfoHeaderLine("1000G", 0, VCFHeaderLineType.Flag, "Membership in 1000 Genomes");
   lazy val somatic = new VCFInfoHeaderLine("SOMATIC", 0, VCFHeaderLineType.Flag, "Somatic event");
-  lazy val transcriptEffects = new VCFInfoHeaderLine("ANN", VCFHeaderLineCount.UNBOUNDED, VCFHeaderLineType.String, "Functional annotations: 'Allele | Annotation | Annotation_Impact | Gene_Name | Gene_ID | Feature_Type | Feature_ID | Transcript_BioType | Rank | HGVS.c | HGVS.p | cDNA.pos / cDNA.length | CDS.pos / CDS.length | AA.pos / AA.length | Distance | ERRORS / WARNINGS / INFO' ");
+  lazy val transcriptEffects = new VCFInfoHeaderLine("ANN", VCFHeaderLineCount.UNBOUNDED, VCFHeaderLineType.String, "Functional annotations: 'Allele | Annotation | Annotation_Impact | Gene_Name | Gene_ID | Feature_Type | Feature_ID | Transcript_BioType | Rank | HGVS.c | HGVS.p | cDNA.pos / cDNA.length | CDS.pos / CDS.length | AA.pos / AA.length | Distance | ERRORS / WARNINGS / INFO'");
```
heuermh: that extra space is verbatim from SnpEff v4.2 output

adam-core/src/main/scala/org/bdgenomics/adam/converters/SupportedHeaderLines.scala
```diff
@@ -29,7 +29,7 @@ import htsjdk.variant.vcf.{
 private[adam] object SupportedHeaderLines {
 
   lazy val ancestralAllele = new VCFInfoHeaderLine("AA", 1, VCFHeaderLineType.String, "Ancestral allele");
-  lazy val alleleCount = new VCFInfoHeaderLine("AC", VCFHeaderLineCount.A, VCFHeaderLineType.String, "Allele count");
+  lazy val alleleCount = new VCFInfoHeaderLine("AC", VCFHeaderLineCount.A, VCFHeaderLineType.Integer, "Allele count");
```
heuermh: nice catch

adam-core/src/main/scala/org/bdgenomics/adam/converters/SupportedHeaderLines.scala
fnothaft: maybe you have better luck than i do but i spent an hour on that

fnothaft: god is dead

heuermh: You probably don't want to look, there is some header line description cleanup code in htsjdk, maybe that is at fault.  We just don't want to include both ANN with the space and ANN without the space.

fnothaft: Yeah we just check for matching ID and type, not matching description (because people can put all sorts of whatnot in the description field).

heuermh:
Thank you, @fnothaft!

Issue #1083, opened 2016-07-18
Clean up packages


fnothaft:
Documentation and packaging cleanup for:

- [ ] `org.bdgenomics.adam.models`
- [x] `org.bdgenomics.adam.projections`
- [ ] `org.bdgenomics.adam.rich` see #1263
- [x] `org.bdgenomics.adam.serialization`
- [x] `org.bdgenomics.adam.util`
- [x] `org.bdgenomics.adam.rdd` classes not cleaned up in #1051

heuermh:
I'd like to propose the following package name changes:
```
org.bdgenomics.adam.rdd.features --> org.bdgenomics.adam.feature
org.bdgenomics.adam.rdd.variation --> org.bdgenomics.adam.variant
```

heuermh:
object `Timers` in package `org.bdgenomics.adam.instrumentation` could use some higher level doc, when to define one, how to use one, etc.

fnothaft:
+1 @heuermh

PR #1263, opened 2016-11-12
Clean up `org.bdgenomics.adam.rich` package.

fnothaft:
Towards #1083.

* Added scaladoc to all classes/methods that were missing docs.
* Deprecated `ReferenceSequenceContext`
* Deprecated implicit conversions to/from `RichAlignmentRecord` and eliminated dead code.
* Made `RichCigar` package private to `org.bdgenomics.adam`. It is only used in `org.bdgenomics.algorithms.consensus`. I considered moving it over, but decided to keep all "enriched" classes in `org.bdgenomics.adam.rich`. I did eliminate the `RichCigar` singleton and make `RichCigar` a case class, while moving the only non-`apply` method over to `org.bdgenomics.adam.algorithms.consensus.ConsensusGenerator` as a private method.

b4f0eeb8c108725ea04ff41b985112c1ee079845 deletes `org.bdgenomics.adam.rich.RichGenotype`, which is unused. I wasn't sure if we wanted to do this, as `RichGenotype` is kinda useful. Thoughts?


adam-core/src/main/scala/org/bdgenomics/adam/rich/RichGenotype.scala
```diff
@@ -1,112 +0,0 @@
...
-object RichGenotype {
-  implicit def genotypeToRichGenotype(g: Genotype) = new RichGenotype(g)
-  implicit def richGenotypeToGenotype(g: RichGenotype) = g.genotype
-}
-
-class RichGenotype(val genotype: Genotype) {
-
-  def ploidy: Int = genotype.getAlleles.size
-
-  def getType: GenotypeType = {
```
heuermh: There is so much difficulty explaining what the different types mean I don't find `GenotypeType` very useful.

Hail defines even more genotype flags

genotype field | definition
------------ | -------------
`isHomRef` | true if this call is 0/0 
`isHet` | true if this call is heterozygous 
`isHetRef` | true if this call is 0/k with k>0 
`isHetNonRef` | true if this call is j/k with j>0 
`isHomVar` | true if this call is j/j with j>0 
`isCalledNonRef` | true if either isHet or isHomVar is true 
`isCalled` | true if the genotype is not ./. 
`isNotCalled` | true if the genotype is ./. 

and still doesn't distinguish partial no-calls.

adam-core/src/test/scala/org/bdgenomics/adam/converters/VariantContextConverterSuite.scala
```diff
@@ -80,7 +80,7 @@ class VariantContextConverterSuite extends ADAMFunSuite {
 
     assert(adamVC.genotypes.size === 0)
 
-    val variant = adamVC.variant
+    val variant = adamVC.variant.variant
```
fnothaft: Correct.

adam-core/src/main/scala/org/bdgenomics/adam/rich/RichAlignmentRecord.scala
```diff
@@ -37,72 +37,46 @@ import scala.collection.immutable.NumericRange
...
-    }
-    len
-  }
-
-  def apply(record: AlignmentRecord) = {
-    new RichAlignmentRecord(record)
-  }
 
+  @deprecated("Use explicit coversion wherever possible in new development.",
+    since = "0.21.0")
```
fnothaft: See above/#577.

adam-core/src/main/scala/org/bdgenomics/adam/rich/ReferenceSequenceContext.scala
```diff
@@ -17,11 +17,24 @@
...
+ * @param pos The position this residue is aligned at.
+ * @param referenceBase The base that is in the reference at this position
+ * @param cigarElement The CIGAR element covering this residue in the
+ *   read-to-reference alignment.
+ * @param cigarElementOffset The position of this base within the CIGAR
+ *   element.
  */
-
-case class ReferenceSequenceContext(pos: Option[ReferencePosition], referenceBase: Option[Char], cigarElement: CigarElement, cigarElementOffset: Int)
+@deprecated("don't use ReferenceSequenceContext in new development",
```
heuermh: Thanks, I'm good since there is a ticket to track

adam-core/src/main/scala/org/bdgenomics/adam/models/VariantContext.scala
```diff
@@ -92,7 +92,7 @@ object VariantContext {
    * @return VariantContext corresponding to the data above.
    */
   def apply(kv: (ReferencePosition, Variant, Iterable[Genotype], Option[DatabaseVariantAnnotation])): VariantContext = {
-    new VariantContext(kv._1, kv._2, kv._3, kv._4)
+    new VariantContext(kv._1, RichVariant(kv._2), kv._3, kv._4)
```
fnothaft: I think we could drop the rich variant, but I'm not entirely sure.

adam-core/src/main/scala/org/bdgenomics/adam/rich/RichGenotype.scala
heuermh: As is, yeah.  If these flags (or even extended ones as in Hail) turn out to be useful, then perhaps they could be added to the schema as flags instead of an enum.

fnothaft: SGTM

fnothaft:
Rebased. @heuermh can you review and see if there are any changes needed. As far as I can tell, we are good to go.

fnothaft:
Rebased and fixed conflicts and compile issue.

heuermh:
Thank you, @fnothaft!

PR #1264, opened 2016-11-13
Finishing up the cleanup on org.bdgenomics.adam.rdd.

fnothaft:
* Made `MDTagging` private to `read`. Exposed it through `AlignmentRecordRDD`.
  Cleaned up where it was used in `Transform`.
* Made `FlagStat` and most related models private to `read`.
* Made all `read.recalibration` and `read.realignment` classes private to
  package. Depending on the class, this varied from:
  * `adam` for classes that need to be registered with `kryo`
  * `read` for the core `RealignIndels` and `BaseQualityRecalibrator` engines
  * Their subpackage where possible.
* Made class values for genomic partitioners as private as possible.
* Added documentation to all `InFormatter`s, `InFormatterCompanion`s, and `OutFormatter`s. Made the `InFormatter`s package private with private constructors.
* Added method/class level scaladoc where missing.
* Moved `org.bdgenomics.adam.cli.FlagStatSuite` to `org.bdgenomics.adam.read`, along with the `NA12878.sam` test resource, which was otherwise unused in the `adam-cli` submodule.


heuermh:
I may have asked elsewhere, it seems like the `InFormatter`s and `OutFormatter`s need to be publicly accessible, i.e.

https://github.com/heuermh/adam-snpeff/blob/master/src/main/scala/com/github/heuermh/adam/snpeff/AdamSnpEff.scala#L52

It may be that I'm not using them correctly here, in that they aren't really implicit if I have to explicitly instantiate them.

fnothaft:
The `InFormatterCompanion` and `OutFormatter` need to be public. The `InFormatter` has to be accessible from `rdd`, but it's constructor only has to be accessible from the companion.

heuermh:
Thanks for the clarification.  I'll take a closer look at the formatter docs and the rest of this pull request tomorrow morning.

heuermh:
Thank you, @fnothaft!

PR #1266, opened 2016-11-14
make py file py3-forward-compatible

ryan-williams:
`FragmentRDDSuite` fails on my laptop, where the first `python` on the path is apparently `python3`


fnothaft:
Good catch! Thanks @ryan-williams! Merged.

PR #1135, opened 2016-08-26
Upgrade to bdg-formats-0.10.0

fnothaft:
For tracking.

PR #1268, opened 2016-11-15
use testFile in some tests

ryan-williams:


adam-cli/src/test/scala/org/bdgenomics/adam/cli/ADAM2FastaSuite.scala
```diff
@@ -17,18 +17,16 @@
  */
 package org.bdgenomics.adam.cli
 
-import com.google.common.io.Files
 import java.io.File
-import org.apache.spark.rdd.RDD
-import org.bdgenomics.adam.rdd.ADAMContext._
+
```
fnothaft: Can you clean up the newline here?

adam-cli/src/test/scala/org/bdgenomics/adam/cli/ADAM2FastaSuite.scala
```diff
@@ -17,18 +17,16 @@
...
+import com.google.common.io.Files
```
fnothaft: Are we still importing `Files` here?

ryan-williams: not sure I understand the q; `Files` is still used in this file.

adam-cli/src/test/scala/org/bdgenomics/adam/cli/ADAM2FastaSuite.scala
ryan-williams: mm my IntelliJ's ⌘⇧O defaults to a blank line between `java` imports, others, and `scala` imports:



I feel like that's a style I've seen around out in the world. I guess ADAM is not set up that way, but it also seems like people are not using IntelliJ's (or anything else's) import-optimization helpers in the codebase, as I've seen multiple unused imports laying around.

Also I'm not sure if I can change that IntelliJ setting per-project :( but lmk if given all this you still want to just have "one import blob" be the rule.

heuermh: I admit to opening up IntelliJ only rarely, and am likely the culprit for some of those unused imports.  If you have settings that work well with our style, maybe we could add them to the contributing doc.

adam-cli/src/test/scala/org/bdgenomics/adam/cli/ADAM2FastaSuite.scala
fnothaft: Yeah, I just wanted to confirm that.

adam-cli/src/test/scala/org/bdgenomics/adam/cli/ADAM2FastaSuite.scala
fnothaft: > Also I'm not sure if I can change that IntelliJ setting per-project :( but lmk if given all this you still want to just have "one import blob" be the rule.

That would be my preference for consistency with the remainder of the codebase. If it's a big hassle from your side, I can fix on merge.

ryan-williams: @fnothaft I changed it

Intersting @heuermh, what do you edit in?

fnothaft:
I can't speak for @heuermh but I never use Intellij either and do all my editing in emacs. I tried switching to Intellij several years back but it's emacs compat mode is pretty rough. I code in a variety of other languages, so cutting over and switching to Intellij doesn't make sense as an investment for me.

fnothaft:
Will merge on tests rerunning and passing.

heuermh:
Emacs and whatever scala mode comes in via melpa for me.  I try to look at all my prs in IntelliJ for the static code checking at least once, but the way it handles git makes me nervous, so I don't use it full time.  And I prefer running in terminals.  :)

fnothaft:
Jenkins, retest this please.

ryan-williams:
why retest OOC?

fnothaft:
@ryan-williams I just merged #1135 which was +~2500 -~600 so I am somewhat distrustful of old Jenkins results is all.

fnothaft:
Merged! Thanks @ryan-williams!

PR #1275, opened 2016-11-16
Expose FeatureRDD to public

Georgehe4:


heuermh:
Thank you, @Georgehe4!

PR #1277, opened 2016-11-17
use ParsedLoci in loadIndexedBam

ryan-williams:
I recently factored a small library for parsing string-representations of (possibly open-ended, e.g. `chr1`, `chr1:100-`) loci ranges out of guacamole and over to https://github.com/hammerlab/genomic-loci, because I was preparing to use it across a few of my projects, and in particular wanted it plus adam-core's `loadIndexedBam` functionality, which this PR adds.

@fnothaft you mentioned there is some similar loci-parsing code laying around somewhere in BDG, lmk if I should find/consider it instead / in addition to this.

PR #1278, opened 2016-11-17
some style nits

ryan-williams:
basically just cleared up the intellij warnings in two files I was working in for #1277:
- unused imports
- many no-op `else records` branches of RDD-instrumenting-`if`-statements 
- some `.size`->`.length` changes
- some unnecessary `.toSeq`s

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
@@ -18,9 +18,9 @@
 package org.bdgenomics.adam.rdd
 
 import java.io.{ File, FileNotFoundException, InputStream }
-import java.util.regex.Pattern
+
+import htsjdk.samtools.util.Locatable
 import htsjdk.samtools.{ SAMFileHeader, ValidationStringency }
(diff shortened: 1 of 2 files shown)


adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -480,7 +472,10 @@ class ADAMContext private (@transient val sc: SparkContext) extends Serializable
       sc.newAPIHadoopFile(filePath, classOf[AnySAMInputFormat], classOf[LongWritable],
         classOf[SAMRecordWritable], ContextUtil.getConfiguration(job))
     }
-    if (Metrics.isRecording) records.instrument() else records
+
+    if (Metrics.isRecording)
```
heuermh: Though I'll probably get downvoted, I prefer to always have curly braces
```scala
if (Metrics.isRecording) {
  records.instrument()
}
```

fnothaft: I'm also an "always braces" type.

heuermh:
Would it be ok to close this without merging?

fnothaft:
@heueh +1

PR #1281, opened 2016-11-18
Add VCF headers for fields in Variant and VariantAnnotation records

heuermh:


fnothaft:
Are we going to add the `FORMAT` keys as well, is that going to be a later PR, or are the htsjdk ones OK there?

heuermh:
Standard `FORMAT` keys will come later, once we nail down the Genotype and GenotypeAnnotation schema.  Although, if you want to whack the useless ones that are there currently . . . :)

fnothaft:
> Standard FORMAT keys will come later, once we nail down the Genotype and GenotypeAnnotation schema. Although, if you want to whack the useless ones that are there currently . . . :)

I'll pick these up. PR coming shortly.

Issue #1244, opened 2016-11-06
Does the nested record Flattener drop Maps/Arrays?


fnothaft:
I saw this when doing a cleanup pass over `org.bdgenomics.adam.util`. Specifically, look at the following lines:

* https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/util/Flattener.scala#L62-L65
* https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/util/Flattener.scala#L129-132

This is somewhat problematic as some of our nested schemas (specifically, Genotype) use Array fields, and many use Maps.

CC @tomwhite. Do you know if this code is still in use? If not, it might make more sense to drop the code instead of patching the code. I think we were using it in Eggo to flatten schemas for Hive/Impala/Spark SQL before those tools supported nested schemas, but I think those tools all support nested schemas now.

fnothaft:
Ping @tomwhite @laserson for comment.

brianpburns:
Any update on this? It looks like you made a commit removing some of the code, but the utility is still present at the path you reference in the first message for the issue.

PR #1288, opened 2016-11-23
Clean rewrite of VariantContextConverter

### LINKED ISSUE #1107: Improve test coverage of VariantContextConverter
We were talking through some edge cases in `VariantContextConverter` and weren't sure if the implementation is correct. We should grab lines out of a gVCF to run the tests.

fnothaft:
Resolves #1107:

* Clean rewrite of VariantContextConverter. Now, there's a conversion function per field, pretty much, and the conversion process is pretty much pure functional, which is kinda cool!
* Adds support for arbitrary attributes as defined in the header lines.
* Supports all of the fields in the schema except for those slated to be dropped in https://github.com/bigdatagenomics/bdg-formats/pull/108.
* Added 58 tests! w00t!

Also see https://github.com/samtools/htsjdk/issues/751.


fnothaft:
I need to make a pass and clean up the docs on this, otherwise it is all good to go.

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantContextConverter.scala
```diff
@@ -412,148 +336,844 @@ private[adam] class VariantContextConverter(dict: Option[SequenceDictionary] = N
...
+  private val annotationFormatFieldConversionFns: Iterable[(HtsjdkGenotype, VariantCallingAnnotations.Builder, Int, Array[Int]) => VariantCallingAnnotations.Builder] = Iterable(
+    formatFilters(_, _, _, _),
+    formatFisherStrandBias(_, _, _, _),
+    formatRmsMapQ(_, _, _, _),
+    formatMapQ0(_, _, _, _)
+  )
+
+  private[converters] def extractFilters(vca: VariantCallingAnnotations,
+                                         gb: GenotypeBuilder): GenotypeBuilder = {
+    Option(vca.getFiltersApplied)
```
heuermh: This is probably correct, but is much harder to read than the non-functional way it is written for Variant filters.  Sorry, I'm getting old.  :)

adam-core/src/test/scala/org/bdgenomics/adam/converters/VariantContextConverterSuite.scala
```diff
@@ -516,8 +509,635 @@ class VariantContextConverterSuite extends ADAMFunSuite {
...
+    extractor: (HtsjdkGenotype, Genotype.Builder, Int, Array[Int]) => Genotype.Builder,
+    fns: Iterable[GenotypeBuilder => GenotypeBuilder] = Iterable.empty): Genotype = {
+    val htsjdkGenotype = makeGenotype(objMap, fns)
+    extractor(htsjdkGenotype,
+      Genotype.newBuilder,
+      1,
+      Array(0, 1, 2)).build
+  }
+
+  test("no phasing set going htsjdk->adam") {
```
heuermh: these tests are nice

heuermh:
Initial review pass.

I like the conversion functions and the way they are executed inside `VariantContextConverter` but am less convinced of the API at a higher level, especially considering changes I'll need to make shortly for `Variant` and `VariantAnnotation`.

We'll also want some way to limit code duplication for similar conversions (i.e. between INFO fields on VariantAnnotation and FORMAT fields on Genotype and GenotypeAnnotation).

adam-core/src/main/scala/org/bdgenomics/adam/converters/SupportedHeaderLines.scala
```diff
@@ -63,10 +108,72 @@ private[adam] object SupportedHeaderLines {
     transcriptEffects
   )
 
+  lazy val genotype = VCFStandardHeaderLines.getFormatLine(
+    VCFConstants.GENOTYPE_KEY)
```
fnothaft: I think I'm OK trusting the GT format line, but if you feel strongly enough about it, I'll rewrite that one too. ;)

adam-core/src/main/scala/org/bdgenomics/adam/converters/SupportedHeaderLines.scala
```diff
@@ -63,10 +108,72 @@ private[adam] object SupportedHeaderLines {
...
+    "Per-sample component statistics which comprise the Fisher's Exact Test to detect strand bias.")
+  lazy val phaseSetId = new VCFFormatHeaderLine(VCFConstants.PHASE_SET_KEY,
+    1,
+    VCFHeaderLineType.Integer,
+    "Phase set ID")
+
+  // note: this is not spec compliant!
+  // however, this is due to a bug in htsjdk --> https://github.com/samtools/htsjdk/issues/751
+  // the standard header lines get populated no matter what the header lines are on the
+  // file. not clear how to disable this...
```
fnothaft: Yeah, I wasted an hour on this last night. The relevant code path is https://github.com/samtools/htsjdk/blob/master/src/main/java/htsjdk/variant/vcf/AbstractVCFCodec.java#L224. If your header line disagrees with the standard header line, HTSJDK clobbers it by default. We can disable that, but I don't think Hadoop-BAM supports disabling the header line clobbering.

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantContextConverter.scala
```diff
@@ -412,148 +336,844 @@ private[adam] class VariantContextConverter(dict: Option[SequenceDictionary] = N
...
+      val pl = g.getPL
+      gb.setNonReferenceLikelihoods(gIndices.map(idx => {
+        jFloat(PhredUtils.phredToLogProbability(pl(idx)))
+      }).toList)
+    } else {
+      gb
+    }
+  }
+
+  private def tryAndCatchStringCast[T](attr: java.lang.Object,
```
fnothaft: This was another hour of pain last night.

adam-core/src/test/scala/org/bdgenomics/adam/converters/VariantContextConverterSuite.scala
fnothaft: Thank you!

fnothaft:
> We'll also want some way to limit code duplication for similar conversions (i.e. between INFO fields on VariantAnnotation and FORMAT fields on Genotype and GenotypeAnnotation).

It would be nice if the htsjdk `Genotype` and `VariantContext` attribute access APIs weren't nonsensically different, wouldn't it?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -934,9 +934,10 @@ class ADAMContext private (@transient val sc: SparkContext) extends Serializable
     // load vcf metadata
     val (sd, samples, headers) = loadVcfMetadata(filePath)
 
-    val vcc = new VariantContextConverter(Some(sd))
+    val vcc = new VariantContextConverter()
+    val convFn = vcc.makeHtsjdkGenotypeConverter(headers)
```
fnothaft: I generally agree that the interface is a crufty. My goal here was to clean up the guts of VariantContextConverter first. That being said, we might as well do it in one pass.

> convFn seems like it should be hidden from the API to me, unless there are caller-modifiable parts to it.

I agree.

> Why not pass headers (after modifications), htsjdk context, and stringency into separate Converter classes/objects?

WRT headers, a small (but significant) optimization is to only build the conversion function once. Thus the headers should be passed to the class constructor.

Otherwise, I agree.

fnothaft: Let me try to fix up these interfaces, I'll ping back here later.

fnothaft:
@heuermh I've resolved your comments.

fnothaft:
Jenkins, retest this please.

fnothaft:
Jenkins, retest this please.

heuermh:
The failing unit test needs to be changed to no longer search for "flatten" in the `--help` text.

fnothaft:
Yeah, I'd done that and then force pushed. Jenkins is doing that frustrating thing where it doesn't pick up force pushes correctly...

fnothaft:
Jenkins, retest this please.

fnothaft:
This works end-to-end now, passes all tests, etc. I'm going to make a pass to do some method documentation cleanup, but let's get started reviewing this behemoth.

fnothaft:
Otherwise, I'm going to start cutting a bdg-formats 0.10.1 release.

fnothaft:
Jenkins, test this please.

heuermh:
> Otherwise, I'm going to start cutting a bdg-formats 0.10.1 release.

Thanks for cutting this!  Did you close the staging repo?  I don't see it up on Maven Central yet.

heuermh:
Heroic effort, my friend!

A few more things that may need to happen, we should triage to add on to this pull request, in a separate pull request for 0.21.0, or save until later release:

* `formatReadDepth` and similar shouldn't set only reference value if the array is the wrong length
* Exceptions thrown when parsing headers don't consider validation stringency (should they?)
* `convert(HtsjdkVariantContext): Seq[ADAMVariantContext]` needs try/catch with validation stringency
* Add unit tests for different `ValidationStringency`s
* Remove `val annotations: Option[VariantAnnotation]` from `VariantContext`
* Consider splitting `VariantContextConverter` into two classes, one for each direction
* Merge `Vcf2ADAM` and `ADAM2Vcf`, then split into `transformVariants` and `transformGenotypes`

fnothaft:
Here's my thoughts:

> formatReadDepth and similar shouldn't set only reference value if the array is the wrong length
> Exceptions thrown when parsing headers don't consider validation stringency (should they?)
> convert(HtsjdkVariantContext): Seq[ADAMVariantContext] needs try/catch with validation stringency
> Add unit tests for different ValidationStringencys
> Remove val annotations: Option[VariantAnnotation] from VariantContext

These are fairly small; let's do these here.

> Merge Vcf2ADAM and ADAM2Vcf, then split into transformVariants and transformGenotypes
> Consider splitting VariantContextConverter into two classes, one for each direction

Let's triage these to a later release.

fnothaft:
As an aside, I've got this done:

> Exceptions thrown when parsing headers don't consider validation stringency (should they?)

But it can't be tested, because there's only two cases where the headers throw exceptions:

* Flag in FORMAT line
* Multi-valued Flag in INFO line

and HTSJDK doesn't allow you to create the former, and if you try to create the latter, HTSJDK rewrites the header line to a single valued line, and of course, these behaviors are not overridable. I'm going to drop an ignored test inline documenting this.

fnothaft:
OK, this is good to go, short of the 0.10.1 release of bdg-formats pushing. I'm not sure why it still hasn't pushed to Maven Central. If I still don't see it in Maven Central by tomorrow morning, I'll open a JIRA with the Sonatype folks and we'll hopefully resolve it. The validation stringency tests are a bit weak; the only field that is easy to test is the SB field, so I've restricted my test to that for now.

heuermh:
>> Merge Vcf2ADAM and ADAM2Vcf, then split into transformVariants and transformGenotypes
>
> Let's triage these to a later release.

I'll take this one on later.
<br/>

> I'm going to drop an ignored test inline documenting [how much htsjdk sucks].

and

>  If I still don't see [bdg-formats] in Maven Central by tomorrow morning, I'll open a JIRA with the Sonatype folks and we'll hopefully resolve it.

Sounds good.

I see only one comment from above still unresolved (https://github.com/bigdatagenomics/adam/pull/1288#discussion_r89363706); I'll do a full review pass this evening.

heuermh:
@fnothaft @jpdna @akmorrow13 @devin-petersohn Please review for code style, correctness, unit test coverage, and API documentation.

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Flatten.scala
```diff
@@ -1,82 +0,0 @@
-/**
```
akmorrow13: Just curious why this is being removed.

fnothaft: We don't use it anymore, we had a ticket (#1244) that was leaning towards removing it, and it started causing unit test failures (due to an issue I was hinting at in said ticket), so I went ahead and just got rid of it here.

akmorrow13: sounds good!

fnothaft:
So interestingly enough, it looks like bdg-formats 0.10.1 pushed to Maven Central (http://search.maven.org/#artifactdetails%7Corg.bdgenomics.bdg-formats%7Cbdg-formats%7C0.10.1%7Cjar works), but doesn't show up if you search for the artifact. I'm guessing this is some weirdness with caching search results. Let me bump the bdg-formats dependency now.

heuermh:
Added validation stringency to Vcf2ADAM in https://github.com/fnothaft/adam/pull/13

adam-core/src/main/scala/org/bdgenomics/adam/rdd/variant/VariantContextRDD.scala
```diff
@@ -124,7 +102,8 @@ case class VariantContextRDD(rdd: RDD[VariantContext],
    *   valid VCF header. Default is false.
    */
   def saveAsVcf(filePath: String,
-                asSingleFile: Boolean = false) {
+                asSingleFile: Boolean = false,
+                stringency: ValidationStringency = ValidationStringency.LENIENT) {
```
akmorrow13: should we document stringency and its default value?

akmorrow13: Isn't it inconsistent to randomly not @param some of the parameters?

heuermh: Yeah, if you go looking you'll find lots of other missing @param tags for `stringency`, or those that don't document the default value, and the doc strings are not consistent

adam-core/src/main/scala/org/bdgenomics/adam/rdd/variant/VCFOutFormatter.scala
```diff
@@ -43,7 +47,8 @@ case class VCFOutFormatter() extends OutFormatter[VariantContext] {
   def read(is: InputStream): Iterator[VariantContext] = {
 
     // make converter and empty dicts
-    val converter = new VariantContextConverter
+    val converter = new VariantContextConverter(headerLines,
+      ValidationStringency.LENIENT)
```
devin-petersohn: why is this lenient?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/variant/VCFInFormatter.scala
```diff
@@ -56,7 +59,8 @@ private[variant] case class VCFInFormatter private (
   protected val companion = VCFInFormatter
 
   // make a converter
-  val converter = new VariantContextConverter(Some(sequences))
+  val converter = new VariantContextConverter(headerLines,
+    ValidationStringency.LENIENT)
```
devin-petersohn: why do we use lenient here?

devin-petersohn: the methods seem to default to strict, and the args default to strict, so I'm confused about the usage here.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/variant/VariantContextRDD.scala
fnothaft: I can go either way WRT fixing the validation stringency tags. I think it'd be good to fix the ones we've touched here. Perhaps I'll make a pass tomorrow (or @heuermh if he's feeling so inclined) and clean them up.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/variant/VCFInFormatter.scala
fnothaft: Ideally, in the long run we'll come up with a good way to pass validation stringency to the pipe code, probably through a Hadoop config value. In the absence of a way to override the default of strict, which will fail on any validation error (which are myriad), I prefer lenient, which will drop any erroneous variants/genotypes, but which will log the record that was dropped and why it was dropped.

heuermh:
@fnothaft Let me know if you want to chat on how we might want to squash commits and do the merge.

fnothaft:
> @fnothaft Let me know if you want to chat on how we might want to squash commits and do the merge.

Sorry I couldn't make the call today. Let me make a pass over the commits and take a gander. Is this PR good to go from your side otherwise?

fnothaft:
Here's what I propose:

* Squash commits 063acf9a9d8ab32533fe7ac0d3dba78a2fdaf694 through 7c69cd46e29b14666c48aca161c36f48c0894307 into a single commit (this is the genotype refactor)
• Squash 66f2a202cd0368a478caf071fcc73fa07f9e6e56 through 8ed8266b6a3cc3ee3aa70f7b2c3fa507d16e0625 down
• Keep 4a9f8b847cf256c921b845e525ed45357727fb77 through 0ec4f73f92a1087a1d04bde0aa4f76fad5dca45c as single commits
* Squash e7e45a86a5b9a073fd7727e966e41d456069b4e2 and 2cc625372be6b2065babec8737082b966df728ae
* Squash 9e38f176f64309025a68deb82b1431b17d8425d2 through 3546c666ffc532a491ddbb56f3889964a4135155
* Squash f747df7f2cae23c6407474c12e96e830efc5b2cc through 9d737bf447ac384de7f6b8f99dc2710dddf4520f
* Keep 49f4b9ac2ff33b72056c7e1077b7c834ac1da56b as a single commit

heuermh:
Good plan, as long as we keep the commit message `HTSJDK, gimme a break.`  :confounded:

fnothaft:
OK I will start squashing this on down.

fnothaft:
Squashed and rebased.

heuermh:
Thank you, @fnothaft @heuermh!

fnothaft:
w00t! Thanks @heuermh for all the hard work getting this in!

heuermh:
Same to you!

Issue #1286, opened 2016-11-20
ReferenceRegion on Genotype seems busted?


fnothaft:
Looking at the [code](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala#L150-L152):

```
def apply(genotype: Genotype): ReferenceRegion = {
  ReferenceRegion(genotype.getVariant)
}
```

After https://github.com/bigdatagenomics/adam/commit/525fda83754ddc7082f17c7d4cbe68d47f0efb38, we don't set some of the `Variant` fields in `Genotype`. I'm curiously concerned as to how we haven't previously caught this?

PR #1291, opened 2016-11-25
Reference region cleanup

### LINKED ISSUE #1285: Clarify strandedness in ReferenceRegion apply methods
Given the discussion by me @akmorrow13 and @jpdna in https://github.com/bigdatagenomics/utils/pull/93#issuecomment-261802968, we should probably clarify this. My proposal is:

* No `apply` methods for datatypes that can be stranded (`AlignmentRecord`s and `Feature`s?)
* Instead, two methods:
  * `def independent` which returns a reference region with an independent strand
  * `def stranded` which returns a reference region with orientation set

We'd then want to modify something in `GenomicRDD` to clarify whether the data in the RDD is stranded or not, for data that can be strand (in)dependent.

I'll take a quick hack at this tomorrow; should be a quick change. If everyone thinks it looks good, we can get it in 0.21.0, else we can push it.

fnothaft:
Resolves #1285 and #1286. The first two commits need to be squashed together.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
@@ -148,7 +148,9 @@ object ReferenceRegion {
    * @return The site where this genotype lives.
    */
   def apply(genotype: Genotype): ReferenceRegion = {
-    ReferenceRegion(genotype.getVariant)
+    ReferenceRegion(genotype.getContigName,
+      genotype.getStart,
+      genotype.getEnd)
   }
 
   /**
--- a/adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
@@ -41,7 +41,7 @@ trait ReferenceOrdering[T <: ReferenceRegion] extends Ordering[T] {
     b: T): Int = {
     val rc = regionCompare(a, b)
     if (rc == 0) {
-      a.orientation.ordinal compare b.orientation.ordinal
+      a.strand.ordinal compare b.strand.ordinal
     } else {
       rc
     }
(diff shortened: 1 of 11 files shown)


fnothaft:
Ping @heuermh @jpdna for review.

adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
```diff
@@ -220,17 +255,45 @@ object ReferenceRegion {
...
+   * @see stranded
    */
-  def apply(feature: Feature): ReferenceRegion = {
+  def unstranded(feature: Feature): ReferenceRegion = {
+    checkFeature(feature)
     new ReferenceRegion(feature.getContigName, feature.getStart, feature.getEnd)
   }
 
   /**
+   * Builds a reference region for a feature without strand information.
```
heuermh: `without strand information.` &rarr; `with strand set.`

adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
```diff
@@ -220,17 +255,45 @@ object ReferenceRegion {
...
+   *
+   * @param feature Feature to extract ReferenceRegion from.
+   * @return Extracted ReferenceRegion
+   *
+   * @see unstranded
+   */
+  def stranded(feature: Feature): ReferenceRegion = {
+    checkFeature(feature)
+    require(feature.getStrand != null,
+      "Strand is not defined in feature %s.".format(feature))
```
heuermh: Add @throws doc for this

adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
```diff
@@ -148,7 +148,9 @@ object ReferenceRegion {
    * @return The site where this genotype lives.
    */
   def apply(genotype: Genotype): ReferenceRegion = {
-    ReferenceRegion(genotype.getVariant)
+    ReferenceRegion(genotype.getContigName,
```
heuermh: A few general questions (not about this line specifically):

Rename `orientation` to `strand`?
Add strand parameters to other methods (e.g. `fromStart`, `toEnd`, `all`)?
Set strand when copying from one `ReferenceRegion` to another (i.e. line 234)?

fnothaft:
Thanks for the comments @heuermh. I've rebased and pushed two commits addressing your comments. When this looks good to you, I'll squash commits 2–4 down into a single commit for merge.

fnothaft:
Squashed down.

heuermh:
Thank you, @fnothaft

PR #1293, opened 2016-11-27
HBase as a separate repo

jpdna:
Requires: https://github.com/jpdna/bdghbase commit  809c554

This PR is a stub for any changes to ADAM to make use of HBase code as a separate repo, and discussion of that approach.

I decided to try splitting out the HBase code into a different repo, bdghbase above.
The motivation I see for this are:

1) We may have a number of different data backends including HBase -  Kudu for one.  It may be better not to put these all into the main ADAM repo.

2) I wasn't able to yet find a way to get "provided" dependency for hbase library code to work, so it will be bloating the compiled hbase module for now , better to keep that out of main ADAM, even if its working fine.

3) Testing with CI is going to be a pain for hbase, for the moment tests are going to require that an hbase instance is accessible, as I can't seem to get the mock "mini" hbase cluster to work.  Thus this complicates testing and CI, so better to keep that complexity in separate repo.

4) nice to have fast compilation time for the hbase code under development

Choice of separate repo or not seems irrelevant to how adam-shell or downstream applications using ADAM as a library would be built, I can interact with ADAM and hbase from adam-shell just as before using this PR.   

- I plan to add a "hbase" profile that will turn of the hbase dependencies in ADAM, sound good?

- I will clearly have to publish bdghbase to maven was currently do for bdg-utils and other seperate bdg repos.

- I'm not sure how to handle the CLI in this instance for hbase.  We could include the vcf2Hbase CLI code as it currently is in this PR, and perhaps it would only work in the -hbase profile was turned on.  We'd want the command and cli help to also be different based on the profile.
This doesn't seem ideal to me though - if bdghbase (or bdgkudu) is to be a separate repo, I feel like the CLI code should be in that repo as well, but I am not sure how to best integrate it with the current ADAM CLI, thoughts?

For now I am looking for general comments on this approach.  I'll ask later for review of the hbase code, as I am nearly done addressing the comments in: #1246

### DIFF
--- a/adam-cli/pom.xml
+++ b/adam-cli/pom.xml
@@ -160,6 +160,10 @@
       <scope>test</scope>
     </dependency>
     <dependency>
+      <groupId>org.bdgenomics.bdghbase</groupId>
+      <artifactId>bdghbase-core</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.scala-lang</groupId>
       <artifactId>scala-library</artifactId>
     </dependency>
--- a/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Vcf2HBase.scala
+++ b/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Vcf2HBase.scala
@@ -1,0 +1,83 @@
+/**
+ * Licensed to Big Data Genomics (BDG) under one
+ * or more contributor license agreements.  See the NOTICE file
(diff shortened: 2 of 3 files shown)


heuermh:
Extending the ADAM command line interface from an external repository is demonstrated at https://github.com/heuermh/adam-commands.  Not necessarily saying that is the right way to go for this, we might need to discuss the overall approach in person.

PR #1304, opened 2016-12-06
Add docs about building downstream applications

### LINKED ISSUE #291: Add docs about building downstream apps on top of ADAM
Add docs covering:
- Building on different hadoop versions
- Registering serializers

heuermh:
Fixes #291

### DIFF
--- a/docs/source/60_building_apps.md
+++ b/docs/source/60_building_apps.md
@@ -22,8 +22,8 @@ that assembles and delegates to the various CLI actions.
 To add a new command:
 
 Extend `Args4jBase` class to specify arguments to the command. Arguments are defined using
-the [args4j library](http://args4j.kohsuke.org/). If reading or writing to Parquet, consider
-including Parquet arguments via `with ParquetArgs`.
+the [args4j library](http://args4j.kohsuke.org/). If reading from or writing to Parquet,
+consider including Parquet arguments via `with ParquetArgs`.
 
 ```scala
 class MyCommandArgs extends Args4jBase with ParquetArgs {
@@ -48,7 +48,7 @@ object MyCommand extends BDGCommandCompanion {
 
 Extend `BDGSparkCommand` class and implement the `run(SparkContext)` method. The `MyCommandArgs`
 class defined above is provided in the constructor and specifies the generic type for `BDGSparkCommand`.
-The companion object defined above is declared as a field.  For access to the Apache Spark
+The companion object defined above is declared as a field.  For access to an
 [slf4j](http://www.slf4j.org/) Logger via the `log` field, specify `with Logging`.
 
 ```scala
@@ -109,9 +109,9 @@ Then consider making a pull request to include the new command in ADAM!
 ## Extend the ADAM CLI by adding new commands in an external repository {#external-commands}
 
 To extend the ADAM CLI by adding new commands in an external repository,
-instead of editing `ADAMMain` to add new commands, create a new main class
-that delegates to `ADAMMain` and provides additional command(s) via its
-constructor.
+instead of editing `ADAMMain` to add new commands as above, create a new
+object with a `main(args: Array[String])` method that delegates to `ADAMMain`
+and provides additional command(s) via its constructor.
 
 ```scala
 import org.bdgenomics.adam.cli.{ ADAMMain, CommandGroup }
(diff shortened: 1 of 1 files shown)


docs/source/60_building_apps.md
```diff
@@ -1,3 +1,250 @@
...
+  def apply(cmdLine: Array[String]) = {
+    new MyCommand(Args4j[MyCommandArgs](cmdLine))
+  }
+}
+```
+
+Extend `BDGSparkCommand` class and implement the `run(SparkContext)` method. The `MyCommandArgs`
+class defined above is provided in the constructor and specifies the generic type for `BDGSparkCommand`.
+The companion object defined above is declared as a field.  For access to the Apache Spark
+[slf4j](http://www.slf4j.org/) Logger via the `log` field, specify `with Logging`.
```
fnothaft: Small nit: not the Spark logger anymore, since they moved that to be package private. It's the BDG utils logger now.

docs/source/60_building_apps.md
```diff
@@ -1,3 +1,250 @@
...
+ADAM is packaged so that it can be used interatively via the ADAM shell, called from
+the command line interface (CLI), or included as a library when building downstream
+applications.
+
+This document covers three patterns for building applications downstream of ADAM:
+
+* Extend the ADAM CLI by [adding new commands](#commands)
+* Extend the ADAM CLI by [adding new commands in an external repository](#external-commands)
+* Use ADAM as a [library in new applications](#library)
+
```
fnothaft: Nit: extra whitespace.

docs/source/60_building_apps.md
```diff
@@ -1,3 +1,250 @@
...
+          countKmers : Counts the k-mers/q-mers from a read dataset.
+    countContigKmers : Counts the k-mers/q-mers from a read dataset.
+...
+
+$ ./bin/adam-submit myCommand input.foo
+```
+
+Then consider making a pull request to include the new command in ADAM!
+
+
```
fnothaft: Nit: extra whitespace.

docs/source/60_building_apps.md
```diff
@@ -1,3 +1,250 @@
...
+## Extend the ADAM CLI by adding new commands in an external repository {#external-commands}
+
+To extend the ADAM CLI by adding new commands in an external repository,
```
fnothaft: This is the pattern that works through dependency injection, right? Would it be useful to link to the dependency injection tool that we use, and how we use it? I'm just thinking that if someone made a subtle error here (not sure what, but a hypothetical error) that caused the command discovery to fail, this might be useful info for debugging. Perhaps we could add this in a subsection, or something like that.

docs/source/60_building_apps.md
```diff
@@ -1,3 +1,250 @@
...
+  adam-submit \
+  --jars my-commands.jar \
+  -- \
+  myCommand1 input.foo
+```
+
+A complete example of this pattern can be found in the
+[heuermh/adam-commands](https://github.com/heuermh/adam-examples) repository.
+
+
```
fnothaft: Nit: extra whitespace.

I'm guessing the double space before sections is intentional. No big deal if you want to keep it.

docs/source/60_building_apps.md
```diff
@@ -1,3 +1,250 @@
...
+object MyExample {
+  def main(args: Array[String]) {
+    if (args.length < 1) {
+      System.err.println("at least one argument required, e.g. input.foo")
+      System.exit(1)
+    }
+```
+
+Create an Apache Spark configuration `SparkConf` and use it to create a new `SparkContext`.
+The following serialization configuration needs to be present to register ADAM classes.
```
fnothaft: I'll open a PR against this. There's another way to do the serializer registration, e.g., if you need to add your own serializer registrations. That being said, your approach is likely the common case.

docs/source/60_building_apps.md
heuermh: It can work though dependency injection or via the ctr as demonstrated here. The Guice example is https://github.com/heuermh/adam-commands/blob/master/src/main/scala/com/github/heuermh/adam/commands/ADAMCommandsGuiceMain.scala, not so much worth mentioning in the docs.

fnothaft:
Thanks @heuermh! I've merged.

I'll PR my additions on here.

PR #1315, opened 2016-12-12
[ADAM-1308] Fix stack overflow in join with custom iterator impl.

### LINKED ISSUE #1308: Lineage for partitionAndJoin in ShuffleRegionJoin causes StackOverflow Errors
I am having issues with partitionAndJoin for LeftOuterShuffleRegionJoin. There is a stack overflow issue. I can oddly avoid this issue by putting counts in partitionAndJoin() after operations, if the inputs are cached.

fnothaft:
Resolves #1308. CC @akmorrow13 @devin-petersohn

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ShuffleRegionJoin.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ShuffleRegionJoin.scala
@@ -263,6 +263,30 @@ private[rdd] case class ManualRegionPartitioner(partitions: Int) extends Partiti
   }
 }
 
+private class AppendableIterator[T] extends Iterator[T] {
+  var iterators: ListBuffer[Iterator[T]] = ListBuffer.empty
+
+  def append(iter: Iterator[T]) {
+    if (iter.hasNext) {
+      iterators += iter
+    }
+  }
+
+  def hasNext: Boolean = {
+    iterators.nonEmpty
+  }
+
+  def next: T = {
+    val nextVal = iterators.head.next
+
+    if (!iterators.head.hasNext) {
+      iterators = iterators.tail
+    }
+
+    nextVal
+  }
+}
+
 private trait SortedIntervalPartitionJoin[T, U, RT, RU] extends Iterator[(RT, RU)] with Serializable {
   val binRegion: ReferenceRegion
   val left: BufferedIterator[((ReferenceRegion, Int), T)]
@@ -271,7 +295,7 @@ private trait SortedIntervalPartitionJoin[T, U, RT, RU] extends Iterator[(RT, RU
   private var prevLeftRegion: ReferenceRegion = _
 
   // stores the current set of joined pairs
-  protected var hits: Iterator[(RT, RU)] = Iterator.empty
+  protected var hits: AppendableIterator[(RT, RU)] = new AppendableIterator
 
   protected def advanceCache(until: Long)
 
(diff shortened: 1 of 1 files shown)


adam-core/src/main/scala/org/bdgenomics/adam/rdd/ShuffleRegionJoin.scala
```diff
@@ -263,6 +263,30 @@ private[rdd] case class ManualRegionPartitioner(partitions: Int) extends Partiti
   }
 }
 
+private class AppendableIterator[T] extends Iterator[T] {
+  var iterators: ListBuffer[Iterator[T]] = ListBuffer.empty
```
heuermh: Do you need to worry about thread safety?  Would immutable copy-on-write help here?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ShuffleRegionJoin.scala
```diff
@@ -263,6 +263,30 @@ private[rdd] case class ManualRegionPartitioner(partitions: Int) extends Partiti
...
+
+  def append(iter: Iterator[T]) {
+    if (iter.hasNext) {
+      iterators += iter
```
heuermh: safe copy?  `iter` could be modified by the caller

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ShuffleRegionJoin.scala
```diff
@@ -263,6 +263,30 @@ private[rdd] case class ManualRegionPartitioner(partitions: Int) extends Partiti
...
+    }
+  }
+
+  def hasNext: Boolean = {
+    iterators.nonEmpty
```
heuermh: shouldn't this query all of the iterators in `iterators`?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ShuffleRegionJoin.scala
fnothaft: Nah, this is only accessed inside of a single thread.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ShuffleRegionJoin.scala
fnothaft: In general, Scala has a pretty strict "advisory" (not quite a contract?) on using [Iterators](http://www.scala-lang.org/api/2.10.3/index.html#scala.collection.Iterator):

> It is of particular importance to note that, unless stated otherwise, one should never use an iterator after calling a method on it.

This is a private class used only in this file, so I think it is OK to make the assumption that `iter` is not modified after being passed.

akmorrow13:
I was going to try running this on the problem before it's merged unless anyone deems that unnecessary.

fnothaft:
> I was going to try running this on the problem before it's merged unless anyone deems that unnecessary.

I would really like that so we can confirm that this is a fix before we merge.

devin-petersohn:
I computed on the same code as before, and did not run into stack overflow errors. I think we can go forward with the merge.

akmorrow13:
I reran my pipeline and it looks good. Thanks @fnothaft !

heuermh:
Thank you for confirming, @devin-petersohn @akmorrow13

heuermh:
Thank you, @fnothaft

PR #1320, opened 2016-12-19
Supports access to indexed fa and fasta files

akmorrow13:


adam-core/src/main/scala/org/bdgenomics/adam/util/IndexedFastaFile.scala
```diff
@@ -0,0 +1,82 @@
...
+  val ref: IndexedFastaSequenceFile = {
+    val uriIdx: URI = new URI(pathWithScheme + ".fai")
+    val pathIdx = Paths.get(uriIdx)
+
+    val idx = new FastaSequenceIndex(pathIdx.toFile)
+    new IndexedFastaSequenceFile(file, idx)
+  }
+
+  // Get sequence dictionary. If sequence dictionary is not defined,
+  // generate sequence dictionary from dile
```
akmorrow13: spelling

adam-core/pom.xml
```diff
@@ -185,5 +185,10 @@
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
     </dependency>
+    <dependency>
```
fnothaft: @heuermh and I would like to delete this dependency

adam-core/src/main/scala/org/bdgenomics/adam/util/IndexedFastaFile.scala
```diff
@@ -0,0 +1,82 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.bdgenomics.adam.util
+
+import java.net.URI
```
fnothaft: Nit: Move `java` after `htsjdk`

adam-core/src/main/scala/org/bdgenomics/adam/util/IndexedFastaFile.scala
```diff
@@ -0,0 +1,82 @@
...
+import java.nio.file.Paths
+
```
fnothaft: Nit: no space

adam-core/src/main/scala/org/bdgenomics/adam/util/IndexedFastaFile.scala
```diff
@@ -0,0 +1,82 @@
...
+/**
+ * Loads and extracts sequences directly from indexed fasta or fa files. filePath requires fai index in the
+ * same directory with same naming convention.
+ *
+ * @param filePath path to fasta or fa index
+ */
+case class IndexedFastaFile(sc: SparkContext, filePath: String) extends ReferenceFile with Logging {
+
+  // get absolute path and scheme to create URI
+  val path = new Path(filePath).toString
```
fnothaft: Let's make all these `val`s private.

adam-core/src/main/scala/org/bdgenomics/adam/util/IndexedFastaFile.scala
```diff
@@ -0,0 +1,82 @@
...
+  val scheme: String = new Path(path).getFileSystem(sc.hadoopConfiguration).getScheme
```
fnothaft: Nit: Do you need the `: String` type definition here?

adam-core/src/main/scala/org/bdgenomics/adam/util/IndexedFastaFile.scala
```diff
@@ -0,0 +1,82 @@
...
+  val pathWithScheme = s"${scheme}://${path}"
+
+  val uri: URI = new URI(pathWithScheme)
```
fnothaft: Nit: Ditto here RE `: URI`

adam-core/src/main/scala/org/bdgenomics/adam/util/IndexedFastaFile.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  val file = Paths.get(uri).toFile
```
fnothaft: We require the dependency on https://github.com/damiencarol/jsr203-hadoop, for this, right? If so, let's document that here. Also, I believe we pull that in right now as a transient dependency through Hadoop-BAM. Let's make that an explicit dependency in the POM.

adam-core/src/main/scala/org/bdgenomics/adam/util/IndexedFastaFile.scala
```diff
@@ -0,0 +1,82 @@
...
+    new IndexedFastaSequenceFile(file, idx)
+  }
+
+  // Get sequence dictionary. If sequence dictionary is not defined,
+  // generate sequence dictionary from dile
+  val sequences =
+    try {
+      SequenceDictionary(ref.getSequenceDictionary)
+    } catch {
+      case e: Exception => {
```
fnothaft: I'd rather not generate the Sequence Dictionary file, my preference would be to fall back on a validation stringency check here. I'm thinking it'd be best to:

```
} catch {
  case e: Exception => {
    if (stringency == ValidationStringency.STRICT) {
      throw e
    } else {
      if (stringency == ValidationStringency.LENIENT) {
        log.warn("Caught exception %s when loading FASTA sequence dictionary. Using empty dictionary instead.".format(e))
      }
      SequenceDictionary.empty
    }
  }
}
```

adam-core/src/main/scala/org/bdgenomics/adam/util/IndexedFastaFile.scala
```diff
@@ -0,0 +1,82 @@
...
+
+  /**
+   * Extracts base sequence from FastaSequenceIndex
+   * @param region The desired ReferenceRegion to extract.
+   * @return The reference sequence at the desired locus.
+   */
+  def extract(region: ReferenceRegion): String = {
+    // Note: index is one based
+    val start = Math.max(1L, region.start)
+    val end = Math.min(sequences(region.referenceName).get.length, region.end)
```
fnothaft: I'd rather not do the `Math.min` here, at least not by default. Do we get a validation error if `region.end` is longer than the sequence length? If so, let's handle that with validation stringency --> perhaps we do the Math.min in the SILENT/LENIENT cases?

pom.xml
```diff
@@ -564,6 +565,17 @@
         <artifactId>scala-guice_${scala.version.prefix}</artifactId>
         <version>4.0.1</version>
       </dependency>
+      <dependency>
```
fnothaft: Would like to remove.

adam-core/src/main/scala/org/bdgenomics/adam/util/IndexedFastaFile.scala
heuermh: Explicit dependency on something for runtime scope?  Might not be worth the headache of keeping our dependency version number in sync with the one provided by Hadoop-BAM.

heuermh:
I didn't notice during our conversation this afternoon that this is for `ReferenceFile`, not something that we might want to use to populate an `RDD` of `Sequence`s or `Slice`s.

Thus I'm generally +1 after @fnothaft's requested changes are in.

adam-core/src/main/scala/org/bdgenomics/adam/util/IndexedFastaFile.scala
akmorrow13: For Mango, we require a sequence dictionary. So in this case, for any fa or fasta file, we will not have a sequence dictionary and thus Mango will fail. If this is a problem and we can't find another way to get the sequence dictionary, I can move this code to Mango.

fnothaft: In that case, you'd pass `stringency = ValidationStringency.STRICT` from Mango. There's several reasons (in no specific order) I'd like to fail(/validate) instead of regenerate the `.dict` file:

1. Consistency with other tools. All of the Picard CLI's (and the GATK) will fail by default if the `.dict` file is missing. They won't regenerate the file.
2. Simplicity of handling edge cases. If loading the sequence dictionary fails, this could've happened because the file is missing (and then regenerating the file is a reasonable answer) but it could also be a file permission issue, corrupt data on file system, bad Hadoop configuration causes JSR-203-hadoop to barf, etc. In all of those cases, regenerating the sequence dictionary is not the correct answer.
3. Packaging. I don't want to add Picard as a dependency.

akmorrow13: Ok, that makes sense, Mango can require a .dict file.

fnothaft: Yeah, I would just pass `stringency = ValidationStringency.STRICT` and try/catch for the error.

adam-core/pom.xml
```diff
@@ -185,5 +185,10 @@
...
+      <groupId>com.github.broadinstitute</groupId>
+      <artifactId>picard</artifactId>
+      <scope>compile</scope>
```
ryan-williams: small drive-by comment but any reason to explicitly declare `compile` scope?

adam-core/src/main/scala/org/bdgenomics/adam/util/IndexedFastaFile.scala
ryan-williams: I often leave it off in these situations, but FWIW IntelliJ has an inspection that warns against not explicitly denoting types of public members, for one reason or another.

adam-core/src/main/scala/org/bdgenomics/adam/util/IndexedFastaFile.scala
akmorrow13: Where does the stringency get passed? Do we let this be a parameter in the case class?

fnothaft: Choice is yours, but yes, sounds reasonable.

adam-core/src/main/scala/org/bdgenomics/adam/util/IndexedFastaFile.scala
akmorrow13: yes, there is an error if region end is longer than the sequence.

akmorrow13: However, if we are going to allow SequenceDictionary.empty mentioned in the above request, we can't call Math.min for getting the sequence length because sequences will be null.

fnothaft: I would rather us not `Math.max`/`.min` here, and throw the error instead, so I see that as a feature and not a bug.

akmorrow13: What I don't like about that is that twobit and NucleotideContig extract() functions don't throw errors when you pass the contig end, so this will be the only ReferenceFile implementation that throws an error for that.

fnothaft: I don't think we define what the function returns if the provided query reference region is invalid.

adam-core/src/main/scala/org/bdgenomics/adam/util/IndexedFastaFile.scala
fnothaft: Also, small aside: I think folks in the Scala ecosystem prefer to catch `Throwable` as opposed to `Exception`.

pom.xml
```diff
@@ -30,6 +30,7 @@
     <bdg-formats.version>0.10.0</bdg-formats.version>
     <bdg-utils.version>0.2.10</bdg-utils.version>
     <htsjdk.version>2.5.0</htsjdk.version>
+    <picard.version>2.8.0</picard.version>
```
fnothaft: Left this one in.

fnothaft:
@akmorrow13 can you run `./scripts/format-source`?

fnothaft:
Merged! Thanks @akmorrow13!

PR #1326, opened 2016-12-25
unroll/optimize some JavaConversions

ryan-williams:
[A `.contains` instead of a `.containsKey`](https://github.com/bigdatagenomics/adam/compare/master...ryan-williams:jc?expand=1#diff-0ed34ae1b42fa7d82a9c42d575490d99R147) is the main thing here; the latter does an O(n) conversion only to use a worse Java method (which takes `Object`, losing type-checking).

The others seem like they should save an O(n) on a `.map` by going through java/scala iterators instead of collections.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/models/SequenceDictionary.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/models/SequenceDictionary.scala
@@ -17,11 +17,12 @@
  */
 package org.bdgenomics.adam.models
 
-import htsjdk.samtools.{ SAMFileHeader, SAMSequenceRecord, SAMSequenceDictionary }
+import htsjdk.samtools.{ SAMFileHeader, SAMSequenceDictionary, SAMSequenceRecord }
 import htsjdk.variant.vcf.VCFHeader
-import org.bdgenomics.formats.avro.{ NucleotideContigFragment, Contig }
+import org.bdgenomics.formats.avro.{ Contig, NucleotideContigFragment }
+
+import scala.collection.JavaConversions.{ asScalaIterator, seqAsJavaList }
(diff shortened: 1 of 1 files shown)


ryan-williams:
wrt `containsKey`, I just noticed I was getting failures in a branch on my fork where I swapped in a value-class `ContigName` for `String` in a few places, including in the `SequenceDictionary` class, and `SequenceDictionary.containsRefName` started returning false-negatives, because it was looking for `String`s in a map whose keys were `ContigName`s; in a typed world / normal Scala an appropriate implicit would be applied to make sure you're not looking up a key in a map with the wrong type.

[`java.util.Map.containsKey` takes an `Object`](https://docs.oracle.com/javase/7/docs/api/java/util/Map.html#containsKey(java.lang.Object)).

Worse still, in order for `.containsKey` to even be callable on the `byName` `Map`, an implicit `JavaConversions` member was being applied, so it was doubly suboptimal.

While fixing that, I commented out the `JavaConversions` import to audit the other implicit uses of it in the file, and thought I may as well change the other instances for the reason I described previously: better to `.map` and convert between java⟺scala on iterators and then materialize as a  java x

fnothaft:
@ryan-williams oh, that makes tons of sense. When this first came in, I just thinking to myself that I wasn't familiar with the `containsKey` method on any of the Scala collection classes... I agree with your point WRT converting between Scala and Java using iterators. That should probably be our best practice as long as we're not going to Avro-land (where java.lang.List rules the day) or to certain areas in htsjdk (where nothing is consistent).

fnothaft:
Thanks @ryan-williams! I've manually merged as https://github.com/bigdatagenomics/adam/commit/252276028b0a0a23191b331e64d309354b7e12f8.

PR #1319, opened 2016-12-18
Add interlocks for move_to_xyz scripts.

fnothaft:
Resolves #1317. Depends on #1318.

CC @shaneknapp this should prevent the problem you saw this week from happening again.

PR #1329, opened 2016-12-27
[ADAM-1307] move_to_spark_2 fails after moving to scala 2.11.

### LINKED ISSUE #1307: Add move_to_spark_1.sh script and update README to mention
For building from source, it is useful to go back and forth between building against Spark 1.x/2.x.  We have a `move_to_spark_2.sh` script that is called as part of the release process but no corresponding `move_to_spark_1.sh` to reverse those changes.

Then as mentioned in #1306 the building from source section of README should mention these scripts.

fnothaft:
The sed command that moved to Spark 2 looked for _2.10 in the ADAM pom artifact names. This commit fixes that issue by looking for _2.1 instead, which matches both Scala 2.10 and 2.11, as well as the eventual Scala 2.12. Resolves #1307.

### DIFF
--- a/scripts/move_to_spark_2.sh
+++ b/scripts/move_to_spark_2.sh
@@ -6,8 +6,8 @@ svp="\${scala.version.prefix}"
 substitution_cmd="s/_$svp/-spark2_$svp/g"
 
 find . -name "pom.xml" -exec sed \
-    -e "/utils-/ s/_2\.10/-spark2_2.10/g" \
-    -e "/adam-/ s/_2\.10/-spark2_2.10/g" \
+    -e "/utils-/ s/_2\.1/-spark2_2.1/g" \
+    -e "/adam-/ s/_2\.1/-spark2_2.1/g" \
     -e "/utils-/ $substitution_cmd" \
     -e "/adam-/ $substitution_cmd" \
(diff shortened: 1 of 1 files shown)


heuermh:
Thanks, this works for me.

We might want to restrict the 2.10/2.11 replace patterns (later is probably fine) as we're accidentally catching this
```
$ git diff .
diff --git a/pom.xml b/pom.xml
index 73976347..a69f47a0 100644
--- a/pom.xml
+++ b/pom.xml
@@ -475,7 +475,7 @@
       </dependency>
       <dependency>
         <groupId>org.apache.parquet</groupId>
-        <!-- This library has no Scala 2.11 version, but using the 2.10 version seems to work. -->
+        <!-- This library has no Scala 2.10 version, but using the 2.10 version seems to work. -->
         <artifactId>parquet-scala_2.10</artifactId>
         <version>${parquet.version}</version>
         <exclusions>
```

and with utils currently at version 0.2.10 and json4s at version 3.2.11 we might run into other trouble.

fnothaft:
> We might want to restrict the 2.10/2.11 replace patterns (later is probably fine) as we're accidentally causing this

IIRC, I fix that in #1319

heuermh:
Thank you, @fnothaft!

PR #1335, opened 2017-01-03
HBase genotypes backend -revised

jpdna:
supersedes #1246 
and addresses comments from there.

Please make a review pass on this.

I'll add instructions soon as to how to run on cluster for reviewers to try live.


adam-cli/src/main/scala/org/bdgenomics/adam/cli/Vcf2HBase.scala
```diff
@@ -0,0 +1,84 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.bdgenomics.adam.cli
+
+import org.apache.spark.SparkContext
+
```
fnothaft: Nit: remove whitespace.

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Vcf2HBase.scala
```diff
@@ -0,0 +1,84 @@
...
+    new Vcf2HBase(Args4j[Vcf2HBaseArgs](cmdLine))
+  }
+}
+
+class Vcf2HBaseArgs extends Args4jBase {
+
+  @Argument(required = true, metaVar = "VCF", usage = "The VCF file to convert", index = 0)
+  var vcfPath: String = _
+
+  @Argument(required = true, metaVar = "-hbase_table", usage = "HBase table name in which to load VCF file")
```
fnothaft: For arguments, should be allcaps, no `-`, i.e., `HBASE_TABLE`

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Vcf2HBase.scala
```diff
@@ -0,0 +1,84 @@
...
+  var hbaseTable: String = null
+
+  @Args4jOption(required = true, name = "-seq_dict_id", usage = "User defined name to apply to the sequence dictionary create from this VCF, or name of existing sequence dictionary to be used")
```
fnothaft: Ditto: `SEQ_DICT_ID`

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Vcf2HBase.scala
```diff
@@ -0,0 +1,84 @@
...
+
+  @Args4jOption(required = false, name = "-use_existing_seqdict", usage = "Use an existing sequence dictionary, don't write a new one")
+  var useExistingSeqDict: Boolean = false
+
+  @Args4jOption(required = false, name = "-repartition", usage = "Repartition into N partitions prior to writing to HBase")
+  var repartitionNum: Int = 0
+
+  @Argument(required = true, metaVar = "-staging_folder", usage = "Location for temporary files during bulk load")
+  var stagingFolder: String = null
+
```
fnothaft: Nit: remove whitespace.

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Vcf2HBase.scala
```diff
@@ -0,0 +1,84 @@
...
+}
+
+class Vcf2HBase(protected val args: Vcf2HBaseArgs) extends BDGSparkCommand[Vcf2HBaseArgs] with Logging {
+  val companion = Transform
```
fnothaft: wrong companion object

adam-core/pom.xml
```diff
@@ -137,6 +137,10 @@
       <artifactId>spark-core_${scala.version.prefix}</artifactId>
     </dependency>
     <dependency>
+      <groupId>org.apache.spark</groupId>
+      <artifactId>spark-streaming_${scala.version.prefix}</artifactId>
```
fnothaft: Why do we need a new dependency on Spark Streaming?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+   * In standard usage user will create a single instance of this object, and pass it as the first parameter
+   * when using the HBaseFunctions public API functions
+   *
+   * @param sc SparkContext
+   * @param inConf optional HBase Configuration parameter, only used if manually configuring HBase connection
+   * @param inConnection optional HBase Connection parameter, only used if manually configuring HBase connection
+   * @param inAdmin optional HBase Admin parameter, only used if manually configuring HBase connection
+   * @param inHBaseContext optional HBaseContext parameter, only used if manually configuring HBase connection
+   */
+
```
fnothaft: Nit: extra space.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+  class HBaseSparkDAO(sc: SparkContext,
```
fnothaft: Thoughts on renaming this to something like `HBaseDataAccessObject`? When I first saw `DAO` in the code, I was confused. I don't think it's an obvious abbreviation. I agree that the long name would be a bit verbose, but I tend to prefer verbose names, since characters are cheap but brain time is expensive. ;) 

Also, a small nit: I feel like it's a bit redundant to say `Spark`, but I don't have a strong feeling here.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+                      inConf: Option[Configuration] = None,
+                      inConnection: Option[Connection] = None,
+                      inAdmin: Option[Admin] = None,
+                      inHBaseContext: Option[HBaseContext] = None) {
+
+    val conf = if (inConf.isDefined) inConf.get else HBaseConfiguration.create()
```
fnothaft: `inConf.getOrElse` would be more idiomatic scala.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+   *
+   * @param sc SparkContext
+   * @param inConf optional HBase Configuration parameter, only used if manually configuring HBase connection
+   * @param inConnection optional HBase Connection parameter, only used if manually configuring HBase connection
+   * @param inAdmin optional HBase Admin parameter, only used if manually configuring HBase connection
+   * @param inHBaseContext optional HBaseContext parameter, only used if manually configuring HBase connection
+   */
+
+  class HBaseSparkDAO(sc: SparkContext,
+                      inConf: Option[Configuration] = None,
```
fnothaft: I'd like to prefix these with `opt` to denote that they're all Options.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+                      inConnection: Option[Connection] = None,
+                      inAdmin: Option[Admin] = None,
+                      inHBaseContext: Option[HBaseContext] = None) {
+
+    val conf = if (inConf.isDefined) inConf.get else HBaseConfiguration.create()
+    val connection = if (inConnection.isDefined) inConnection.get else ConnectionFactory.createConnection(conf)
+    val admin = if (inAdmin.isDefined) inAdmin.get else connection.getAdmin
+    val hbaseContext = if (inHBaseContext.isDefined) inHBaseContext.get else new HBaseContext(sc, conf)
+
+    def createTable(tableDescriptorMeta: HTableDescriptor) { admin.createTable(tableDescriptorMeta) }
```
fnothaft: Add Scaladocs.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+    def getTable(tableName: TableName): Table = { connection.getTable(tableName) }
+
+    def getHBaseRDD(tableName: org.apache.hadoop.hbase.TableName,
```
fnothaft: I feel like this should be package private. I feel that an RDD of non-decoded results shouldn't be exposed as a public API.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+                    scans: org.apache.hadoop.hbase.client.Scan): RDD[scala.Tuple2[org.apache.hadoop.hbase.io.ImmutableBytesWritable, org.apache.hadoop.hbase.client.Result]] = {
+      hbaseContext.hbaseRDD(tableName, scans)
+
+    }
+
+    def hbaseBulkLoad(genodata: RDD[(Array[Byte], List[(String, Array[Byte])])],
+                      hbaseTableName: String,
+                      flatMap: scala.Function1[(Array[Byte], List[(String, Array[Byte])]), scala.Iterator[scala.Tuple2[org.apache.hadoop.hbase.spark.KeyFamilyQualifier, scala.Array[scala.Byte]]]],
```
fnothaft: change `scala.Function1[(Array[Byte], List[(String, Array[Byte])]), scala.Iterator[scala.Tuple2[org.apache.hadoop.hbase.spark.KeyFamilyQualifier, scala.Array[scala.Byte]]]]` to `((Array[Byte], List[(String, Array[Byte])]) => Iterator[(org.apache.hadoop.hbase.spark.KeyFamilyQualifier, Array[Byte])]`

Also, use short names, no?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+                      stagingFolder: String) = {
+
+      val familyHBaseWriterOptions = new java.util.HashMap[Array[Byte], FamilyHFileWriteOptions]
+      val f1Options = new FamilyHFileWriteOptions("GZ", "ROW", 128, "FAST_DIFF")
```
fnothaft: Any way to make these configurable? If not, let's document what we have them set to.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+      // This permission change appears necessary, plan to revisit to find better way
+      ("hadoop fs -chmod -R 660 " + stagingFolder) !
+
+      val load = new LoadIncrementalHFiles(conf)
+      load.doBulkLoad(new Path(stagingFolder), admin, connection.getTable(TableName.valueOf(hbaseTableName)),
+        connection.getRegionLocator(TableName.valueOf(hbaseTableName)))
+
+    }
+
+    def hbaseBulkDelete() {}
```
fnothaft: Remove or fill with `???` so that a NotImplementedException gets thrown when it is called.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+      genodata.hbaseBulkLoad(hbaseContext,
+        TableName.valueOf(hbaseTableName),
+        flatMap,
+        stagingFolder,
+        familyHBaseWriterOptions,
+        compactionExclude = false,
+        HConstants.DEFAULT_MAX_FILE_SIZE)
+
+      // This permission change appears necessary, plan to revisit to find better way
+      ("hadoop fs -chmod -R 660 " + stagingFolder) !
```
fnothaft: For portability, we should do this through the [Hadoop FileSystem API](hadoop.apache.org/docs/r2.4.1/api/org/apache/hadoop/fs/FileSystem.html).

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+   *
+   * @tparam T Type of rowKeyInfo class which is container for data needed to construct a row key
+   * @tparam U Type of rangePrefixInfo class which is a container for the data needed to define a key range
+   */
+  private[hbase] abstract class KeyStrategy[T, U] {
+    def getKey(rowKeyInfo: T): Array[Byte]
+    def getKeyRangePrefix(rangePrefixInfo: U): (Array[Byte], Array[Byte])
+  }
+
+  private[hbase] case class KeyStrategy1rowKeyInfo(contigName: String,
```
fnothaft: Can you explain the name here? I don't quite get it. Class looks good otherwise though.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+   *   start
+   *   refAlleel
+   *   AltAllele
+   *   length
+   *
+   */
+
+  private[hbase] object KeyStrategy1 extends KeyStrategy[KeyStrategy1rowKeyInfo, KeyStrategy1RangePrefixInfo] {
+    def getKey(rowKeyInfo: KeyStrategy1rowKeyInfo): Array[Byte] = {
+      Bytes.toBytes(rowKeyInfo.contigName + "_" +
```
fnothaft: I would replace this with:
```
"%s_%10d_%s_%s_%d".format(rowKeyInfo.contigName,
  rowKeyInfo.start,
  rowKeyInfo.refAllele,
  rowKeyInfo.altAllele,
  rowKeyInfo.end - rowKeyInfo.start)
```

Also, do we need to handle nulls wrt altAllele? For gVCF reference model genotypes, we set alt to null.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+        String.format("%10s", rowKeyInfo.start.toString).replace(' ', '0') + "_" +
+        rowKeyInfo.refAllele + "_" +
+        rowKeyInfo.altAllele + "_" +
+        (rowKeyInfo.end - rowKeyInfo.start).toString)
+    }
+
+    def getKeyRangePrefix(rangePrefixInfo: KeyStrategy1RangePrefixInfo): (Array[Byte], Array[Byte]) = {
+      val start = Bytes.toBytes(rangePrefixInfo.queryRegion.referenceName + "_" +
+        String.format("%10s", rangePrefixInfo.queryRegion.start.toString).replace(' ', '0'))
```
fnothaft: Ditto here wrt using `.format`

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+      val stop = Bytes.toBytes(rangePrefixInfo.queryRegion.referenceName + "_" +
+        String.format("%10s", rangePrefixInfo.queryRegion.end.toString).replace(' ', '0'))
+      (start, stop)
+    }
+  }
+
+  /**
+   * saves Sequence Dictionary data to HBase
+   *
+   * @param dao
```
fnothaft: Fill in Scaladoc.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+
+  private[hbase] def saveSequenceDictionaryToHBase(dao: HBaseSparkDAO,
+                                                   hbaseTableName: String,
+                                                   sequences: SequenceDictionary,
+                                                   sequenceDictionaryId: String): Unit = {
+
+    val table = dao.getTable(TableName.valueOf(hbaseTableName))
+    val contigs = sequences.toAvro
+
+    val baos: java.io.ByteArrayOutputStream = new ByteArrayOutputStream()
```
fnothaft: We have this code elsewhere (ADAMContext?) so perhaps we can factor it out.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+    val table = dao.getTable(TableName.valueOf(hbaseTableName))
+
+    var resultList = new ListBuffer[Sample]
+
+    sampleIds.foreach((sampleId) => {
+      val get = new Get(Bytes.toBytes(sampleId))
+      val result = table.get(get)
+      val dataBytes = result.getValue(Bytes.toBytes("meta"), Bytes.toBytes("sampledata"))
+
+      val sampleDatumReader: DatumReader[Sample] =
```
fnothaft: You should be able to create the `DatumReader[Sample]` outside of the loop and reuse it, right?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+   *
+   * @param dao HBase Data Access Object
+   * @param hbaseTableName HBase Table Name
+   * @param sampleIds Sample Id list
+   * @param sampleListFile File containing sample IDs, one per line
+   * @param queryRegion Genomic ReferenceRegion used to limit query
+   * @param partitions number of partitions in which to repartition RDD
+   * @return
+   */
+
```
fnothaft: Nit: extra whitespace.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+                                                  hbaseTableName: String,
+                                                  sampleIds: Option[List[String]] = None,
+                                                  sampleListFile: Option[String] = None,
+                                                  queryRegion: Option[ReferenceRegion] = None,
+                                                  partitions: Option[Int] = None,
+                                                  cachingValue: Int = 100): RDD[VariantContext] = {
+
+    val scan = new Scan()
+
+    scan.setCaching(cachingValue)
```
fnothaft: Can we make these user configurable, or document them?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+        var result = new ListBuffer[Genotype]
+        sampleIdsFinal.foreach((sample) => {
+          val sampleVal = curr._2.getColumnCells(Bytes.toBytes("g"), Bytes.toBytes(sample))
+          val decoder = DecoderFactory.get().binaryDecoder(CellUtil.cloneValue(sampleVal.get(0)), null)
+          while (!decoder.isEnd) {
+            result += genotypeDatumReader.read(null, decoder)
+          }
+        })
+
+        result
```
fnothaft: Unused code?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+        val x = result.toList
+
+        val firstGenotype: Genotype = x.head
```
fnothaft: Remove type hints?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+   *
+   * @param dao HBase Data Access Object
+   * @param hbaseTableName HBase Table Name
+   * @param splitsFileName File path and name for file with pre-defined splits of HBase key space
+   */
+
+  def createHBaseGenotypeTable(dao: HBaseSparkDAO, hbaseTableName: String, splitsFileName: Option[String] = None) {
+
+    val tableDescriptor = new HTableDescriptor(TableName.valueOf(hbaseTableName))
+    tableDescriptor.addFamily(new HColumnDescriptor("g".getBytes()).setCompressionType(Algorithm.GZ)
```
fnothaft: Ditto RE: making configurable or documenting.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+      .setDataBlockEncoding(DataBlockEncoding.FAST_DIFF)
+      .setMaxVersions(1))
+
+    if (splitsFileName.isDefined) {
+
+      // splits defines a set of redefined HBase region splits based on an array of Array[Byte] indicating the
+      // split points
+      val splits: Array[Array[Byte]] = Source.fromFile(splitsFileName.get).getLines.toArray
+        .map(x => x.split(" "))
+        .map(g => Bytes.toBytes(g(0) + "_"
```
fnothaft: replace with `.format`

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+          + String.format("%10s", g(1)).replace(' ', '0')))
+
+      dao.admin.createTable(tableDescriptor, splits)
+    } else dao.admin.createTable(tableDescriptor)
+
+    val hbaseTableNameMeta = hbaseTableName + "_meta"
+
+    val tableDescriptorMeta: HTableDescriptor = new HTableDescriptor(TableName.valueOf(hbaseTableNameMeta))
+    tableDescriptorMeta.addFamily(new HColumnDescriptor("meta".getBytes())
+      .setCompressionType(Algorithm.GZ)
```
fnothaft: Let's make configurable or document.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+   * @param stagingFolder - user defined location (on HDFS) to which to save temporary staging files
+   */
+
+  def saveVariantContextRDDToHBaseBulk(dao: HBaseSparkDAO,
+                                       vcRdd: VariantContextRDD,
+                                       hbaseTableName: String,
+                                       sequenceDictionaryId: String,
+                                       stagingFolder: String,
+                                       saveSequenceDictionary: Boolean = true,
+                                       partitions: Option[Int] = None) = {
```
fnothaft: I would vote for removing the ability to repartition inside this command.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+    if (saveSequenceDictionary) saveSequenceDictionaryToHBase(dao,
+      hbaseTableName + "_meta",
+      vcRdd.sequences,
+      sequenceDictionaryId)
+
+    val rdd = if (partitions.isDefined) vcRdd.rdd.repartition(partitions.get)
+    else vcRdd.rdd
+
+    val genodata: RDD[(Array[Byte], List[(String, Array[Byte])])] = rdd.mapPartitions((iterator) => {
+      val genotypebaos: java.io.ByteArrayOutputStream = new ByteArrayOutputStream()
```
fnothaft: I think we have similar code in the `rdd` package, so perhaps we can refactor it out?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+
+        (myRowKey, genotypes)
+      })
+      genotypesForHbase
+    })
+
+    val bulkLoadFlatMap = (t: (Array[Byte], List[(String, Array[Byte])])) => {
+      val data = new ListBuffer[(KeyFamilyQualifier, Array[Byte])]
+      val rowKey = t._1
+      val family = Bytes.toBytes("g")
```
fnothaft: We have `Bytes.toBytes("g")` show up a lot in this file. Perhaps we can factor it out as a constant?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+      t._2.foreach((x) => {
+        val qualifier: String = x._1
+        val value: Array[Byte] = x._2
+        val mykeyFamilyQualifier = new KeyFamilyQualifier(rowKey, family, Bytes.toBytes(qualifier))
+        data += Tuple2(mykeyFamilyQualifier, value)
```
fnothaft: `Tuple2(mykeyFamilyQualifier, value)` -> `(mykeyFamilyQualifier, value)`

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+
+  /**
+   *
+   * @param dao HBase Data Access Object
+   * @param hbaseTableName HBase table name
+   * @param sampleIds List of sample ids
+   * @param sampleListFile File containing one sample id per row
+   * @param partitions number of paritions to use when performing bulk delete
+   */
+
```
fnothaft: Nit: Delete whitespace, finish up scaladocs.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+  def deleteGenotypeSamplesFromHBase(dao: HBaseSparkDAO,
+                                     hbaseTableName: String,
+                                     sampleIds: Option[List[String]] = None,
+                                     sampleListFile: Option[String] = None,
```
fnothaft: Nit: prefix with `opt`

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+                                     partitions: Option[Int] = Some(1)): Unit = {
+
+    val scan = new Scan()
+    scan.setCaching(100)
+    scan.setMaxVersions(1)
+
+    val sampleIdsFinal: List[String] =
+      if (sampleListFile.isDefined) {
```
fnothaft: This should be rewritten as either a `getOrElse` or `fold` or some similar pattern.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
```diff
@@ -0,0 +1,587 @@
...
+    val sampleMetadata = loadSampleMetadataFromHBase(dao, hbaseTableName + "_meta", sampleIds)
+    val genotypes = loadVariantContextsFromHBase(
+      dao,
+      hbaseTableName,
+      Option(sampleIds),
+      queryRegion = queryRegion,
+      partitions = partitions)
+
+    VariantContextRDD(genotypes, sequenceDictionary, sampleMetadata)
+
```
fnothaft: Nit: extra whitespaces.

adam-core/src/main/scala/org/bdgenomics/adam/serialization/ADAMKryoRegistrator.scala
```diff
@@ -67,6 +72,13 @@ class ADAMKryoRegistrator extends KryoRegistrator {
     // Register Avro classes using fully qualified class names
     // Sort alphabetically and add blank lines between packages
 
+    kryo.register(classOf[org.apache.hadoop.hbase.io.ImmutableBytesWritable])
```
fnothaft: Nit: should be sorted later in the file.

adam-core/src/test/scala/org/bdgenomics/adam/hbase/HBaseSuite.scala
```diff
@@ -0,0 +1,111 @@
+package org.bdgenomics.adam.hbase
```
fnothaft: Nit: missing license header.

adam-core/src/test/scala/org/bdgenomics/adam/hbase/HBaseSuite.scala
```diff
@@ -0,0 +1,111 @@
...
+
+import org.bdgenomics.adam.util.ADAMFunSuite
```
fnothaft: Nit: sort imports.

adam-core/src/test/scala/org/bdgenomics/adam/hbase/HBaseSuite.scala
```diff
@@ -0,0 +1,111 @@
...
+
+    assert(genoData.getSampleId === "NA12878")
+    assert(genoData.getContigName === "1")
+    assert(genoData.getStart === 14396)
+    assert(genoData.getEnd === 14400)
+    assert(genoData.getAlleles.get(0) === GenotypeAllele.REF)
+    assert(genoData.getAlleles.get(1) === GenotypeAllele.ALT)
+  }
+
+  sparkAfter("Done with HBase Tests") {
```
fnothaft: Do we need this?

pom.xml
```diff
@@ -1,5 +1,15 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+
```
fnothaft: Nit: extra whitespace before/after.

pom.xml
```diff
@@ -465,6 +475,22 @@
         </exclusions>
       </dependency>
       <dependency>
+        <groupId>org.apache.spark</groupId>
+        <artifactId>spark-streaming_2.10</artifactId>
+        <version>${spark.version}</version>
+        <scope>provided</scope>
+        <exclusions>
+          <exclusion>
```
fnothaft: We shouldn't need these exclusions, IIRC.

pom.xml
```diff
@@ -565,9 +591,33 @@
...
+        <version>1.2.0-cdh5.8.0</version>
+      </dependency>
+      <dependency>
+        <groupId>org.apache.hbase</groupId>
+        <artifactId>hbase-spark</artifactId>
+        <version>1.2.0-cdh5.8.0</version>
+      </dependency>
     </dependencies>
   </dependencyManagement>
-
```
fnothaft: Nit: keep whitespace.

adam-core/pom.xml
jpdna: not sure, but HBase seems to want it, when it is removed, does not compile

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
jpdna: like inConf -> optInConf
or perhaps inConf -> optConf
?
as the "in" prefix was kinda dumb anyhow, but the "opt" is useful as you suggest.

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
jpdna: I'll finish and test other change and return to the null question

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
jpdna: Where/how should I best document this?

adam-core/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala
jpdna: For reason I don't know Scala compiler seems to only accept the explicit Tuple2() which is why I used it.  Maybe some precedence/tokenizing issue  with += and the other nesting.

jpdna:
This branch at 0ab45cd8ba8 builds and passes tests locally for me after coning from my repo.
Any ideas on why it fails when running on Jenkins?  I did run the format script.

jpdna:
@heuermh  suggests this could be a JDK version issue given the byte level comparisons I am making.

I am running java 1.8.0_111 loaclly
and Jenkins is running: JDK_7u60

Do we intended / need to not use java 8 for Jenkins?

heuermh:
Jenkins must be using JDK8, since our build requires it.  I don't know the exact version though.

jpdna:
what's with then the
```JAVA_HOME 	/home/jenkins/tools/hudson.model.JDK/JDK_7u60```
I see here:
https://amplab.cs.berkeley.edu/jenkins/job/ADAM-prb/1734/injectedEnvVars/

**Comment by fnothaft** [comment]
There's a long boring story to why the JAVA_HOME values are configured the way they are, which mostly has to do with Jenkins specific oddities as @shaneknapp can attest to. If you grep later in the build logs for `export JAVA_HOME`, you'll see that we do indeed set the Java version to `jdk1.8.0_60`.

**Comment by jpdna** [comment]
arg - now the remaining problem is that the HBase spark library version I am using seems to not support Spark 2.0 and thus we are getting complaints about lack of that Spark Logging class that went away in Spark 2.0.   I can check into if the newest version of HBase/Spark connection supports Spark 2.0 as it seems to here: https://github.com/apache/hbase/blob/master/hbase-spark/pom.xml
however, I am not sure if that will be compatible with the CDH version we'd want for the cluster.

Ideas about how to best manage these issues of different versions of libraries and Jenkins?
Maven Profiles?

**Comment by jpdna** [comment]
As  you can see here:
https://amplab.cs.berkeley.edu/jenkins//job/ADAM-prb/1735/

My HBase PR is failing only the 2.0.0 column of the matrix
due to:
"[ERROR] error: bad symbolic reference. A signature in HBaseContext.class refers to type Logging"

I take this to be the Spark 2.0 Logging issue.

The first target I see for HBase work is making it work as we have thus far with CDH, and CDH spark/hbase support clearly doesn't support yet Spark 2.- given its continued use of 
`org.apache.spark.logging`
```
https://github.com/cloudera/hbase/blob/cdh5-1.2.0_5.9.0/hbase-spark/src/main/scala/org/apache/hadoop/hbase/spark/HBaseContext.scala#L41
```

In general, it looks like the Spark 2.0 HBase connector may not exist at all:
http://stackoverflow.com/questions/40908891/which-hbase-connector-for-spark-2-0-should-i-use

So - it may be for now impossible for this PR to pass against Spark 2.0 (without re-writing connector).

How should we manage this wrt Jenkins? Thoughts @heuermh  and @fnothaft  ?

**Comment by jpdna** [comment]
Ping for suggestions about how I should deal with this Spark 2.0 compatibility and Jenkins.

**Comment by fnothaft** [comment]
We could move it into a module that is only accessible via a profile and then disable that profile when building for Spark 2. That would be simple.

**Comment by heuermh** [comment]
> We could move it into a module that is only accessible via a profile and then disable that profile when building for Spark 2.

Yep, that would work.  And this would be a good opportunity to bother upstream about it.

**Comment by shaneknapp** [comment]
yep, i agree.

**Comment by jpdna** [comment]
Alright, so do  you suggest I make an HBase module at the same level as the `adam-core` and `adam-cli` modules?   

It seems like the approach need in the POM would then be something like I see here:
http://maven.40175.n5.nabble.com/Excluding-certain-modules-in-a-profile-td81566.html#a81577
```
<project>
  ... 
  <modules>
    <module>module3</module>
    <module>module4</module>
  </modules>

  <profiles>
    <profile>
      <activation>
        <property>
          <name>!exclude-cpp-qa</name>
        </property>
      </activation>
      <modules>
        <module>module1</module>
        <module>module2</module>
      </modules>
    </profile>
  </profiles>
  ... 
</project>
```

Can you comment on what would be needed in the main project POM or elsewhere to selectively disable HBase module for the Spark 2 build?

heuermh:
There are a few different ways to activate a profile, see
http://books.sonatype.com/mvnref-book/reference/profiles-sect-activation.html

jpdna:
> http://books.sonatype.com/mvnref-book/reference/profiles-sect-activation.html

Thanks, but I still feel like I'll be futzing around in the dark here - so to pick  your brain more:

So it looks like in all of  these I'd use the `<activation>` section to activate or with `!` deactivate the module mentioned in the profile.

Some questions:
1) Assuming we are trying to this to work with Jenkins based on what version of Spark is active which I assume Jenkins configuration is doing ( not by specifying the name or a Profile or setting a special env variable directly) what is the parameter/property I will be able to test inside the `<activate>` to determine which version of Spark as active?   I'm not clear if this property is itself allowed to be a POM defined Property.

2) Example 5.3.1 in the doc you pointed to seems like it as an AND combination of values when we'd possibly need an OR like "activate for Spark 1.6.2, 1.6.3, 1.6.4"   I'm not sure how I'll include or exclude these multiple possible versions.  If Jenkins could set an OS env variable like  Spark2=True that would be set for any Spark versions greater than Spark 1.x then `<activation>` could test just for that.

fnothaft:
I would just activate it through a profile as we do with the distribution module.

jpdna:
I could use another set of eyes if someone could look at below:

I'm attempting to move Hbase into a separate module in this new branch dev/b6 at my repo here:
https://github.com/jpdna/adam/tree/dev/b6    

Trying to build after moving the module (skipping test for the moment) I get this error:

[ERROR] /home/paschallj/2_4_2017/adam/adam-hbase/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala:52: error: object rdd is not a member of package org.bdgenomics.adam
[ERROR] import org.bdgenomics.adam.rdd.variant.VariantContextRDD
[ERROR]                            ^
[ERROR] /home/paschallj/2_4_2017/adam/adam-hbase/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala:53: error: object rich is not a member of package org.bdgenomics.adam
[ERROR] import org.bdgenomics.adam.rich.RichVariant
[ERROR]                            ^
[ERROR] /home/paschallj/2_4_2017/adam/adam-hbase/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala:55: error: object models is not a member of package org.bdgenomics.adam
[ERROR] import org.bdgenomics.adam.models.{ ReferencePosition, ReferenceRegion, SequenceDictionary, VariantContext }

coming from:
https://github.com/jpdna/adam/blob/dev/b6/adam-hbase/src/main/scala/org/bdgenomics/adam/hbase/HBaseFunctions.scala#L52
 
Be sure to skip tests for the moment if you try to replicate from:
https://github.com/jpdna/adam/tree/dev/b6

I've got more POM cleanup to do, but first need to figure out why can't currently see:
org.bdgenomics.adam.rdd.variant.VariantContextRDD
from above.   Thanks!

fnothaft:
Continued in #1388. Closing.

PR #1338, opened 2017-01-03
[ADAM-1337] Remove os.{flush,close} calls after writing VCF header.

### LINKED ISSUE #1337: Close called multiple times on VCF header stream
I think this doesn't show up in our tests because the stream returned by Hadoop's LocalFileSystem implementation allows close to be called multiple times, but the stream returned by HDFS does not allow close to be called multiple times.

fnothaft:
Resolves #1337. The VCF writer calls close on the underlying stream.


heuermh:
Thank you, @fnothaft

PR #1357, opened 2017-01-19
[ADAM-1352] Clean up consensus model usage.

### LINKED ISSUE #1352: ConsensusModel's can't really be instantiated
Their constructors are all package private. I think that's the right approach, but we should have a singleton class that creates them. I'll:

* Make a singleton to create consensus models
* Improve the known INDEL consensus model, because it needs a bit of TLC
* Add functions for merging multiple consensus models

I need this for https://github.com/bigdatagenomics/avocado/issues/200.

fnothaft:
Resolves #1352:

* Fixed issues with consensus model instantiation.
* Added union consensus model (to support multiple models).
* Cleaned up consensus model documentation.
* Call distinct in IndelRealigner on consensus sequences to eliminate dupes.
* Added test for known indel model. Fixed off by one bug.
* Add test for union model.

### DIFF
--- a/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
+++ b/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
@@ -174,9 +174,9 @@ class Transform(protected val args: TransformArgs) extends BDGSparkCommand[Trans
 
       // fold and get the consensus model for this realignment run
       val consensusGenerator = Option(args.knownIndelsFile)
-        .fold(new ConsensusGeneratorFromReads().asInstanceOf[ConsensusGenerator])(
-          new ConsensusGeneratorFromKnowns(_, rdd.rdd.context).asInstanceOf[ConsensusGenerator]
-        )
+        .fold(ConsensusGenerator.fromReads)(file => {
+          ConsensusGenerator.fromKnownIndels(rdd.rdd.context.loadVariants(file))
+        })
 
       // run realignment
       val realignmentRdd = rdd.realignIndels(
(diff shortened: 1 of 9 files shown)


adam-cli/src/main/scala/org/bdgenomics/adam/cli/Transform.scala
```diff
@@ -174,9 +174,9 @@ class Transform(protected val args: TransformArgs) extends BDGSparkCommand[Trans
 
       // fold and get the consensus model for this realignment run
       val consensusGenerator = Option(args.knownIndelsFile)
-        .fold(new ConsensusGeneratorFromReads().asInstanceOf[ConsensusGenerator])(
-          new ConsensusGeneratorFromKnowns(_, rdd.rdd.context).asInstanceOf[ConsensusGenerator]
-        )
+        .fold(ConsensusGenerator.fromReads)(file => {
+          ConsensusGenerator.fromKnownIndels(rdd.rdd.context.loadVariants(file))
+        })
```
heuermh: this reads a lot better

adam-core/src/main/scala/org/bdgenomics/adam/models/IndelTable.scala
```diff
@@ -76,11 +64,12 @@ private[adam] object IndelTable {
           val deletionLength = v.getReferenceAllele.length - v.getAlternateAllele.length
           val start = v.getStart + v.getAlternateAllele.length
 
-          Consensus("", ReferenceRegion(referenceName, start, start + deletionLength))
+          Consensus("", ReferenceRegion(referenceName, start, start + deletionLength + 1))
```
heuermh: oops  :)  was this a bug, or something you found when cleaning up?

fnothaft: This was a bug.

fnothaft:
This is good to go now.

heuermh:
Thank you, @fnothaft!

PR #1363, opened 2017-01-23
[ADAM-1362] Fixing issue where FromKnowns consensus model fails if no reads hit a target.

### LINKED ISSUE #1362: FromKnowns consensus generator fails if no reads overlap a consensus
This happens because we call a `.min` on an empty collection:

```
  org.apache.spark.SparkException: Job aborted due to stage failure: Task 0 in stage 8.0 failed 1 times, most recent failure: Lost task 0.0 in stage 8.0 (TID 28, localhost): java.lang.UnsupportedOperationException: empty.min
	at scala.collection.TraversableOnce$class.min(TraversableOnce.scala:209)
	at scala.collection.AbstractTraversable.min(Traversable.scala:105)
	at org.bdgenomics.adam.algorithms.consensus.ConsensusGeneratorFromKnowns.findConsensus(ConsensusGeneratorFromKnowns.scala:82)
	at org.bdgenomics.adam.rdd.read.realignment.RealignIndels$$anonfun$realignTargetGroup$1.apply(RealignIndels.scala:260)
	at org.bdgenomics.adam.rdd.read.realignment.RealignIndels$$anonfun$realignTargetGroup$1.apply(RealignIndels.scala:240)
	at scala.Option.fold(Option.scala:157)
	at org.apache.spark.rdd.Timer.time(Timer.scala:48)
	at org.bdgenomics.adam.rdd.read.realignment.RealignIndels.realignTargetGroup(RealignIndels.scala:240)
	at org.bdgenomics.adam.rdd.read.realignment.RealignIndels$$anonfun$realignIndels$1.apply(RealignIndels.scala:498)
	at org.bdgenomics.adam.rdd.read.realignment.RealignIndels$$anonfun$realignIndels$1.apply(RealignIndels.scala:498)
```

fnothaft:
Fixes #1362:

* Adding ConsensusGeneratorFromKnownsSuite.
* ConsensusGeneratorFromKnowns drops any targets that did not get hit with any reads, instead of trying to identify the full reference region overlapped by reads. While this case (target with no reads) cannot happen in the FromReads consensus model, it can happen in the FromKnowns model (e.g., use all known INDELs from 1000G with a WES dataset).
* Additionally, tracked down a minor off by one bug in RealignIndels that caused one index in the consensus sequence to not get tested. This could cause reads that aligned to the end of the target to not get realigned.
* To improve legibility, broke up a set of long lines in RealignIndels.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/algorithms/consensus/ConsensusGeneratorFromKnowns.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/algorithms/consensus/ConsensusGeneratorFromKnowns.scala
@@ -49,9 +49,7 @@ private[adam] class ConsensusGeneratorFromKnowns(rdd: RDD[Variant]) extends Cons
   def targetsToAdd(): Option[RDD[IndelRealignmentTarget]] = {
 
     Some(rdd.filter(v => v.getReferenceAllele.length != v.getAlternateAllele.length)
-      .map(v => ReferenceRegion(v.getContigName,
-        v.getStart,
-        v.getStart + v.getReferenceAllele.length))
+      .map(v => ReferenceRegion(v))
       .map(r => new IndelRealignmentTarget(Some(r), r)))
   }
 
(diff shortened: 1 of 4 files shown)


heuermh:
Thank you, @fnothaft!

PR #1366, opened 2017-01-23
[ADAM-1365] Apply validation stringency to reads on missing contigs when MD tagging

### LINKED ISSUE #1365: ValidationStringency in MDTagging should apply to reads on unknown references
Right now, if you have a read that is mapped to an unknown reference contig (not provided in reference build) when you are recomputing MD tags, it throws an exception. This makes sense, but we should be able to suppress these exceptions with ValidationStringency (i.e., if the read is mapped to a decoy contig), which is already passed in to MdTagging.

fnothaft:
Resolves #1365. If a read is mapped to a contig that is missing from the broadcasted fragment map, we look at the user's provided validation stringency before throwing an exception.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/MDTagging.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/MDTagging.scala
@@ -71,8 +71,20 @@ private[read] case class MDTagging(
         contig <- Option(read.getContigName)
         if read.getReadMapped
       } yield {
-        maybeMDTagRead(read, referenceFileB.value
-          .extract(ReferenceRegion.unstranded(read)))
+        try {
+          maybeMDTagRead(read, referenceFileB.value
+            .extract(ReferenceRegion.unstranded(read)))
+        } catch {
+          case t: Throwable => {
+            if (validationStringency == ValidationStringency.STRICT) {
+              throw t
+            } else if (validationStringency == ValidationStringency.LENIENT) {
+              log.warn("Caught exception when processing read %s: %s".format(
+                read.getContigName, t))
+            }
+            read
+          }
+        }
       }).getOrElse({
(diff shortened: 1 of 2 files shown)


adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/MDTagging.scala
```diff
@@ -71,8 +71,20 @@ private[read] case class MDTagging(
...
         if read.getReadMapped
       } yield {
-        maybeMDTagRead(read, referenceFileB.value
-          .extract(ReferenceRegion.unstranded(read)))
+        try {
+          maybeMDTagRead(read, referenceFileB.value
+            .extract(ReferenceRegion.unstranded(read)))
+        } catch {
+          case t: Throwable => {
+            if (validationStringency == ValidationStringency.STRICT) {
```
heuermh: The more we use `ValidationStringency` outside of forwarding to htsjdk, the less I like using their enum.

I know this isn't the place; what do you think of an ADAM-specific enum that maps to `ValidationStringency` where necessary?  We could use the opportunity to come up with a better name, since it isn't really about validation, it is more about error reporting.

fnothaft: I'm +0. I don't love ValidationStringency, but I'm a bit too lazy to deconvolve it from our codebase. I guess I'd say that it is on the sorted list of refactors that I'd like to do, but lower down on the list. That being said, it's probably not that much work. CC @ryan-williams for thoughts as well.

heuermh: Yeah that's where I'm at too.  If I had a better name ready, I'd do it.

heuermh:
Thank you, @fnothaft!

PR #1375, opened 2017-01-25
[ADAM-1372] Increase visibility for DefaultHeaderLines.allHeaderLines

### LINKED ISSUE #1372: Increase visibility for SupportedHeaderLines.allHeaderLines
```
[INFO] Compiling 6 Scala sources to target/scala-2.10.6/classes...
[ERROR] .../Freebayes.scala:77: object SupportedHeaderLines in package converters
cannot be accessed in package org.bdgenomics.adam.converters
[ERROR] implicit val uFormatter = new VCFOutFormatter(SupportedHeaderLines.allHeaderLines)
```
In this specific use case I'd like to add Freebayes-specific headers to the ones we already support, something like
```scala
val freebayesHeaderLines = Seq( ... )
val headerLines = SupportedHeaderLines.addHeaderLines ++ freebayesHeaderLines
implicit val uFormatter = new VCFOutFormatter(headerLines)
```

heuermh:
Fixes #1372

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/converters/DefaultHeaderLines.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/converters/DefaultHeaderLines.scala
@@ -27,9 +27,9 @@ import htsjdk.variant.vcf.{
 }
 
 /**
- * All header lines for VCF INFO and GT format fields that are supported in ADAM.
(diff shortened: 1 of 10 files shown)


fnothaft:
Merged! Thanks @heuermh!

PR #1389, opened 2017-02-13
[ADAM-1381] Fix Variant end position.

### LINKED ISSUE #1381: Set the wrong value for end for symbolic alts
See https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantContextConverter.scala#L1851. We should not be taking the reference allele length; we should use the end position value...

fnothaft:
Resolves #1381. Sets variant end position to the proper site for symbolic alleles.


fnothaft:
Hi @heuermh! If this looks good to you as-is, we can close https://github.com/bigdatagenomics/bdg-formats/pull/125 unmerged.

heuermh:
Thank you, @fnothaft!

PR #1324, opened 2016-12-23
Efficient Joins and (re)Partitioning

devin-petersohn:
Ready for review.

PR #1401, opened 2017-02-24
Added instrumentation timers around joins.

fnothaft:
Had this on my branch testing #1324 with the INDEL realigner. Thought it was more generally useful.


heuermh:
Thank you, @fnothaft

PR #1427, opened 2017-03-07
Add cache argument to loadFeatures, additional Feature timers

### LINKED ISSUE #1321: FeatureRDD instantiation tries to cache the RDD
Currently, when you create a `FeatureRDD` via, say, `sc.loadBed()`, the `apply` function tries to cache the RDD bc it first computes a `SequenceDictionary` off of it:

https://github.com/bigdatagenomics/adam/blob/47fa49cb8904f770307a22ed0dffcbd21c9c6b06/adam-core/src/main/scala/org/bdgenomics/adam/rdd/feature/FeatureRDD.scala#L95

I am working on my laptop on a 120 GB bed dataset and this is not gonna end well.  This feels unexpected...is it necessary?  Perhaps allow the user to specify the storage level?  Thoughts?

heuermh:
Fixes #1321

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
@@ -1148,6 +1148,7 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
    * Loads features stored in GFF3 format.
    *
    * @param filePath The path to the file to load.
+   * @param cache True if the rdd should be cached before building the SequenceDictionary.
    * @param minPartitions An optional minimum number of partitions to load. If
    *   not set, falls back to the configured Spark default parallelism.
    * @param stringency Optional stringency to pass. LENIENT stringency will warn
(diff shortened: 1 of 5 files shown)


adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -1156,18 +1157,20 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
    * @return Returns a FeatureRDD.
    */
   def loadGff3(filePath: String,
+               cache: Boolean = true,
```
fnothaft: I would set the storage level instead of caching these. Provides more flexibility, esp. if you're running out-of-core.

heuermh: I thought storage level was for `rdd.persist`, as used in cli `Transform`?  This uses `rdd.cache`.

fnothaft: `rdd.cache` = `rdd.persist(StorageLevel.MEMORY_ONLY)`

heuermh: Ah, so sorry if I'm being thick, we should pass the StorageLevel around the feature APIs instead, and then replace the call to `rdd.cache` in `FeatureRDD` with `rdd.persist(storageLevel)`?

fnothaft: No prob; we all have our thick days. What you suggested (passing the StorageLevel) is what I was thinking.

fnothaft:
Jenkins, test this please.

fnothaft:
Merged! Thanks @heuermh!

PR #1443, opened 2017-03-21
[ADAM-1442] Fix thread pool deadlock in GenomicRDD.pipe

### LINKED ISSUE #1442: Thread pool will block inside of pipe command for streams too large to buffer
Sigh, this is a silly one on my behalf! Inside of the pipe command, we return the output iterator by getting the `Future` from a `Callable` which is submitted to a thread pool. This only works if the output stream is small enough that it can be buffered fully. On large inputs, this causes a deadlock.

fnothaft:
Resolves #1442. Eliminates use of thread pool in GenomicRDD and instead launches the InFormatterRunner (which extends `Callable`) as a thread and runs the OutFormatterRunner in the original thread that launched the piped subprocess.

Additionally, rewrote the `AnySAMOutFormatter` to improve performance.

With this, BWA-on-Cannoli works (https://github.com/heuermh/cannoli/pull/2).

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
@@ -18,7 +18,7 @@
 package org.bdgenomics.adam.rdd
 
 import htsjdk.variant.vcf.{ VCFHeader, VCFHeaderLine }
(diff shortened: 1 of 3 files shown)


fnothaft:
Jenkins, retest this please.

heuermh:
Thank you, @fnothaft!

Issue #1439, opened 2017-03-15
FeatureRDD.apply() does not allow addition of other parameters with defaults in the case class


devin-petersohn:
I blame Scala for this:

Scala will not allow us to have multiple overloaded methods to have parameters with defaults. Currently, we have a default for `storageLevel` in [FeatureRDD Line 117](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/feature/FeatureRDD.scala#L117). 

In #1324 we need a default for the partition map when the objects are created shown on [Line 242 of the PR branch](https://github.com/devin-petersohn/adam/blob/partitioner/adam-core/src/main/scala/org/bdgenomics/adam/rdd/feature/FeatureRDD.scala#L242), so I'm a little stuck. I am getting this error:

`adam/adam-core/src/main/scala/org/bdgenomics/adam/rdd/feature/FeatureRDD.scala:105: error: in object FeatureRDD, multiple overloaded alternatives of apply define default arguments`

We only have a couple of options and I don't know that I like any of them. @fnothaft @heuermh I'd love some thoughts on this.

heuermh:
Yeah, I see the problem.

With `FeatureRDD` there are two paths for instantiation, one where the sequence dictionary is known (IntervalList format currently, and possibly GFF3 later if/when we add header support), and where it is not known and needs to be calculated (BED, GTF/GFF2, NarrowPeak formats). It'd be nice to have defaults whereever necessary, but if we have to pick and choose that's not the end of the world.

fnothaft:
I'd actually rather those two paths be two separate methods without defaults.

heuermh:
It didn't really feel right passing `storageLevel` around, especially to a constructor.  Perhaps the creation of the sequence dictionary should move to where the native file format parsing happens instead?

PR #1446, opened 2017-03-21
[ADAM-1445] Make apply method for FragmentRDD public.

### LINKED ISSUE #1445: Apply method in FragmentRDD is package private
This makes access to the case class generated apply method package private. We did this to keep a method in the object package private. This is not a big issue, since the constructor on `FragmentRDD` is public still.

fnothaft:
Resolves #1445.


heuermh:
I want to merge this sooner rather than later as I am about to push a fix for #1439.

fnothaft:
@heuermh SGTM! Fire when ready.

heuermh:
Thank you, @fnothaft!

PR #1448, opened 2017-03-21
[ADAM-1424] Adding test for reads dropped in 0.21.0.

### LINKED ISSUE #1424: Question : Why the number of paired sequence in adam-0.21.0 less than adam-0.19.0?
I load fastq and count by :

```
    val rdd= ac.loadAlignments(fastq1,filePath2Opt = Option(fastq2)).rdd
    println("count" + rdd.count())
```
fastq1 and fastq2 are 10000+10000=20000 sequences
The count number is 15458 with adadm-0.21.0,but the count is 20000 both in adam-0.18.2 and adam-0.19.0. I do not known the reason.

I try to load fastq1 and fastq2:

```
    val df4 = ac.loadAlignments(fastq1).rdd
    println(df4.count)
    val df5 = ac.loadAlignments(fastq2).rdd
    println(df5.count)
```
the count are

```
7729
7729
```

Why both are not 10000?

Question : Why the number of paired sequence in adam-0.21.0 less than adam-0.19.0?

Thanks

heuermh:
Fixes #1424.

Adapted unit test contributed by @xubo245 to use existing test resources.


heuermh:
Jenkins, retest this please.

adam-core/src/test/scala/org/bdgenomics/adam/rdd/ADAMContextSuite.scala
```diff
@@ -494,4 +494,15 @@ class ADAMContextSuite extends ADAMFunSuite {
       case _ => fail("unexpected variant start " + v.getStart)
     })
   }
+
+  sparkTest("loadAlignments should not fail on single-end and paired-end fastq reads") {
+    val readsFilepath1 = testFile("proper_pairs_1.fq")
```
fnothaft: OOC, does this test fail on 0.21.0?

heuermh: Nope

heuermh:
Squashed to a single commit

PR #1460, opened 2017-03-27
[ADAM-1459] Don't split FASTQ when compressed.

### LINKED ISSUE #1459: FASTQ may try to split GZIPed files
We need to implement [isSplittable](https://hadoop.apache.org/docs/r2.7.1/api/org/apache/hadoop/mapred/FileInputFormat.html#isSplitable(org.apache.hadoop.fs.FileSystem, org.apache.hadoop.fs.Path)) in the two FastqInputFormats.

fnothaft:
Resolves #1459.


heuermh:
Thank you, @fnothaft

PR #1468, opened 2017-04-01
[ADAM-1467] Skip `concat` call if there is only one shard.

### LINKED ISSUE #1467: Parallel file merger fails if the output file is smaller than the HDFS block size
If this happens, we don't have any files to merge _into_ the head file (i.e., the second argument of [concat](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ParallelFileMerger.scala#L332) is empty).

fnothaft:
Resolves #1467.


heuermh:
Thank you, @fnothaft!

Issue #1438, opened 2017-03-14
No asSingleFile parameter for saveAsFasta in NucleotideContigFragmentRDD


akmorrow13:
This could be solved by moving https://github.com/bigdatagenomics/adam/blob/6ee0b8b28fdbdb5e52041519ac9a671f741bc5e5/adam-core/src/main/scala/org/bdgenomics/adam/rdd/feature/FeatureRDD.scala#L312 from FeatureRDD to a more central location to allow NucleotideContigFragmentRDD to save single file fastas.

fnothaft:
Perhaps we could move that into `FileMerger`?

heuermh:
Yeah the save stuff needs work in a lot of places; see e.g. https://github.com/bigdatagenomics/adam/issues/1295#issuecomment-265571803.  I was planning to address some of it when implementing the fix for https://github.com/bigdatagenomics/adam/issues/1327, waiting for after the next bdg-utils release for ADAM version 0.23.0.

heuermh:
For reference
```bash
$ jar -tvf adam-assembly/target/adam_2.10-0.21.1-SNAPSHOT.jar | grep ".class" | grep bdgenomics | grep -v '\$' | awk '{print $8}' | sed -e 's/\//./g' -e 's/.class//g' | xargs javap -private -s -classpath adam-assembly/target/adam_2.10-0.21.1-SNAPSHOT.jar | grep save | grep -v "default"
  public <U extends org.apache.avro.specific.SpecificRecordBase> void saveAvro(java.lang.String, org.apache.spark.SparkContext, org.apache.avro.Schema, scala.collection.Seq<U>, scala.reflect.ClassTag<U>);
  public void saveRddAsParquet(org.bdgenomics.utils.cli.SaveArgs);
  public void saveRddAsParquet(java.lang.String, int, int, org.apache.parquet.hadoop.metadata.CompressionCodecName, boolean, scala.Option<org.apache.avro.Schema>);
  public void saveMetadata(java.lang.String);
  public void saveAsParquet(org.bdgenomics.utils.cli.SaveArgs);
  public void saveAsParquet(java.lang.String, int, int, org.apache.parquet.hadoop.metadata.CompressionCodecName, boolean);
  public void saveAsParquet(java.lang.String, java.lang.Integer, java.lang.Integer, org.apache.parquet.hadoop.metadata.CompressionCodecName, java.lang.Boolean);
  public void saveAsParquet(java.lang.String);
[...]

Issue #1295, opened 2016-11-30
Clarifying ADAMSaveAnyArgs fields


ryan-williams:
At the very least, one or both of these should probably be renamed? ([code](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMRDDFunctions.scala#L53-L69)):

```scala
  /**
   * If true and saving as a legacy format, we will write shards so that they
   * can be merged into a single file.
   *
   * @see deferMerging
   */
  var asSingleFile: Boolean

  /**
   * If true and asSingleFile is true, we will not merge the shards once we
   * write them, and will leave them for the user to merge later. If false and
   * asSingleFile is true, then we will merge the shards on write. If
   * asSingleFile is false, this is ignored.
   *
   * @see asSingleFile
   */
  var deferMerging: Boolean
```

It seems like `asSingleFile` now really means "written such that each shard can be merged, i.e. BAM header is not prepended on each partition, only the first one", and `deferMerging` means "leave the unmerged, headerless partitions in place"?

There are actually 3 possibilities, right? Would an enum be clearer?

fnothaft:
> There are actually 3 possibilities, right? Would an enum be clearer?

I would agree with that. An Enum would SGTM!

heuermh:
I'd be ok if `ADAMSaveAnyArgs` went away and the enum moved to `SaveArgs` in utils:

https://github.com/bigdatagenomics/utils/blob/master/utils-cli/src/main/scala/org/bdgenomics/utils/cli/ParquetArgs.scala#L30

`sortFastqOutput` also feels misplaced.

heuermh:
On further consideration, `SaveArgs`, `LoadFileArgs`, `SaveFileArgs`, and `ParquetLoadSaveArgs` are all rather worthless, since we always redefine INPUT and OUTPUT meta variables so that we can be more specific about supported file types in the description.

heuermh:
Note there is also `disableFastConcat` now

PR #1479, opened 2017-04-07
[ADAM-1480] Add switch to disable the fast concat method.

### LINKED ISSUE #1478: Fall back to sequential concat when HDFS concat fails
There's a dizzying array of undocumented cases that will cause HDFS' concat method to fail. The most recent one that I found is running on files in an encrypted HDFS directory:

```
17/04/07 01:45:00 ERROR yarn.ApplicationMaster: User class threw exception: org.apache.hadoop.ipc.RemoteException(org.apache.hadoop.HadoopIllegalArgumentException): concat can not be called for files in an encryption zone.
	at org.apache.hadoop.hdfs.server.namenode.FSNamesystem.concatInternal(FSNamesystem.java:2116)
	at org.apache.hadoop.hdfs.server.namenode.FSNamesystem.concatInt(FSNamesystem.java:2081)
	at org.apache.hadoop.hdfs.server.namenode.FSNamesystem.concat(FSNamesystem.java:2043)
	at org.apache.hadoop.hdfs.server.namenode.NameNodeRpcServer.concat(NameNodeRpcServer.java:814)
	at org.apache.hadoop.hdfs.server.namenode.AuthorizationProviderProxyClientProtocol.concat(AuthorizationProviderProxyClientProtocol.java:280)
	at org.apache.hadoop.hdfs.protocolPB.ClientNamenodeProtocolServerSideTranslatorPB.concat(ClientNamenodeProtocolServerSideTranslatorPB.java:562)
	at org.apache.hadoop.hdfs.protocol.proto.ClientNamenodeProtocolProtos$ClientNamenodeProtocol$2.callBlockingMethod(ClientNamenodeProtocolProtos.java)
	at org.apache.hadoop.ipc.ProtobufRpcEngine$Server$ProtoBufRpcInvoker.call(ProtobufRpcEngine.java:617)
	at org.apache.hadoop.ipc.RPC$Server.call(RPC.java:1073)
	at org.apache.hadoop.ipc.Server$Handler$1.run(Server.java:2086)
	at org.apache.hadoop.ipc.Server$Handler$1.run(Server.java:2082)
	at java.security.AccessController.doPrivileged(Native Method)
	at javax.security.auth.Subject.doAs(Subject.java:422)
	at org.apache.hadoop.security.UserGroupInformation.doAs(UserGroupInformation.java:1693)
	at org.apache.hadoop.ipc.Server$Handler.run(Server.java:2080)
```

Yahtzee! Anyways, I'm thinking that the reasonable way to move forward is to fall back on the serial concat code when parallel concatenation fails.

fnothaft:
Resolves #1478.

### DIFF
--- a/adam-cli/src/main/scala/org/bdgenomics/adam/cli/ADAM2Vcf.scala
+++ b/adam-cli/src/main/scala/org/bdgenomics/adam/cli/ADAM2Vcf.scala
@@ -56,6 +56,10 @@ class ADAM2VcfArgs extends Args4jBase with ParquetArgs {
   @Args4jOption(required = false, name = "-single", usage = "Save as a single VCF file.")
   var single: Boolean = false
 
+  @Args4jOption(required = false, name = "-disable_fast_concat",
+    usage = "Disables the parallel file concatenation engine.")
+  var disableFastConcat: Boolean = false
+
   @Args4jOption(required = false, name = "-stringency", usage = "Stringency level for various checks; can be SILENT, LENIENT, or STRICT. Defaults to STRICT")
   var stringency: String = "STRICT"
 }
@@ -95,6 +99,7 @@ class ADAM2Vcf(val args: ADAM2VcfArgs) extends BDGSparkCommand[ADAM2VcfArgs] wit
 
     maybeSortedVcs.saveAsVcf(args.outputPath,
       asSingleFile = args.single,
-      stringency)
+      stringency,
+      disableFastConcat = args.disableFastConcat)
   }
 }
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMRDDFunctions.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMRDDFunctions.scala
@@ -64,6 +64,12 @@ trait ADAMSaveAnyArgs extends SaveArgs {
    * @see asSingleFile
    */
   var deferMerging: Boolean
+
+  /**
+   * If asSingleFile is true and deferMerging is false, disables the use of the
+   * fast file concatenation engine.
+   */
+  var disableFastConcat: Boolean
 }
 
 private[rdd] abstract class ADAMRDDFunctions[T <% IndexedRecord: Manifest] extends Serializable with Logging {
--- a/adam-apis/src/test/java/org/bdgenomics/adam/apis/java/JavaADAMCoverageConduit.java
+++ b/adam-apis/src/test/java/org/bdgenomics/adam/apis/java/JavaADAMCoverageConduit.java
@@ -34,7 +34,7 @@ final class JavaADAMCoverageConduit {
         // make temp directory and save file
         Path tempDir = Files.createTempDirectory("javaAC");
         String fileName = tempDir.toString() + "/testRdd.coverage.adam";
-        recordRdd.save(fileName, false);
+        recordRdd.save(fileName, false, false);
 
         // create a new adam context and load the file
         JavaADAMContext jac = new JavaADAMContext(ac);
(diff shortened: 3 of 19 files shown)


adam-cli/src/main/scala/org/bdgenomics/adam/cli/ADAM2Vcf.scala
```diff
@@ -56,6 +56,10 @@ class ADAM2VcfArgs extends Args4jBase with ParquetArgs {
   @Args4jOption(required = false, name = "-single", usage = "Save as a single VCF file.")
   var single: Boolean = false
 
+  @Args4jOption(required = false, name = "-disable_fast_concat",
+    usage = "Disables the parallel file concatenation engine.")
+  var disableFastConcat: Boolean = false
```
heuermh: Not a fan of double-negatives; is there a better way to do args4j flags that default to true?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMRDDFunctions.scala
```diff
@@ -64,6 +64,12 @@ trait ADAMSaveAnyArgs extends SaveArgs {
    * @see asSingleFile
    */
   var deferMerging: Boolean
+
+  /**
+   * If asSingleFile is true and deferMerging is false, disables the use of the
+   * fast file concatenation engine.
+   */
+  var disableFastConcat: Boolean
```
heuermh: Per this, #1295, and #1438, perhaps we should brainstorm on how to clean up the save methods and ensure consistency.  One or more enums, better SaveArgs classes, a context map, etc.

heuermh:
I'd be willing to merge this without changes.  It adds more pressure towards a future refactoring of save methods for consistency.

fnothaft:
@heuermh with #1487 in the pipe, shall we merge this as is?

fnothaft:
Ping.

heuermh:
> Not a fan of double-negatives; is there a better way to do args4j flags that default to true?

What I'm thinking is something like this
http://stackoverflow.com/questions/15008758/parsing-boolean-values-with-argparse/31347222#31347222

where instead of `-disable` arguments that default to false, we have positive and negative arguments in a mutally exclusive group with the positive one being defaulted to true.  That way the field can be named in the positive and code reads better.  I'm not quite sure how to set up such with Args4j though.

fnothaft:
I don't really see the `-disable` "negative" name as an issue, and I don't have the bandwidth to track down how to do what you're talking about with Args4j. Can we merge this as is?

heuermh:
Args4j can simulate mutually exclusive groups with forbids, however I don't see a way of combining two `--feature` and `--no-feature` style options into a single value as I describe above.

An alternative would be to use `ExplicitBooleanOptionHandler` and default to true, i.e.
```scala
@Args4jOption
  (
    required = false,
    name = "-fast_concat",
    usage = "Enable the parallel file concatenation engine. Defaults to true.",
    handler=classOf[ExplicitBooleanOptionHandler]
  )
var fastConcat: Boolean = true
```

Then these would be true
```
$ adam-submit -- adam2vcf -single foo.adam foo.vcf
$ adam-submit -- adam2vcf -single -fast_concat true foo.adam foo.vcf
```
and this would be false
```
$ adam-submit -- adam2vcf -single -fast_concat false foo.adam foo.vcf
```
and this would fail
```
$ adam-submit -- adam2vcf -single -fast_concat foo.adam foo.vcf
"foo.adam" is not a legal boolean value
 VCF    : The VCF file to convert
 ADAM   : Location to write ADAM Variant data
...
```

Would that be an improvement?

heuermh:
Created #1503 for further discussion

heuermh:
Thank you, @fnothaft!

Issue #1481, opened 2017-04-10
Refactor ADAMContext loadXxx methods for consistency


heuermh:
Suggestions:

`@param` doc strings should be consistent
`@return` doc strings should be consistent
`log.info` messages across load methods should be consistent
Method parameters of type `Path` should be `path`
Method parameters of type `String` should be `pathName`
Method parameters of type `Option` should be named `optXxx`
Include `StorageLevel` parameter if `cache()` is called
All load methods should include `stringency` parameter with `STRICT` default
All load methods should handle lenient and silent stringency consistently
All load methods should instrument the RDD if `Metrics.isRecording`
All load methods should be wrapped in `LoadXxx.time {`  metrics
Parquet load methods should include `optPredicate` and `optProjection` parameters with `None` defaults
Type-guessing methods should use methods similar to `isVcfExt` for consistency
Type-guessing methods should explicitly document support for compressed files
Type-guessing load methods should include `optPredicate` and `optProjection` parameters with `None` defaults

heuermh:
```bash
$ javap -classpath adam-core/target/adam-core_2.10-0.23.0-SNAPSHOT.jar org.bdgenomics.adam.rdd.ADAMContext | grep load | grep -v default | sort
  public <T extends org.apache.avro.specific.SpecificRecordBase> scala.collection.Seq<T> org$bdgenomics$adam$rdd$ADAMContext$$loadAvro(java.lang.String, org.apache.avro.Schema, scala.reflect.ClassTag<T>);
  public <T> org.apache.spark.rdd.RDD<T> loadParquet(java.lang.String, scala.Option<org.apache.parquet.filter2.predicate.FilterPredicate>, scala.Option<org.apache.avro.Schema>, scala.Function1<T, org.apache.avro.specific.SpecificRecord>, scala.reflect.Manifest<T>);
  public org.bdgenomics.adam.models.RecordGroupDictionary loadAvroReadGroupMetadata(java.lang.String);
  public org.bdgenomics.adam.models.RecordGroupDictionary loadBamReadGroups(htsjdk.samtools.SAMFileHeader);
  public org.bdgenomics.adam.models.RecordGroupDictionary org$bdgenomics$adam$rdd$ADAMContext$$loadAvroReadGroupMetadataFile(java.lang.String);
  public org.bdgenomics.adam.models.SequenceDictionary loadAvroSequences(java.lang.String);
[...]

PR #1482, opened 2017-04-10
[ADAM-1456] Remove .{type}.adam file extension conversions in type-guessing methods.

### LINKED ISSUE #1456: Don't convert .{different-type}.adam in loadAlignments and loadFragments
I don't like the `.{type}.adam` file extension conversions in the type-guessing methods, since we don't use or document that convention elsewhere.

In `ADAMContext.loadFragments`
```scala
} else if (filePath.endsWith(".reads.adam")) {
  log.info(s"Loading $filePath as ADAM AlignmentRecords and converting to Fragments.")
  loadAlignments(filePath).toFragments
```

And `ADAMContext.loadAlignments`
```scala
} else if (filePath.endsWith("contig.adam")) {
  log.info(s"Loading $filePath as Parquet of NucleotideContigFragment and converting to AlignmentRecords. Projection is ignored.")
  AlignmentRecordRDD.unaligned(loadParquetContigFragments(filePath).toReads)
```

heuermh:
Fixes #1456.

See also #1481.


fnothaft:
Merged! Thanks @heuermh!

PR #1489, opened 2017-04-14
[ADAM-1488] Only increment BQSR min quality by 33 once.

### LINKED ISSUE #1488: Min Phred score gets bumped by 33 twice in BQSR
Paranoia sweeps the nation as an implicit conversion from Char to Int allows the `minAcceptablePhred` score to be incremented by 33 twice in the AlignmentRecordRDD->BaseQualityScoreRecalibration codepath. Troubled developers ask each other: could my `Char`s be the next target of implicit conversion? Find out more tonight.

For those watching at home, the first bump is [here](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDD.scala#L648) and the second bump is [here](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/recalibration/BaseQualityRecalibration.scala#L349).

fnothaft:
Resolves #1488.


heuermh:
Thank you, @fnothaft

PR #1491, opened 2017-04-14
[ADAM-1377] Adding fragment InFormatter for Bowtie tab6 format

### LINKED ISSUE #1377: Add tab5/6 as native output format for AlignmentRecordRDD
From https://github.com/BenLangmead/bowtie2/issues/75:

`--tab5` Each read or pair is on a single line. An unpaired read line is `[name]\t[seq]\t[qual]\n`. A paired-end read line is `[name]\t[seq1]\t[qual1]\t[seq2]\t[qual2]\n`. An input file can be a mix of unpaired and paired-end reads and Bowtie 2 recognizes each according to the number of fields, handling each as it should.

`--tab6` is similar except, for paired-end reads, the second end can have a different name from the first: `[name1]\t[seq1]\t[qual1]\t[name2]\t[seq2]\t[qual2]\n`.

heuermh:
Fixes #1377.  Work in progress, need to write tests and eliminate copy-pasta.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/converters/AlignmentRecordConverter.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/converters/AlignmentRecordConverter.scala
@@ -30,31 +30,26 @@ import scala.collection.JavaConversions._
 class AlignmentRecordConverter extends Serializable {
 
   /**
-   * Converts a single record to FASTQ. FASTQ format is:
-   *
-   * {{{
-   * @readName
-   * sequence
-   * +<optional readname>
-   * ASCII quality scores
-   * }}}
+   * Prepare a single record for conversion to FASTQ and similar formats by
+   * splitting into a tuple of (name, sequence, qualityScores).
    *
    * If the base qualities are unknown (qual is null or equals "*"), the quality
    * scores will be a repeated string of 'B's that is equal to the read length.
    *
-   * @param adamRecord Read to convert to FASTQ.
+   * @param adamRecord Read to prepare for conversion to FASTQ and similar formats.
    * @param maybeAddSuffix If true, check if a "/%d" suffix is attached to the
    *   read. If there is no suffix, a slash and the number of the read in the
    *   sequenced fragment is appended to the readname. Default is false.
    * @param outputOriginalBaseQualities If true and the original base quality
    *   field is set (SAM "OQ" tag), outputs the original qualities. Else,
    *   output the qual field. Defaults to false.
-   * @return Returns this read in string form.
+   * @return Returns tuple of (name, sequence, qualityScores).
    */
-  def convertToFastq(
+  private def prepareFastq(
     adamRecord: AlignmentRecord,
-    maybeAddSuffix: Boolean = false,
-    outputOriginalBaseQualities: Boolean = false): String = {
+    maybeAddSuffix: Boolean,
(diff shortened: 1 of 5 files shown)


adam-core/src/main/scala/org/bdgenomics/adam/rdd/fragment/Tab6InFormatter.scala
```diff
@@ -0,0 +1,110 @@
...
+ * InFormatter companion that creates an InFormatter that writes Bowtie tab6 format.
+ */
+object Tab6InFormatter extends InFormatterCompanion[Fragment, FragmentRDD, Tab6InFormatter] {
+
+  /**
+   * Hadoop configuration path to check for a boolean value indicating whether
+   * the current or original read qualities should be written. True indicates
+   * to write the original qualities. The default is false.
+   */
+  val WRITE_ORIGINAL_QUALITIES = "org.bdgenomics.adam.rdd.fragment.Tab6InFormatter.writeOriginalQualities"
```
fnothaft: What say you to making these a constant in `org.bdgenomics.adam.rdd.fragment.FragmentRDD` or the like so that we don't repeat them across all the InFormatters?

adam-core/src/main/scala/org/bdgenomics/adam/converters/AlignmentRecordConverter.scala
```diff
@@ -30,31 +30,25 @@ import scala.collection.JavaConversions._
...
   /**
-   * Converts a single record to FASTQ. FASTQ format is:
-   *
-   * {{{
-   * @readName
-   * sequence
-   * +<optional readname>
-   * ASCII quality scores
-   * }}}
+   * Prepare a single record for conversion to FASTQ and similar formats.
```
fnothaft: I would expand what you mean by "prepare" here. Once I read through the method, I understood what you meant, but ideally I shouldn't need to read the source first.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/fragment/Tab6InFormatter.scala
```diff
@@ -0,0 +1,110 @@
...
+      if (reads.size > 1) {
+        val second = "\t" + converter.convertToTab6(reads(1),
+          maybeAddSuffix = writeSuffixes,
+          outputOriginalBaseQualities = writeOriginalQualities)
+
+        os.write(second.getBytes)
+      }
+
+      // end line
+      os.write("\n".getBytes)
```
fnothaft: I would compute the value of `"\n".getBytes` outside of the for loop.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/fragment/Tab6InFormatter.scala
heuermh: Yeah, that is what I meant by copy-pasta . . .  I would've put them in an abstract class that both InFormatters extended from but was not so sure about the companion stuff.

heuermh:
Yes, thank you.

fnothaft:
Thanks! Any chance you could squash the 1st and 3rd commits down? If not, I will merge manually later.

fnothaft:
Merged! Thanks @heuermh!

PR #1493, opened 2017-04-17
[ADAM-1483] Remove collapse parameter from AlignmentRecordRDD.toCoverage

### LINKED ISSUE #1483: Default value for reads.toCoverage(collapse) should be false
IMO, it is more common that you're going to run an analysis where you look at the number of reads at a site than where you're going to visualize coverage. As such, I think it makes more sense to not generate the collapsed coverage, which collapses sites that have identical coverage. @akmorrow13 your thoughts?

fnothaft:
Resolves #1483. Instead of providing a parameter in the `toCoverage` method to allow users to collapse the coverage, makes the `toCoverage` method stupid. This makes the `toCoverage` method simpler and the behavior easier to reason about.


heuermh:
Ping @akmorrow13 for review

fnothaft:
@heuermh Rebased.

heuermh:
Thank you, @fnothaft!

Issue #1219, opened 2016-10-19
Support for INSDC Sequence records (i.e., Genbank/EMBL format)?


laserson:
Any thoughts about supporting the INSDC data model as well?  This would be for the data in genbank/embl format.  This corresponds to the `SeqRecord` data type in biopython.  Apologies if this questions is totally redundant/annoying.

fnothaft:
I think that'd map well to @heuermh's https://github.com/bigdatagenomics/bdg-formats/pull/83, no? If so, SGTM!

laserson:
Yes, but it would extend a ways past it.  What @heuermh currently implemented is similar to Biopython's `Seq` object.  The `SeqRecord` object includes one of those, but also includes per-letter annotations (e.g., quality scores), dbxrefs, and a list of `Feature` objects.  Basically, the sequence and all the annotations on top of it.  Is there any appetite to develop that?

fnothaft:
Oh, that's kinda cool. I might say that we could add the per-base annotations and the dbxrefs, but I would leave out the Features, since we can reconstruct that from a join. @heuermh thoughts?

heuermh:
I was thinking `Sequence` and `Feature` as currently defined, and then a `SequenceAnnotation` or similar for the list of features and other stuff that goes in a Biopython `SeqRecord` or Biojava `RichSequence`.  Didn't the GA4GH recently propose something for this?

fnothaft:
Nice! I like that proposal.

heuermh:
@laserson What would you like to use as the specification/reference model for this?

I've written Biojava 1.x ([biojava-legacy](https://github.com/biojava/biojava-legacy)) to/from `Sequence/Slice/Read` conversions [here](https://github.com/heuermh/dishevelled-bio/tree/master/convert/src/main/java/org/dishevelled/bio/convert/biojava), similar could be done for Biojava 5.x on HEAD ([biojava](https://github.com/biojava/biojava)), and it wouldn't take much to use those libraries to support Genbank/EMBL formats in addition to FASTA and FASTQ.

Since Biojava is LGPL version 2.1, there is a question of where this code should live.  There is also a question of the proposal at https://github.com/bigdatagenomics/bdg-convert.

laserson:
I don't have experience with Biojava but probably a good place to start?

heuermh:
A good place to start, sure, but some of the code is 15 years old.  Is there a more up-to-date specification or reference?

There's this for the feature table
[...]

Issue #1173, opened 2016-09-15
Top level WrappedRDD or similar abstraction


heuermh:
Thinking about implementing phenotype support, or sequence/slice/read as proposed in bdg-formats, it appears that `GenomicRDD` is not quite the top level abstraction that it should be.

I propose a new top level trait/interface/abstract class `WrappedRDD` (or a better name) that includes only

``` scala
  val rdd: RDD[T]

  lazy val jrdd: JavaRDD[T] = {
    rdd.toJavaRDD()
  }

  def transform(tFn: RDD[T] => RDD[T]): U = {
    replaceRdd(tFn(rdd))
  }

  protected def replaceRdd(newRdd: RDD[T]): U
```

and perhaps some of the saveAsParquet stuff (not exactly sure, does that come in from `ADAMRDDFunctions`?) Then `GenomicRDD` adds the sequence dictionary, and so on.

This would give us something to extend from when a sequence dictionary is not required and a place to put code level documentation about the wrapped-RDD pattern that we've established.  It may also help address #1092.

fnothaft:
I don't think the wrapped RDD pattern makes sense _unless_ you have metadata that you want to package up with the RDD. That's the sole reason I wanted to move to the `GenomicRDD` pattern; we had lots of metadata we needed to access (`Sequence`/`RecordGroupDictionary`) that was expensive to aggregate and messy to pass around as tuples. Since then, there have been lots of other nice things that have dropped out of said pattern (being able to remove boilerplate around doing the region joins in the common case), but that all derives from having said `SequenceDictionary` metadata around. If you don't need the metadata wrapped up, then the wrapped RDD is cumbersome, since you need to wrap all RDD transformations in the `transform` function.

> and perhaps some of the saveAsParquet stuff (not exactly sure, does that come in from ADAMRDDFunctions?)

[...]

PR #1505, opened 2017-04-26
Adding read, sequence, and slice RDDs

heuermh:
Work in progress.


fnothaft:
Nice!

adam-core/src/main/scala/org/bdgenomics/adam/rdd/sequence/SliceRDD.scala
```diff
@@ -0,0 +1,232 @@
...
+                              outputPath: String,
+                              asSingleFile: Boolean) {
+    if (asSingleFile) {
+
+      // write rdd to disk
+      val tailPath = "%s_tail".format(outputPath)
+      rdd.saveAsTextFile(tailPath)
+
+      // get the filesystem impl
+      val fs = FileSystem.get(rdd.context.hadoopConfiguration)
```
fnothaft: You'll want to use:

```
val path = new Path(outputPath)
val fs = path.getFileSystem(rdd.context.hadoopConfiguration)
```

Otherwise, this will fail if the path is not in the file system that is the `fs.defaultFs` from the Configuration. `Path` here is `org.apache.hadoop.fs.Path`.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/sequence/ReadRDD.scala
```diff
@@ -0,0 +1,198 @@
...
+                              outputPath: String,
+                              asSingleFile: Boolean) {
+    if (asSingleFile) {
+
+      // write rdd to disk
+      val tailPath = "%s_tail".format(outputPath)
+      rdd.saveAsTextFile(tailPath)
+
+      // get the filesystem impl
+      val fs = FileSystem.get(rdd.context.hadoopConfiguration)
```
fnothaft: See comment below about getting the `fs`

adam-core/src/main/scala/org/bdgenomics/adam/rdd/sequence/SequenceRDD.scala
```diff
@@ -0,0 +1,195 @@
...
+                              outputPath: String,
+                              asSingleFile: Boolean) {
+    if (asSingleFile) {
+
+      // write rdd to disk
+      val tailPath = "%s_tail".format(outputPath)
+      rdd.saveAsTextFile(tailPath)
+
+      // get the filesystem impl
+      val fs = FileSystem.get(rdd.context.hadoopConfiguration)
```
fnothaft: See comment below about getting the `fs`

adam-core/src/main/scala/org/bdgenomics/adam/rdd/sequence/ReadRDD.scala
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  /**
+   * Writes an RDD to disk as text and optionally merges.
+   *
+   * @param rdd RDD to save.
+   * @param outputPath Output path to save text files to.
+   * @param asSingleFile If true, combines all partition shards.
+   */
+  private def writeTextRdd[T](rdd: RDD[T],
```
fnothaft: As an aside, this seems like it is used in a lot of places. I think we have this in `FragmentRDD` too? Perhaps we can move it into `GenomicRDD` as a protected method?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/sequence/SequenceRDD.scala
```diff
@@ -0,0 +1,195 @@
...
+   * @param lineWidth Hard wrap FASTA formatted sequence at line width, default 60.
+   */
+  def saveAsFasta(filePath: String, asSingleFile: Boolean = false, lineWidth: Int = 60) {
+
+    def toFasta(sequence: Sequence): String = {
+      val sb = new StringBuilder()
+      sb.append(">")
+      sb.append(sequence.getName)
+      Option(sequence.getDescription).foreach(n => sb.append(" ").append(n))
+      for (line <- Splitter.fixedLength(lineWidth).split(sequence.getSequence)) {
```
fnothaft: I would prefer `sequence.getSequence.grouped(lineWidth).foreach(line => {` instead.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/sequence/SliceRDD.scala
```diff
@@ -0,0 +1,232 @@
...
+   * to Parquet. Defaults to 60 character line length, if saving to FASTA.
+   *
+   * @param filePath Path to save files to.
+   * @param asSingleFile If true, saves output as a single file.
+   */
+  def save(filePath: java.lang.String, asSingleFile: java.lang.Boolean) {
+    if (filePath.endsWith(".fa") || filePath.endsWith(".fasta")) {
+      saveAsFasta(filePath, asSingleFile = asSingleFile)
+    } else {
+      if (asSingleFile) {
```
fnothaft: A bit silly, but perhaps we can move this to `GenomicRDD` as a protected method `def warnIfSingleFile(asSingleFile: Boolean): Unit`?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/sequence/SliceRDD.scala
```diff
@@ -0,0 +1,232 @@
...
+   */
+  def saveAsFasta(filePath: String, asSingleFile: Boolean = false, lineWidth: Int = 60) {
+
+    def toFasta(slice: Slice): String = {
+      val sb = new StringBuilder()
+      sb.append(">")
+      sb.append(slice.getName)
+      Option(slice.getDescription).foreach(n => sb.append(" ").append(n))
+      sb.append(s" slice.getStart-slice.getEnd:slice.getStrand")
+      for (line <- Splitter.fixedLength(lineWidth).split(slice.getSequence)) {
```
fnothaft: See comment WRT `grouped` above.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/sequence/SliceRDD.scala
```diff
@@ -0,0 +1,232 @@
...
+   * @param lineWidth Hard wrap FASTA formatted slice at line width, default 60.
+   */
+  def saveAsFasta(filePath: String, asSingleFile: Boolean = false, lineWidth: Int = 60) {
+
+    def toFasta(slice: Slice): String = {
+      val sb = new StringBuilder()
+      sb.append(">")
+      sb.append(slice.getName)
+      Option(slice.getDescription).foreach(n => sb.append(" ").append(n))
+      sb.append(s" slice.getStart-slice.getEnd:slice.getStrand")
```
fnothaft: OOC, does this syntax come from somewhere? If so, I think it'd be good to document it inline/in the Scaladoc.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/sequence/SliceRDD.scala
```diff
@@ -0,0 +1,232 @@
...
+   */
+  def merge(): SequenceRDD = {
+    def toSequence(slice: Slice): Sequence = {
+      Sequence.newBuilder()
+        .setName(slice.getName)
+        .setDescription(slice.getDescription)
+        .setAlphabet(slice.getAlphabet)
+        .setSequence(slice.getSequence)
+        .setLength(slice.getLength)
+        //.setAttributes(slice.getAttributes)
```
fnothaft: Why's this commented?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/sequence/SliceRDD.scala
```diff
@@ -0,0 +1,232 @@
...
+      Sequence.newBuilder(first)
+        .setLength(first.getLength + second.getLength)
+        .setSequence(first.getSequence + second.getSequence)
+        .build
+    }
+
+    val merged: RDD[Sequence] = rdd
+      .sortBy(slice => (slice.getName, slice.getStart))
+      .map(slice => (slice.getName, toSequence(slice)))
+      .reduceByKey(mergeSequences)
```
fnothaft: This won't respect the ordering guarantee you need.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/sequence/SliceRDD.scala
heuermh: That should be cleaned up as part of other `save` related refactoring.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/sequence/ReadRDD.scala
heuermh: Yes, that was mentioned elsewhere

adam-core/src/main/scala/org/bdgenomics/adam/rdd/sequence/SliceRDD.scala
heuermh: I suppose the attributes should be merged?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/sequence/SliceRDD.scala
heuermh: Is this different than that in [`NucleotideContigFragmentRDD`](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/contig/NucleotideContigFragmentRDD.scala#L208)?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/sequence/SliceRDD.scala
heuermh: Ah wait, this requires bdg-formats 0.10.2-SNAPSHOT to pick up https://github.com/bigdatagenomics/bdg-formats/commit/b0e7bb9fa67c127f91b1ad3130719d5ec4d24bd7 and https://github.com/bigdatagenomics/bdg-formats/commit/ddbc8b7708106737fec76968c5375438834742c8

adam-core/src/main/scala/org/bdgenomics/adam/rdd/sequence/SliceRDD.scala
fnothaft: It isn't, but that code is wrong.

heuermh:
I seem to have git-f'ed this one up.  No matter, it won't build until #1523 is merged.

heuermh:
Merged and rebased; depends on #1523.

heuermh:
> There's currently no entry point to these schemas from ADAMContext, or through any of the extant schemas.

The current entry points are https://github.com/heuermh/biojava-adam and https://github.com/heuermh/biojava-legacy-adam, which can't be included in our repositories since they are LPGL.

> Sequence and Slice are duplicative of NucleotideContigFragment. How do we plan to rectify this long term?

I was thinking to include these in 0.23.0 and then replace NCF with Sequence and Slice in 0.24.0, for reading genome assembly sequences.  Sequence also supports the use case for generic sequence records, RefSeq, SwissProt, etc.

The argument for `Read` is weaker, in that conceptually one can have reads and fragments without having run any alignment program, say if the pipeline was based solely on assembly.  Shoving those into `AlignmentRecord` seems not quite right.

That said, GA4GH has taken a similar approach, in that fragment, read, and alignments are all squashed into a single record, `ReadAlignment` https://ga4gh-schemas.readthedocs.io/en/documentation/api/reads.html#readalignment-detailed-discussion.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/sequence/ReadRDD.scala
```diff
@@ -0,0 +1,226 @@
...
+  /**
+   * Builds a ReadRDD, creating the sequence dictionary
+   * from read names and lengths.
+   *
+   * @param rdd The underlying Read RDD to build from.
+   * @return Returns a new ReadRDD.
+   */
+  def apply(rdd: RDD[Read]): ReadRDD = {
+
+    // get sequence dictionary
```
heuermh: Yeah this sucks, and is only there to support the contract of `GenomicRDD`.

We could either come up with a better fake sequence dictionary or have this not extend `GenomicRDD` (see #1173).

heuermh: Talking to @devin-petersohn today, it sounds like the `SequenceDictionary` isn't necessary for joins, as I thought it was.  Thus I'll take this out and leave the sequence dictionary empty.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/sequence/SequenceRDD.scala
```diff
@@ -0,0 +1,330 @@
...
+
+  /**
+   * Slice the sequences in this RDD overlapping the specified region.
+   *
+   * @param region Region to overlap.
+   * @return Returns a new SliceRDD from the sequences in this RDD sliced
+   *    to overlap the specified region.
+   */
+  def slice(region: ReferenceRegion): SliceRDD = {
+    SliceRDD(filterByOverlappingRegion(region).rdd.map(sequence => slice(sequence, region)))
```
heuermh: `filterByOverlappingRegion` calls &rarr; `replaceRdd` calls &rarr; `copy`
or perhaps I'm missing what you're saying?

heuermh:
In case it is not clear from comment https://github.com/bigdatagenomics/adam/pull/1505#issuecomment-311676371 above, I would like to

 - Merge this now and release with version 0.23.0, in its currently detached-from-`ADAMContext` state
 - With ADAM version 0.23.0 as a dependency, the biojava-adam and biojava-legacy-adam projects could be merged into biojava and biojava-legacy proper, or kept as separate repos and migrated to the biojava organization

Then for version 0.24.0:
 - Add `Read` to `AlignmentRecord` conversions
 - Consider renaming `AlignmentRecord` &rarr; `ReadAlignment`
 - Replace `NucleotideContigFragment` with `Sequence` and `Slice` as appropriate
 - Consider renaming or generalizing `Fragment`
 - Consider renaming `Contig` &rarr; `Reference`; `contigName` &rarr; `referenceName`
 - Consider renaming `SequenceDictionary` &rarr; `ReferenceSet` or similar, and replace `SequenceRecord` with former-`Contig`-now-`Reference`

The issues for these are scattered around a bit, I didn't link here because they don't really need to be associated with this pull request.

fnothaft:
Hi @heuermh, sorry for not replying to the comment earlier. I still really don't grok this PR and am pretty hesitant to add it to 0.23.0 since I don't see it on the critical path for any deliverables. From a cursory read, I'm not really a fan of the changes you propose for 0.24.0.  I don't think any of them are bad, but they're pretty disruptive changes in terms of LOC touched with fairly negligible benefits, and at a minimum I see them as dramatically lower in priority than finishing the Genotype refactor, as well as the other issues we've currently got slated for 0.24.0.

As an aside, do you have buy-in from the rest of the biojava folks for:

> * With ADAM version 0.23.0 as a dependency, the biojava-adam and biojava-legacy-adam projects could be merged into biojava and biojava-legacy proper, or kept as separate repos and migrated to the biojava organization

heuermh:
The main driver issue is #1219.  The changes I propose above for 0.24.0 round out the models we have in formats and match pretty well with those in the OBF projects, with the exception that we keep `Sequence` and `Feature` separate, instead of `Sequence`s having `Feature`s.

A lot of bioinformatics happens on `Sequence`s that are not `NucleotideContigFragment`s.

There is interest in biojava-adam and biojava-legacy-adam from the biojava folks and also from Peter Rose separately.

fnothaft:
First, I'm philosophically opposed to merging this until it connects up to ADAMContext and real world data. E.g., how can we validate that this resolves #1219 without actually pushing INSDC records through it? I don't disagree that real world analysis happens on non-Nucleotide data, but wouldn't the simple and straightforward way to address this be to rename NCF to `Sequence` or something of the like, instead of adding 1.5KLOC that will need to be separately validated? As is, the new `Sequence` and `Slice` classes aren't at feature parity with NCF.

WRT Biojava, is there a thread where the merger of ADAM integration into biojava is discussed? If this is a direction we're going, I'd like to discuss the actual brass tacks of how the integration plays out.

heuermh:
I'll save this until post-0.23.0 then, and do the NCF refactor simultaneously, which will require a new bdg-formats release.

I don't propose us writing our own INSDC parser, so the resolution to #1219 will happen outside of the ADAM codebase.

For the biojava repos on my account, when their snapshot dependencies are resolved, I'll pull request them as new Maven modules against biojava and biojava-legacy.  The pr against biojava will most likely fail due to resistance to adding new dependencies there.  I'm pretty much the only one maintaining biojava-legacy.  If the prs don't take, then I'd propose migrating the repos to the biojava organization.

antonkulaga:
This PR looks very interesting, I like the new classes and I hope it will increase the speed of file conversions

heuermh:
Closing in favor of #1895.

PR #1508, opened 2017-04-29
[ADAM-1502] Preserve contig ordering in TwoBitFile sequence dictionary.

### LINKED ISSUE #1502: Sequence dictionary records in TwoBitFile are not stable
The order of the contigs in a dictionary produced from TwoBitFile is not the same as the order they are stored in in the .2bit file.

I think the problem is the records are being stored in a standard map, rather than one that preserves insertion order (like ListMap).

See https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/util/TwoBitFile.scala#L70

fnothaft:
Resolves #1502.

@tomwhite this should resolve the issue you were seeing. Can you test this out on your end?


tomwhite:
I haven't got a straightforward way of testing this at the moment, but this looks good to me.

Thanks for the fix @fnothaft.

heuermh:
Thank you, @fnothaft!

PR #1515, opened 2017-05-05
Making TreeRegionJoin consistent with ShuffleRegionJoin

### LINKED ISSUE #1514: Minor refactor for TreeRegionJoin for consistency
The way that each `TreeRegionJoin` case class extends `RegionJoin` should be consistent with `ShuffleRegionJoin`. The parent trait should implement `RegionJoin`.

devin-petersohn:
Resolves #1514


devin-petersohn:
This is just something I found along the way. I split it out of #1324 because this has a chance to potentially break some things.

fnothaft:
Merged! Thanks @devin-petersohn!

devin-petersohn:
Thanks @fnothaft!

PR #1518, opened 2017-05-09
[ADAM-1517] Move to Parquet 1.8.2 in preparation for moving to Spark 2.2.0

### LINKED ISSUE #1517: Move to Spark 2.2, Parquet 1.8.2
Requires shading Avro.

fnothaft:
Bumps Parquet version to 1.8.2. Relocates `org.apache.avro` to `org.bdgenomics.avro`. Resolves #1517 pending release of and bump to Spark 2.2.0.

### DIFF
--- a/adam-apis/pom.xml
+++ b/adam-apis/pom.xml
@@ -91,6 +91,18 @@
 
   <dependencies>
     <dependency>
+      <groupId>org.apache.parquet</groupId>
+      <artifactId>parquet-avro</artifactId>
+      <version>1.8.1</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.parquet</groupId>
+      <artifactId>parquet-scala_2.10</artifactId>
+      <version>1.8.1</version>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
       <groupId>org.scala-lang</groupId>
       <artifactId>scala-library</artifactId>
       <scope>compile</scope>
--- a/adam-assembly/pom.xml
+++ b/adam-assembly/pom.xml
@@ -71,6 +71,10 @@
                   <pattern>org.apache.avro</pattern>
                   <shadedPattern>org.bdgenomics.avro</shadedPattern>
                 </relocation>
+                <relocation>
+                  <pattern>org.apache.parquet</pattern>
+                  <shadedPattern>org.bdgenomics.parquet</shadedPattern>
+                </relocation>
               </relocations>
             </configuration>
           </execution>
@@ -117,6 +121,16 @@
     </plugins>
   </build>
   <dependencies>
+    <dependency>
+      <groupId>org.apache.parquet</groupId>
+      <artifactId>parquet-avro</artifactId>
+      <scope>compile</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.parquet</groupId>
+      <artifactId>parquet-scala_2.10</artifactId>
+      <scope>compile</scope>
+    </dependency>
     <dependency>
       <groupId>org.bdgenomics.adam</groupId>
       <artifactId>adam-cli_${scala.version.prefix}</artifactId>
--- a/adam-assembly/pom.xml
+++ b/adam-assembly/pom.xml
@@ -66,6 +66,12 @@
               <transformers>
                 <transformer implementation="org.apache.maven.plugins.shade.resource.ServicesResourceTransformer"/>
               </transformers>
+              <relocations>
+                <relocation>
+                  <pattern>org.apache.avro</pattern>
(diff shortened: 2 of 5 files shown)


heuermh:
From Spark 1.6.1 Jenkins test failure:
```
java.lang.NoSuchMethodError: org.apache.parquet.io.api.Binary.fromCharSequence(Ljava/lang/CharSequence;)Lorg/apache/parquet/io/api/Binary;
	at org.apache.parquet.avro.AvroWriteSupport.fromAvroString(AvroWriteSupport.java:367)
	at org.apache.parquet.avro.AvroWriteSupport.writeValueWithoutConversion(AvroWriteSupport.java:342)
	at org.apache.parquet.avro.AvroWriteSupport.writeValue(AvroWriteSupport.java:274)
	at org.apache.parquet.avro.AvroWriteSupport.writeRecordFields(AvroWriteSupport.java:187)
	at org.apache.parquet.avro.AvroWriteSupport.write(AvroWriteSupport.java:161)
	at org.apache.parquet.hadoop.InternalParquetRecordWriter.write(InternalParquetRecordWriter.java:121)
	at org.apache.parquet.hadoop.ParquetRecordWriter.write(ParquetRecordWriter.java:123)
	at org.apache.parquet.hadoop.ParquetRecordWriter.write(ParquetRecordWriter.java:42)
	at org.apache.spark.rdd.InstrumentedRecordWriter$$anonfun$write$1.apply$mcV$sp(InstrumentedOutputFormat.scala:55)
```

That `InstrumentedRecordWriter` is ours from bdg-utils, right?

fnothaft:
@heuermh I believe this error indicates that moving to Parquet 1.8.2 breaks backwards compatibility with Spark 1.x.

heuermh:
The bad things are to depend on 1.8.2 and shade 1.8.1 to include at runtime?  Yeah that sucks.  :)

fnothaft:
Yeah, I'm not proud of what I've done. It's a real mess; there's a lot going on. Here's the big four:

* spark-sql depends on parquet-column and whatnot, but doesn't depend on parquet-avro, but spark-core also depends separately on avro. Since they depend on avro but don't depend on parquet-avro, there's effectively a transitive dependency conflict that exists, but Spark doesn't notice it. spark-sql does include parquet-avro in test scope, but then pins parquet-avro to 1.8.1 even after parquet-column moved to 1.8.2 (iirc, I'd need to double check this).
* parquet-avro kinda broke binary compat going 1.8.1 to 1.8.2. I feel a bit weird calling it "breaking binary compat" because parquet-avro 1.8.2 relies on a new public method in parquet-column 1.8.2, and thus cannot be linked to parquet-column 1.8.1. Also, it is an interface that is logically wholly contained inside of parquet (see below). *shrug* I don't fault the parquet folks—this is fairly reasonable behavior—but there's a dependency management story between Parquet and Spark and Avro that people haven't really thought through.
* We've depended on a higher version of Avro for a while. The one thing this seems to break is persisting the sorted partition map to disk; otherwise this seems to work OK. I'm not 100% sure how we're getting away with this, but, we have been.
* Arguably, the least bad solution to all of the above problems is to relocate both Parquet and Avro in a shaded JAR before building adam-core; then we don't have the dependency version mismatches in test scope (this is bad, right?). Again, this is a pretty bad "least bad" solution, but I digress. However, if you try to do this, you need to relocate both parquet-avro (and it's transitive dependencies) and parquet-scala. However, relocating parquet-scala appears to break the org.apache.parquet.filter2.dsl.Dsl companion object link, which means that you can't create predicates.

WRT the last point, we wind up relocating parquet in the shaded überjar, which is probably a bad thing for people who are using just adam-core, or alternatively, a bad thing for people who are using the adam-assembly überjar.

Here's the parquet binary compat issue. Parquet 1.8.2 relies on a method that's not present in Parquet 1.8.1:

```
java.lang.NoSuchMethodError: org.apache.parquet.io.api.Binary.fromCharSequence(Ljava/lang/CharSequence;)Lorg/apache/parquet/io/api/Binary;&#010; at org.apache.parquet.avro.AvroWriteSupport.fromAvroString(AvroWriteSupport.java:367)&#010; at org.apache.parquet.avro.AvroWriteSupport.writeValueWithoutConversion(AvroWriteSupport.java:342)&#010; at org.apache.parquet.avro.AvroWriteSupport.writeValue(AvroWriteSupport.java:274)&#010; at org.apache.parquet.avro.AvroWriteSupport.writeRecordFields(AvroWriteSupport.java:187)&#010; at org.apache.parquet.avro.AvroWriteSupport.write(AvroWriteSupport.java:161)&#010; at org.apache.parquet.hadoop.InternalParquetRecordWriter.write(InternalParquetRecordWriter.java:121)&#010; at 
```

heuermh:
Closing in favor of #1722

PR #1525, opened 2017-05-12
[ADAM-1486] Respect validation stringency if BAM header load fails.

### LINKED ISSUE #1486: Should throw error when BAM header load fails
Right now, [the code that loads the BAM file header logs an error when loading a header fails](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala#L573-L576), but it isn't clear to me that this is the correct behavior. Thoughts @heuermh @ryan-williams?

fnothaft:
Resolves #1486.


fnothaft:
Helps to include the new test file...

heuermh:
Thank you, @fnothaft!

PR #1531, opened 2017-05-17
[ADAM-1530] Support loading GO:query (S/CR/B)AMs as fragments.

### LINKED ISSUE #1530: Queryname sorted check should check for queryname grouped as well
See https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala#L506-L534. Our goal is to check whether files have paired reads are grouped together, which is satisfied by sort order = queryname or group order = queryname.

fnothaft:
Resolves #1530. Adds support for loading (S/CR/B)AM files that are globally unsorted yet have group order set to query (read name) as Fragments. Additionally, adds two unit tests. In the process of running the unit tests, I found an issue in `SingleReadBucket` that was dropping a read from each pair after the first pair in a partition due to an odd behavior in the `Iterator.takeWhile` function, which drops the first item that does not get taken.


heuermh:
Thank you, @fnothaft

PR #1532, opened 2017-05-17
Refactor Vcf2ADAM and ADAM2Vcf into TransformGenotypes and TransformVariants

heuermh:
Work in progress.
- [x] Update docs
- [x] Failing unit tests due to FT Number=. vs. FT Number=1
- [x] Test on cluster


adam-cli/src/main/scala/org/bdgenomics/adam/cli/TransformGenotypes.scala
```diff
@@ -0,0 +1,146 @@
...
+      "Cannot set both -sort_on_save and -sort_lexicographically_on_save.")
+
+    val genotypes = sc.loadGenotypes(
+      args.inputPath,
+      optPredicate = None,
+      optProjection = None,
+      stringency = stringency)
+
+    // convert to variant contexts
+    val variantContexts = genotypes.toVariantContextRDD
```
fnothaft: Should only convert to variant contexts if saving as VCF. The sort and transform methods are all defined on `GenomicRDD` so just make the `maybeCoalesce` functions take `[T: GenomicRDD]` or something like that.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/variant/GenotypeRDD.scala
```diff
@@ -159,17 +162,39 @@ case class GenotypeRDD(rdd: RDD[Genotype],
   /**
    * Explicitly saves to VCF.
    *
+   * @param args Arguments defining where to save the file.
+   * @param stringency The validation stringency to use when writing the VCF.
+   *   Defaults to LENIENT.
+   */
+  def saveAsVcf(args: ADAMSaveAnyArgs,
```
fnothaft: Thoughts on removing this method?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/variant/GenotypeRDD.scala
```diff
@@ -159,17 +162,39 @@ case class GenotypeRDD(rdd: RDD[Genotype],
...
+   *   output shards as a headerless file, but we will not merge the shards.
+   * @param disableFastConcat If asSingleFile is true and deferMerging is false,
+   *   disables the use of the parallel file merging engine.
    * @param stringency The validation stringency to use when writing the VCF.
    */
   def saveAsVcf(filePath: String,
                 asSingleFile: Boolean,
-                stringency: ValidationStringency) {
-    toVariantContextRDD.saveAsVcf(filePath, asSingleFile, stringency)
+                deferMerging: Boolean,
```
fnothaft: Thoughts on removing this method as well?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/variant/VariantRDD.scala
```diff
@@ -112,12 +113,56 @@ case class VariantRDD(rdd: RDD[Variant],
...
+
+  /**
+   * Explicitly saves to VCF.
+   *
+   * @param args Arguments defining where to save the file.
+   * @param stringency The validation stringency to use when writing the VCF.
+   *   Defaults to LENIENT.
+   */
+  def saveAsVcf(args: ADAMSaveAnyArgs,
+                stringency: ValidationStringency = ValidationStringency.LENIENT): Unit = {
```
fnothaft: I'd prefer to not add this method.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/variant/VariantRDD.scala
```diff
@@ -128,12 +173,22 @@ case class VariantRDD(rdd: RDD[Variant],
...
+   *   output shards as a headerless file, but we will not merge the shards.
+   * @param disableFastConcat If asSingleFile is true and deferMerging is false,
+   *   disables the use of the parallel file merging engine.
    * @param stringency The validation stringency to use when writing the VCF.
    */
   def saveAsVcf(filePath: String,
                 asSingleFile: Boolean,
-                stringency: ValidationStringency) {
-    toVariantContextRDD.saveAsVcf(filePath, asSingleFile, stringency)
+                deferMerging: Boolean,
```
fnothaft: I'd prefer to remove this method.

adam-core/src/test/resources/sorted-variants.vcf
```diff
@@ -0,0 +1,74 @@
+##fileformat=VCFv4.2
```
fnothaft: We already have a `adam-core/src/test/resources/sorted.vcf`. Why is this file needed?

adam-cli/src/main/scala/org/bdgenomics/adam/cli/TransformGenotypes.scala
heuermh: That is the way it was implemented before the latest commit and failed unit tests.  Sorting `GenotypeRDD` didn't work as expected.

adam-core/src/test/resources/sorted-variants.vcf
heuermh: `sorted-variants.vcf` is the site-only `-only_variants` version for `TransformVariants`

adam-core/src/main/scala/org/bdgenomics/adam/rdd/variant/GenotypeRDD.scala
heuermh: I wouldn't have added it if I thought it should be removed ;)

As mentioned elsewhere, save methods are a bit of a mess and need to be made consistent across subclasses of `GenomicRDD`.  I'm ok if this PR makes it a little worse before it gets better.

adam-core/src/test/resources/sorted-variants.vcf
fnothaft: Ah, I see! Nevermind.

adam-cli/src/main/scala/org/bdgenomics/adam/cli/TransformGenotypes.scala
fnothaft: Yeah, sorting `GenotypeRDD` and then converting to `VariantContextRDD` currently does a groupBy without a defined partitioner, which causes a hash-partitioned shuffle, and thus un-sorts the `VariantContextRDD`. Yes, this code does fix that, but is incredibly inefficient if you're going VCF to parquet in which case you flatMap `VariantContext` to `Genotype`, then groupBy on `Genotype`s to reconstruct `VariantContext`s, which you then flatMap back to Parquet. Doesn't make sense to me.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/variant/GenotypeRDD.scala
fnothaft: I think this is inconsistent with the direction we've been going, where we've broken up `GenomicRDD` methods that transiently change the underlying type of the `GenomicRDD` (e.g., `toCoverage(collapse = true)` becoming `toCoverage().collapse()`. I think that this method is particularly problematic because the changes to the method actually break the utility of the method. Specifically, this change means that the `.save` method always saves unsorted VCF, which is non-compliant with the VCF spec.

heuermh: Isn't the fix here and above to address the "groupBy without a defined partitioner, which causes a hash-partitioned shuffle, and thus un-sorts the `VariantContextRDD`."?

fnothaft: I believe it is a bit more complex than that, and would be best to handle after #1324 merges (since you lose partitioning on "externally sorted" data).

heuermh: Hmm, yeah I do see that.

So you're suggesting no format guessing `save` methods for `GenotypeRDD`, `VariantRDD`, and `VariantContextRDD`.  `GenotypeRDD` and `VariantRDD` have only `saveAsParquet` and `VariantContextRDD` has only `saveAsVcf`.  The format guessing happens in the CLI classes.  I could support that.

fnothaft: Yup! I +1 that approach.

heuermh: ok, what are the scala syntax sprinkles necessary for `maybeSort` to operate on `GenotypeRDD` and return the specific subclass?

fnothaft:
@heuermh Ping me when this is ready for review.

heuermh:
The commit history looks a mess, otherwise I think it is ready to review.

adam-cli/src/main/scala/org/bdgenomics/adam/cli/TransformGenotypes.scala
```diff
@@ -0,0 +1,132 @@
...
+    require(!(args.sort && args.sortLexicographically),
+      "Cannot set both -sort_on_save and -sort_lexicographically_on_save.")
+
+    val genotypes = sc.loadGenotypes(
+      args.inputPath,
+      optPredicate = None,
+      optProjection = None,
+      stringency = stringency)
+
+    if (args.outputPath.endsWith(".vcf")) {
```
fnothaft: +1! I like this refactor. One thing here though is that we lost the check for compressed formats (e.g., ".vcf.gz").

adam-cli/src/main/scala/org/bdgenomics/adam/cli/TransformVariants.scala
```diff
@@ -0,0 +1,132 @@
...
+    require(!(args.sort && args.sortLexicographically),
+      "Cannot set both -sort_on_save and -sort_lexicographically_on_save.")
+
+    val variants = sc.loadVariants(
+      args.inputPath,
+      optPredicate = None,
+      optProjection = None,
+      stringency = stringency)
+
+    if (args.outputPath.endsWith(".vcf")) {
```
fnothaft: Ditto here RE: compression.

adam-core/src/main/scala/org/bdgenomics/adam/converters/DefaultHeaderLines.scala
```diff
@@ -150,7 +150,7 @@ object DefaultHeaderLines {
     VCFHeaderLineType.Float,
     "Read-backed phasing quality")
   lazy val genotypeFilter = new VCFFormatHeaderLine("FT",
-    1,
+    VCFHeaderLineCount.UNBOUNDED,
```
fnothaft: Can you keep this split out into a different commit so that we have a single commit per issue? No sweat if you've already squashed.

heuermh: Will do

adam-cli/src/main/scala/org/bdgenomics/adam/cli/TransformGenotypes.scala
heuermh: There wasn't a check for compressed formats before, and I don't think we support writing compressed VCF through HadoopBAM.  With a code change to use `FileExtensions.isVcfExt`
```bash
$ ./bin/adam-submit transformVariants \
  -single \
  adam-core/src/test/resources/small.vcf \
  small.vcf.gz

$ gzcat small.vcf.gz 
gzcat: small.vcf.gz: not in gzip format

$ head small.vcf.gz 
##fileformat=VCFv4.2
##FILTER=<ID=IndelFS,Description="FS > 200.0">
##FILTER=<ID=IndelQD,Description="QD < 2.0">
...
```
Do we have to set a configuration parameter or otherwise kick HadoopBAM to write .gz or .bgz/.bgzf?

fnothaft: Ah, sorry about that, you are correct that we haven't been checking that for output.

That said, I read through the source and Hadoop-BAM writes BGZIP if you set the `FileOutputFormat.COMPRESS` configuration value to `true`.

heuermh: How about leaving this pr as-is then, squashed to two commits to keep commit c1b7375 separate, and then a new pr to add support for writing .gz (as BGZIP VCF) and .bgz/.bgzf (also as BGZIP VCF, I believe the exception error message above is incorrect)?

fnothaft:
Merged! Thanks @heuermh!

PR #1541, opened 2017-05-20
Add a simple unit test to SingleFastqInputFormat

kkaneda:
Hi,

I was randomly reading the code and decided to add a simple test. I didn't go further since I wasn't sure this is something actually useful. 

Could you please take a look? Thanks!

### DIFF
--- a/adam-core/pom.xml
+++ b/adam-core/pom.xml
@@ -201,5 +201,10 @@
       <artifactId>guava</artifactId>
       <scope>compile</scope>
     </dependency>
+    <dependency>
+      <groupId>junit</groupId>
+      <artifactId>junit</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 </project>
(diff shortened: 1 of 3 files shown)


heuermh:
Jenkins, test this please

heuermh:
Even though `SingleFastqInputFormat` is written in java, the rest of our unit test code is in scala.  Would it be possible to port this to scala using ScalaTest?  We use [the FunSuite style](http://www.scalatest.org/user_guide/selecting_a_style), which reads very similar to JUnit.

The source file will also need a license header, which is located in the repo at https://github.com/bigdatagenomics/adam/blob/master/LICENSE_header.txt.

kkaneda:
Thanks for taking a look, @heuermh ! Ah, didn't notice that. 

So, we have `SingleFastqInputFormatSuite`, and it seems that has sufficient coverage. I'll decline the PR.

PR #1547, opened 2017-05-24
[ADAM-941] Automatically generate projection enums. 

### LINKED ISSUE #941: Autogen field names in projection package
@jpdna has identified a bug whose root cause has to do with a default field being used when a field is projected out using Parquet's projection feature and our `org.bdgenomics.adam.projection` package. The TL;DR on the bug (@jpdna can you open an issue for this and link the issue to this issue?) is that the `firstOfPair`/`secondOfPair` fields changed to `readInFragment` but the [AlignmentRecordField](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/projections/AlignmentRecordField.scala) enum didn't get changed, so the projection didn't get fixed. To not have this problem come up in the future, we should autogenerate the enums from the avro schemas from bdg-formats.

fnothaft:
Resolves #941. Depends on #1391.


heuermh:
This is pretty sweet :)

fnothaft:
Rebased.

adam-codegen/src/main/scala/org/bdgenomics/adam/codegen/DumpSchemasToProjectionEnums.scala
```diff
@@ -0,0 +1,110 @@
...
+    val classNameNoPackage = className.split('.').last
+
+    "\n\nobject %sField extends FieldEnumeration(%s.SCHEMA$) {\n  val %s = SchemaValue\n}".format(
+      classNameNoPackage,
+      className,
+      fields(schema).mkString(", "))
+  }
+
+  private def writeHeader(fw: FileWriter, packageName: String) {
+    val hdr = Seq(
```
devin-petersohn: Can this header be factored out now to a separate package object?

fnothaft: Yeah, that'd be reasonable. Let me do that real quick.

fnothaft:
Rebased and factored out header writer. Good to go from my side.

heuermh:
Thank you, @fnothaft

PR #1555, opened 2017-06-01
[ADAM-1513] Strandedness for FeatureRDDs

devin-petersohn:
This was the lightest change that added the most flexibility for the future.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
@@ -478,7 +478,9 @@ trait GenomicRDD[T, U <: GenomicRDD[T, U]] extends Logging {
     newRdd: RDD[T],
     newPartitionMap: Option[Array[Option[(ReferenceRegion, ReferenceRegion)]]] = None): U
 
-  protected def getReferenceRegions(elem: T): Seq[ReferenceRegion]
+  protected def getReferenceRegions(
+    elem: T,
+    stranded: Boolean = false): Seq[ReferenceRegion]
 
   protected def flattenRddByRegions(): RDD[(ReferenceRegion, T)] = {
     rdd.flatMap(elem => {
@@ -692,7 +694,10 @@ trait GenomicRDD[T, U <: GenomicRDD[T, U]] extends Logging {
       buildTree(flattenRddByRegions()),
       genomicRdd.flattenRddByRegions()),
       sequences ++ genomicRdd.sequences,
-      kv => { (kv._1.flatMap(getReferenceRegions) ++ genomicRdd.getReferenceRegions(kv._2)).toSeq })
+      kv => {
+        (kv._1.flatMap(getReferenceRegions(_)) ++
+          genomicRdd.getReferenceRegions(kv._2)).toSeq
+      })
       .asInstanceOf[GenomicRDD[(Iterable[T], X), Z]]
   }
 
@@ -1185,7 +1190,8 @@ private case class GenericGenomicRDD[T](
     IntervalArray(rdd)
   }
 
-  protected def getReferenceRegions(elem: T): Seq[ReferenceRegion] = {
+  protected def getReferenceRegions(elem: T,
+                                    stranded: Boolean = false): Seq[ReferenceRegion] = {
     regionFn(elem)
   }
 
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
@@ -482,9 +482,9 @@ trait GenomicRDD[T, U <: GenomicRDD[T, U]] extends Logging {
     elem: T,
     stranded: Boolean = false): Seq[ReferenceRegion]
 
-  protected def flattenRddByRegions(): RDD[(ReferenceRegion, T)] = {
+  protected def flattenRddByRegions(stranded: Boolean = false): RDD[(ReferenceRegion, T)] = {
     rdd.flatMap(elem => {
-      getReferenceRegions(elem).map(r => (r, elem))
+      getReferenceRegions(elem, stranded).map(r => (r, elem))
     })
   }
 
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/contig/NucleotideContigFragmentRDD.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/contig/NucleotideContigFragmentRDD.scala
@@ -166,8 +166,6 @@ case class NucleotideContigFragmentRDD(
   }
 
   /**
-   * Returns the regions that this fragment covers.
-   *
    * Save nucleotide contig fragments as Parquet or FASTA.
    *
    * If filename ends in .fa or .fasta, saves as Fasta. If not, saves fragments
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/contig/NucleotideContigFragmentRDD.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/contig/NucleotideContigFragmentRDD.scala
@@ -155,7 +155,8 @@ case class NucleotideContigFragmentRDD(
    *   reference region. If the fragment start position and name is not defined,
    *   returns no regions.
    */
-  protected def getReferenceRegions(elem: NucleotideContigFragment): Seq[ReferenceRegion] = {
+  protected def getReferenceRegions(elem: NucleotideContigFragment,
+                                    stranded: Boolean = false): Seq[ReferenceRegion] = {
     ReferenceRegion(elem).toSeq
   }
 
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/contig/NucleotideContigFragmentRDD.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/contig/NucleotideContigFragmentRDD.scala
@@ -153,9 +153,8 @@ case class NucleotideContigFragmentRDD(
    * Returns all reference regions that overlap this NucleotideContigFragment.
    *
    * @param elem Fragment to extract a region from.
-   * @param stranded Whether or not to report stranded data for each
-   *   NucleotideContigFragment such that true reports stranded data and false
-   *   does not.
+   * @param stranded True to report stranded data for each
+   *   NucleotideContigFragment.
    * @return If a fragment is aligned to a reference location, returns a single
    *   reference region. If the fragment start position and name is not defined,
    *   returns no regions.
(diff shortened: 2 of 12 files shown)


adam-core/src/main/scala/org/bdgenomics/adam/rdd/feature/FeatureRDD.scala
```diff
@@ -327,8 +327,23 @@ case class FeatureRDD(rdd: RDD[Feature],
    * @return Since a feature maps directly to a single genomic region, this
    *   method will always return a Seq of exactly one ReferenceRegion.
    */
-  protected def getReferenceRegions(elem: Feature): Seq[ReferenceRegion] = {
-    Seq(ReferenceRegion.unstranded(elem))
+  protected def getReferenceRegions(elem: Feature,
+                                    stranded: Boolean = false): Seq[ReferenceRegion] = {
+    stranded match {
```
fnothaft: Don't `match` on a boolean.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/feature/FeatureRDD.scala
```diff
@@ -327,8 +327,23 @@ case class FeatureRDD(rdd: RDD[Feature],
...
+      case true => {
+        Seq({
+          try {
+            ReferenceRegion.stranded(elem)
+          } catch {
+            case e: IllegalArgumentException => ReferenceRegion.unstranded(elem)
```
fnothaft: How do we get an `IAE` here?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/fragment/FragmentRDD.scala
```diff
@@ -190,7 +190,8 @@ case class FragmentRDD(rdd: RDD[Fragment],
    * @param elem The Fragment to get the region from.
    * @return Returns all regions covered by this fragment.
    */
-  protected def getReferenceRegions(elem: Fragment): Seq[ReferenceRegion] = {
+  protected def getReferenceRegions(elem: Fragment,
+                                    stranded: Boolean = false): Seq[ReferenceRegion] = {
```
fnothaft: This method should allow stranded behavior.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDD.scala
```diff
@@ -232,7 +232,8 @@ case class AlignmentRecordRDD(
    * @param elem Read to produce regions for.
    * @return The seq of reference regions this read covers.
    */
-  protected def getReferenceRegions(elem: AlignmentRecord): Seq[ReferenceRegion] = {
+  protected def getReferenceRegions(elem: AlignmentRecord,
+                                    stranded: Boolean = false): Seq[ReferenceRegion] = {
```
fnothaft: Ditto here.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/feature/FeatureRDD.scala
devin-petersohn: `ReferenceRegion.stranded(feature: Feature)` has a require that the `Feature` have stranded data.

fnothaft: That seems like an erroneous requirement to me. I would suggest that we change that. Alternatively, I'd suggest that we check whether the feature has `INDEPENDENT` strand prior to calling, instead of catching an exception.

devin-petersohn: This requirement was recently [added](https://github.com/bigdatagenomics/adam/commit/e98ee2d1c5582364673fada70779aab5a73856b1#diff-521713a7bcc29cc21ce94e2f9cd722bfR294), and I'm not sure that we should remove it. I don't mind getting rid of the `try..catch` block, but I think that a `null` field should at least be logged. Maybe `ValidationStringency` here?

fnothaft: Ah, I see! I misunderstood that the `require` would fire if strand was Independent. I'd actually suggest that we not catch here.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/contig/NucleotideContigFragmentRDD.scala
```diff
@@ -136,12 +136,18 @@ case class NucleotideContigFragmentRDD(
   }
 
   /**
+   * Returns all reference regions that overlap this NucleotideContigFragment.
+   *
    * @param elem Fragment to extract a region from.
+   * @param stranded Whether or not to report stranded data for each
```
heuermh: True to report stranded data for each NucleotideContigFragment.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/feature/CoverageRDD.scala
```diff
@@ -236,9 +236,12 @@ case class CoverageRDD(rdd: RDD[Coverage],
    * return a Seq of exactly one ReferenceRegion.
    *
    * @param elem The Coverage to get an underlying region for.
+   * @param stranded Whether or not to report stranded data for each Coverage
```
heuermh: ...same as above

adam-core/src/main/scala/org/bdgenomics/adam/rdd/feature/FeatureRDD.scala
```diff
@@ -323,12 +323,21 @@ case class FeatureRDD(rdd: RDD[Feature],
   }
 
   /**
+   * Returns all reference regions that overlap this feature.
```
heuermh: In a different PR, I suggest adding doc to the definition in `GenomicRDD`, removing doc in all extensions of that method, and using the `override` keyword.  Does that work for Scala like it does in Java (i.e. the `@Override` annotation)?

fnothaft:
@devin-petersohn can you rebase this and clear the conflicts?

fnothaft:
Also, shouldn't the `stranded` parameter propagate through to the `flattenByReferenceRegion` method and all the joins?

devin-petersohn:
Yes, it should propagate. I also need to add unit tests for `stranded = true`.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -1185,7 +1190,8 @@ private case class GenericGenomicRDD[T](
     IntervalArray(rdd)
   }
 
-  protected def getReferenceRegions(elem: T): Seq[ReferenceRegion] = {
+  protected def getReferenceRegions(elem: T,
+                                    stranded: Boolean = false): Seq[ReferenceRegion] = {
```
fnothaft: In this function, `regionFn` should be updated to take `stranded` as a parameter, no?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -478,11 +478,13 @@ trait GenomicRDD[T, U <: GenomicRDD[T, U]] extends Logging {
     newRdd: RDD[T],
     newPartitionMap: Option[Array[Option[(ReferenceRegion, ReferenceRegion)]]] = None): U
 
-  protected def getReferenceRegions(elem: T): Seq[ReferenceRegion]
+  protected def getReferenceRegions(
+    elem: T,
+    stranded: Boolean = false): Seq[ReferenceRegion]
 
-  protected def flattenRddByRegions(): RDD[(ReferenceRegion, T)] = {
+  protected def flattenRddByRegions(stranded: Boolean = false): RDD[(ReferenceRegion, T)] = {
```
fnothaft: A small request here. Can we add an abstract field `canBeStranded`, and then log a warning message if `flattenRddByRegions` is called with `(!canBeStranded && stranded)`?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/feature/CoverageRDD.scala
```diff
@@ -251,9 +251,11 @@ case class CoverageRDD(rdd: RDD[Coverage],
    * return a Seq of exactly one ReferenceRegion.
    *
    * @param elem The Coverage to get an underlying region for.
+   * @param stranded True to report stranded data for each Coverage.
    * @return Sequence of ReferenceRegions extracted from Coverage.
    */
-  protected def getReferenceRegions(elem: Coverage): Seq[ReferenceRegion] = {
+  protected def getReferenceRegions(elem: Coverage,
+                                    stranded: Boolean = false): Seq[ReferenceRegion] = {
```
fnothaft: We do support generating coverage data for independent/positive/negative strands. I don't know if we expose that anywhere. If we don't, can you open a ticket for us to follow up on this? CC @akmorrow13

adam-core/src/main/scala/org/bdgenomics/adam/rdd/fragment/FragmentRDD.scala
```diff
@@ -205,11 +205,24 @@ case class FragmentRDD(rdd: RDD[Fragment],
...
-  protected def getReferenceRegions(elem: Fragment): Seq[ReferenceRegion] = {
-    elem.getAlignments
-      .flatMap(r => ReferenceRegion.opt(r))
-      .toSeq
+  protected def getReferenceRegions(elem: Fragment,
+                                    stranded: Boolean = false): Seq[ReferenceRegion] = {
+    if (stranded) {
+      elem.getAlignments
+        .map(r => {
+          if (r.getReadMapped && r.getReadNegativeStrand != null) {
```
fnothaft: IMO, if you call `stranded` on reads with `readNegativeStrand == null`, you should get the `IllegalArgumentException`, not an `unstranded` region.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDD.scala
```diff
@@ -255,10 +255,16 @@ case class AlignmentRecordRDD(
...
    *
    * @param elem Read to produce regions for.
+   * @param stranded True to report stranded data for each AlignmentRecord.
    * @return The seq of reference regions this read covers.
    */
-  protected def getReferenceRegions(elem: AlignmentRecord): Seq[ReferenceRegion] = {
-    ReferenceRegion.opt(elem).toSeq
+  protected def getReferenceRegions(elem: AlignmentRecord,
+                                    stranded: Boolean = false): Seq[ReferenceRegion] = {
+    if (stranded && elem.getReadNegativeStrand != null) {
```
fnothaft: Same here, WRT `readNegativeStrand == null`.

heuermh:
Closing as WontFix.  Feel free to create a new PR after rebasing against git head.

PR #1557, opened 2017-06-05
[ADAM-1449] Add loadSequenceDictionary to ADAM context.

### LINKED ISSUE #1449: Port over code for populating SequenceDictionaries from .dict files
Some code to do this is in https://github.com/heuermh/cannoli/pull/11. Additionally, we should update the various places in the code that "create" a sequence dictionary from thin air to support providing a `.dict` file.

### LINKED ISSUE #1534: Compute a SequenceDictionary from a *.genome file
Bedtools uses `*.genome` files to serve the same purpose as our `SequenceDictionary`, so instead of computing based on the bounds of the data would it be possible to use the `*.genome` file to create the `SequenceDictionary`?

heuermh:
Fixes #1449 
Fixes #1534 

Implements `ADAMContext.loadSequenceDictionary` that supports HTSJDK sequence dictionary (.dict), Bedtools genome file (.genome), and UCSC Genome Browser chromInfo (.txt) files.

Taking suggestions on how best to add this to other load methods.


devin-petersohn:
Do we want to allow users to provide `SequenceDictionary` when they've loaded in a `*.adam` file?

adam-core/src/main/scala/org/bdgenomics/adam/util/FileExtensions.scala
```diff
@@ -118,4 +118,28 @@ object FileExtensions {
       pathName.endsWith(".vcf.bgzf") ||
       pathName.endsWith(".vcf.bgz")
   }
+
+  /**
+   * @param pathName The path name to match.
+   * @return Returns true if the path name matches an HTSJDK sequence dictionary (.dict) extension.
+   */
+  private[adam] def isDictExt(pathName: String): Boolean = {
```
fnothaft: Small nit, but instead of making all the methods `private[adam]`, can we make the object `private[adam]`.

adam-core/src/main/scala/org/bdgenomics/adam/util/SequenceDictionaryReader.scala
```diff
@@ -0,0 +1,66 @@
...
+
+    // open the underlying file and wrap it up in a buffer for htsjdk
+    val is = fs.open(path)
+    val reader = new BufferedLineReader(is)
+
+    // create a codec
+    // even though we are just using the codec to read, it requires us to give
+    // a writer...? setting the writer to stderr seems like a reasonable choice.
+    val codec = new SAMSequenceDictionaryCodec(
+      new BufferedWriter(new OutputStreamWriter(System.err)))
```
fnothaft: I agree that this is a reasonable choice. That said, what happens if we provide a null pointer?

adam-core/src/main/scala/org/bdgenomics/adam/util/FileExtensions.scala
heuermh: +1

heuermh:
Suggestions from meeting today:
 - Bedtools documentation often refers to `.genome.txt` files
 - Check for sequence dictionary file with matching base name when loading features

adam-core/src/test/scala/org/bdgenomics/adam/rdd/ADAMContextSuite.scala
```diff
@@ -538,4 +538,91 @@ class ADAMContextSuite extends ADAMFunSuite {
...
+
+    val sequences = featureRdd.sequences
+    assert(sequences.records.size === 93)
+    assert(sequences("chr1").isDefined)
+    assert(sequences("chr1").get.length === 249250621L)
+    assert(sequences("chr17_gl000206_random").isDefined)
+    assert(sequences("chr17_gl000206_random").get.length === 41001L)
+  }
+
+  sparkTest("load BED features with Bedtools .genome file as sequence dictionary, no matching features") {
```
heuermh: What should happen here?
Similar with features that match on reference name but not on coordinates (e.g. start > reference length).

adam-core/src/main/scala/org/bdgenomics/adam/util/SequenceDictionaryReader.scala
heuermh: It passes unit tests.  I'd lean towards keeping this as-is, given how little I trust htsjdk

heuermh:
I am liking the suggestion that we check for sequence dictionary file with matching base name when loading features less now that I've thought about it.

What if `pathName` is a glob that matches more than one feature file?  I'd rather the user be more explicit, calling
```scala
val sd = sc.loadSequenceDictionary("hg19.genome")
val features = sc.loadFeatures("**.bed", optSequenceDictionary = Some(sd))
```
to share one sequence dictionary across a bunch of BED files.

adam-core/src/main/scala/org/bdgenomics/adam/util/FileExtensions.scala
```diff
@@ -112,10 +112,35 @@ object FileExtensions {
...
+    pathName.endsWith(".dict")
+  }
+
+  /**
+   * @param pathName The path name to match.
+   * @return Returns true if the path name matches a Bedtools genome file (.genome) extension.
+   */
+  def isGenomeExt(pathName: String): Boolean = {
+    pathName.endsWith(".genome") ||
+      pathName.endsWith(".genome.txt")
```
devin-petersohn: We can probably get rid of this and be strict about it being a `*.genome` file. As long as we give good feedback about our requirements. I haven't seen people use `genome.txt` extensions, it's been mostly either `.txt` or `.genome`.

heuermh: All three (`.genome`, `.genome.txt`, and `.txt`) go through the same code path.  The only difference is the log message.

heuermh:
> Do we want to allow users to provide SequenceDictionary when they've loaded in a *.adam file?

No, I don't really understand what that would mean.

adam-core/src/test/scala/org/bdgenomics/adam/rdd/ADAMContextSuite.scala
heuermh: Will add `ValidationStringency` to ignore or throw exception.

fnothaft:
>> Do we want to allow users to provide SequenceDictionary when they've loaded in a *.adam file?
>
> No, I don't really understand what that would mean.

I mean, if anyone wants to do that, they can load the `SequenceDictionary` manually and attach it to the `GenomicRDD`.

adam-core/src/test/scala/org/bdgenomics/adam/rdd/ADAMContextSuite.scala
heuermh: On further consideration, since checking at load time might significantly slow down the common case unnecessarily, where the sequence dictionary is created from the features themselves, I'm leaning towards a separate pull request to implement

```scala
def filterBySequenceDictionary(sd: SequenceDictionary): U = {
  ...
}
```
and
```scala
def filterBySequenceDictionary(): U = {
  filterBySequenceDictionary(this.sequences)
}
```
methods on `GenomicRDD`.

fnothaft:
@heuermh `filterBySequenceDictionary` SGTM! Would that filter out all items that were mapped to contigs that were not present, and all items that mapped off the end of a contig that was known? How would it handle unmapped items?

heuermh:
See #1575 

> Would that filter out all items that were mapped to contigs that were not present, and all items that mapped off the end of a contig that was known?

Yes

> How would it handle unmapped items?

I'm thinking of converting `SequenceDictionary` to `RDD[Feature]` and using the appropriate set operation.

adam-core/src/test/scala/org/bdgenomics/adam/rdd/ADAMContextSuite.scala
heuermh: Per discussion on #1575, will leave this test case as is

heuermh:
Rebased and squashed down to two commits.  Fixes #1588.

adam-core/src/main/scala/org/bdgenomics/adam/util/GenomeFileReader.scala
```diff
@@ -0,0 +1,61 @@
...
+   *
+   * @param filePath The path to read the dictionary from.
+   * @param sc The SparkContext to use for configuration.
+   * @return Returns a populated sequence dictionary.
+   */
+  def apply(filePath: String,
+            sc: SparkContext): SequenceDictionary = {
+
+    val records = sc
+      .textFile(filePath)
```
fnothaft: There's nothing inherently wrong with using `sc.textFile` to load this, but it seems a bit... aggressive to use it to load such a small file.

adam-core/src/main/scala/org/bdgenomics/adam/util/SequenceDictionaryReader.scala
fnothaft: SGTM

fnothaft:
Merged! Thanks @heuermh!

PR #1561, opened 2017-06-09
[ADAM-1533] Set Theory

devin-petersohn:
WIP. Looking for quick feedback on architecture changes. The idea is to move the prepare code into the individual set theory primitive classes, which will unbloat `GenomicRDD` a bit.

Most of the primitives can be reduced to post-processing on the `ShuffleRegionJoin` implementations now that I have generalized joins to allow distances also. 

TODO:
- [ ]  Implement one-to-self set theory primitives
- [ ]  Create Test cases
- [ ]  Better/More complete Docs


adam-core/src/main/scala/org/bdgenomics/adam/rdd/settheory/Closest.scala
```diff
@@ -0,0 +1,190 @@
...
+
+/**
+ * A trait describing closest implementations that are based on sort-merge.
+ *
+ * @tparam T The type of the left records.
+ * @tparam U The type of the right records.
+ * @tparam RT The resulting type of the left after the join.
+ * @tparam RU The resulting type of the right after the join.
+ */
+sealed abstract class Closest[T: ClassTag, U: ClassTag, RT, RU]
```
fnothaft: Please add inline docs to this class.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/settheory/SetTheory.scala
```diff
@@ -0,0 +1,346 @@
...
+import org.bdgenomics.adam.models.ReferenceRegion
+import scala.collection.mutable.ListBuffer
+
+private[settheory] sealed abstract class SetTheory extends Serializable {
+
+  protected val threshold: Long
+  protected val optPartitionMap: Option[Array[Option[(ReferenceRegion, ReferenceRegion)]]]
+
+  /**
+   * The condition that should be met in order for the primitive to be
```
fnothaft: This would be clearer with an example.

fnothaft: Also, if this package is set theoretic, then we should describe it through set theoretic notation, right? I think this method would be "This method evaluates whether a given pair of regions should be members of the set output by this set theoretic operation." Is that correct?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/settheory/SetTheory.scala
```diff
@@ -0,0 +1,346 @@
...
+
+/**
+ * The parent class for all inter-collection set theory operations.
+ *
+ * @tparam T The left side row data.
+ * @tparam U The right side row data.
+ * @tparam RT The return type for the left side row data.
+ * @tparam RU The return type for the right side row data.
+ */
+private[settheory] abstract class SetTheoryBetweenCollections[T, U, RT, RU] extends SetTheory {
```
fnothaft: OOC, is there a reason to prefer an abstract class to a trait here? I know that `ShuffleRegionJoin` is an abstract class, but I was looking through the history and couldn't figure out what was the impetus for the change. CCing in @ryan-williams who made the change in https://github.com/bigdatagenomics/adam/commit/9ff0fa5ec256bf057d29d5c2613e424ae7fdde01.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/settheory/SetTheory.scala
```diff
@@ -0,0 +1,346 @@
...
+
+/**
+ * Contains all the caching data for a set theory operation.
+ *
+ * @tparam U The right side record type.
+ * @tparam RT The left side result type.
+ * @tparam RU The right side result type.
+ */
+private[settheory] class SetTheoryCache[U, RT, RU] {
+  // caches potential hits
```
fnothaft: Whitespace before comments throughout this class.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/settheory/ShuffleRegionJoin.scala
```diff
@@ -0,0 +1,334 @@
...
+ *
+ * @tparam T The type of the left records.
+ * @tparam U The type of the right records.
+ * @tparam RT The resulting type of the left after the join.
+ * @tparam RU The resulting type of the right after the join.
+ */
+sealed abstract class ShuffleRegionJoin[T: ClassTag, U: ClassTag, RT, RU]
+    extends SetTheoryBetweenCollections[T, U, RT, RU] with SetTheoryPrimitive {
+
+  override protected def condition(firstRegion: ReferenceRegion,
```
fnothaft: Isn't `condition` abstract in the superclass? If so, we shouldn't be using the `override` modifier.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/settheory/ShuffleRegionJoin.scala
```diff
@@ -0,0 +1,334 @@
...
+  override protected def prepare(): (RDD[(ReferenceRegion, T)], RDD[(ReferenceRegion, U)]) = {
+    
+    val (preparedLeft, destinationPartitionMap) = {
+      if(optPartitionMap.isDefined) {
+        (leftRdd, optPartitionMap.get.map(_.get))
+      } else {
+        val sortedLeft = leftRdd.sortByKey()
+        val partitionMap =
+          sortedLeft.mapPartitions(getRegionBoundsFromPartition).collect
+        (sortedLeft, partitionMap.map(_.get))
```
fnothaft: If the left RDD is not already sorted, this will lead to an execution DAG that does two full sorts of the left RDD.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/settheory/ShuffleRegionJoin.scala
```diff
@@ -0,0 +1,334 @@
...
+            firstRegion.start,
+            secondRegion.end),
+            index)
+        }
+      })
+
+    // convert to an IntervalArray for fast range query
+    val partitionMapIntervals = IntervalArray(
+      adjustedPartitionMapWithIndex,
+      adjustedPartitionMapWithIndex.maxBy(_._1.width)._1.width,
```
fnothaft: Prefer `adjustedPartitionMapWithIndex.map(_._1.width).max`.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/settheory/Closest.scala
```diff
@@ -0,0 +1,190 @@
...
+    extends SetTheoryBetweenCollections[T, U, RT, RU]
+    with SetTheoryPrimitive {
+
+  var currentClosest: ReferenceRegion = ReferenceRegion.empty
```
fnothaft: Thread safety...

adam-core/src/main/scala/org/bdgenomics/adam/rdd/settheory/Closest.scala
fnothaft: Also, I feel like we're inverting an abstraction here. IMO, the correct way to implement this would be to run a self joinAndGroupByLeft and to then:

```
val (left, rightIter) = kv
(left, rightIter.min(_.distance(left)))
```

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -836,25 +791,39 @@ trait GenomicRDD[T, U <: GenomicRDD[T, U]] extends Logging {
...
-    optPartitions: Option[Int] = None)(
+    optPartitions: Option[Int] = None,
+    threshold: Long = 0L)(
       implicit tTag: ClassTag[T],
       xTag: ClassTag[X],
       txTag: ClassTag[(T, X)]): GenomicRDD[(T, X), Z] = InnerShuffleJoin.time {
 
-    val (leftRddToJoin, rightRddToJoin) =
-      prepareForShuffleRegionJoin(genomicRdd, optPartitions)
+    val preparedLeft =
```
fnothaft: If we're moving the preparation logic into the join primitive, this code should move with it.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/settheory/Closest.scala
devin-petersohn: It hasn't been a problem yet, but I am working on a fix for this in the case that the Spark Serialization rules change. I don't like this here either.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/settheory/ShuffleRegionJoin.scala
devin-petersohn: This was an attempted optimization. Rather than iterating over the entire list twice, we do it once with `maxBy`. There is certainly little effect because the number of partitions will presumably be relatively small. I'll go ahead and change it.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/settheory/Closest.scala
devin-petersohn: Closest doesn't necessarily overlap, and is performed between two `RDD`s. In your proposed architecture, in order to be truly exhaustive we'd end up joining each record on the left with all records with the same `referenceName` on the right.

adam-core/src/main/scala/org/bdgenomics/adam/models/ReferenceRegion.scala
```diff
@@ -85,6 +85,13 @@ object ReferenceRegion {
   implicit def orderingForOptionalPositions = OptionalRegionOrdering
 
   /**
+   * Creates an empty ReferenceRegion.
+   *
+   * @return An empty ReferenceRegion.
+   */
+  private[adam] val empty: ReferenceRegion = ReferenceRegion("", 0L, 0L)
```
devin-petersohn: I agree. This was in here temporarily for some tests.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/settheory/SetTheory.scala
heuermh: "SetTheory" doesn't work for me as a class and package name.
Perhaps for package `rdd.sets`, and `SetOperation` for the top level abstract class?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/settheory/ShuffleRegionJoin.scala
heuermh: Is that proper style? Is it still useful for inheriting docs, as `@Overrides` does in java?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -66,7 +67,6 @@ private[rdd] object GenomicRDD {
    * Replaces file references in a command.
    *
    * @see pipe
-   *
```
devin-petersohn: I will revert these on a future push. I want to avoid reverting and re-reverting repeatedly.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/settheory/SetTheory.scala
devin-petersohn: I went ahead and put this back to a trait. I believe it makes more sense, especially now that we are using the `GenomicRDD`.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/settheory/SetTheory.scala
devin-petersohn: +1 on package and class rename

devin-petersohn:
> all of the set theoretic operations we are implementing should be able to be implemented on top of the region join primitive

Except for unbounded closest, this is true. Unbounded closest is the one of the reasons the `SetOperation` class is abstracted this way; it uses much of the same architecture as joins, there are only differences in how to compute the closest and how to copartition the data. There are also `SetOperation` abstractions we need when we are dealing with a single collection, i.e. `Merge`. `Merge` can be implemented with a series of joins, but it would be much more expensive due to the increased cost/number of shuffle phases. I still have the `SetOperationWithSingleCollection` (and all the one-to-self primitives) to implement.

> I believe that all the primitives should map into a `join [ -> aggregate ] -> predicate` flow

I agree 100% that the majority of the operations should be performed with a join as the first phase, independent of the type of join. However, this will not work, or be efficient, for `UnboundedClosest` and many one-to-self set theory operations, which are not yet in this PR. I can start laying the groundwork for allowing this, but fully implementing it will probably require a minor refactor of the `broadcastRegionJoin` code, and perhaps belongs in a separate PR.

> I think that if the entrypoint to `ShuffleRegionJoin` clearly assumes the contract that the two RDDs are copartitioned

The way that it is currently architected, it does not assume this. The goal for pulling the `prepareForShuffleRegionJoin` code out was to reduce the bloat in `GenomicRDD`. It seems reasonable to me for `SetOperation`s to prepare the data themselves, particularly if they have a different partitioning requirement than `shuffleRegionJoin`. This also gives us strong guarantees around not keeping `RDD`s that have duplicated records from copartitioning. If we do want to allow both `SetOperation(GenomicRDD)` and `SetOperation(RDD[(ReferenceRegion, T)])`, it seems that having the prepare code in the `SetOperation` class would be easiest. In the case that users want to call `ShuffleRegionJoin(...).compute()` themselves, they can do so without having to worry about how their data is partitioned. I agree that ideally, users would just call `GenomicRDD.shuffleRegionJoin()`, but there are cases where they cannot. I personally believe there is a stronger case for moving it into the `SetOperation`s and letting each class define the optimal partition scheme for its respective operation, but I am happy to hear the opposing arguments.

> I'm a strong -1 on opening up protections on the partition map.

I am not sure what you mean by opening up protections on the partition map since I have moved from `private[rdd]` -> `protected`. I do understand what you are saying about setting the `optPartitionMap` access to `private[GenomicRDD]`, but I disagree with making it that strict. Part of the reason I switched to passing in `GenomicRDD`s for the `SetOperation`s was because we have the `prepare()` code in the `SetOperation`s, so they also need to know how the data is partitioned. One-to-self operations need access to the `PartitionMap` structure to avoid extra shuffle phases and reduce skew.

> As you may have guessed, I am also a strong -1 on making a `PartitionMap` class, esp. if it is not private to `GenomicRDD`.

The `PartitionMap` class is a part of a (hopefully) better way of handling sorted data. I think we would rather know that the data is sorted, independent of whether or not the data has a partition map (which would begin to solve our issues of sorted legacy file formats). I plan to have an accompanying `object` that builds the `PartitionMap` when given a `GenomicRDD` or `RDD[(ReferenceRegion, T)]. Making the `partitionMap` a `lazy val` will also reduce the amount of code and avoid computing it when it isn't needed. `GenomicRDD`s will take in a `Boolean` value for `sorted` rather than an optional `partitionMap`. There are also a lot of common operations performed on the `PartitionMap` (`toIntervalArray`). The `PartitionMap` `class` would live in its own file and be `private[rdd]`. Having a separate `PartitionMap` class would also clean out a lot of code in `GenomicRDD` related to computing the `optPartitionMap`.

Sorry for the wall of text. Feel free to address anything I've said.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/sets/ShuffleRegionJoin.scala
```diff
@@ -0,0 +1,408 @@
...
+                                         iter: Iterable[(ReferenceRegion, X)]): Iterable[(T, X)] = {
+    iter.map(f => (currentLeft._2, f._2))
+  }
+}
+
+/**
+ * Perform an Inner ShuffleRegionJoin. This is publicly accessible to be
+ * compatible with legacy code.
+ */
+object InnerShuffleRegionJoin {
```
devin-petersohn: Pinging @heuermh and @fnothaft to get your thoughts on this architecture. This was what I came up with to guarantee consistency with our previous implementation. From the user point of view, it looks exactly the same as it did before, despite looking different under the hood.

devin-petersohn:
Closing as won't merge. This work belongs in a downstream app.

PR #1565, opened 2017-06-09
[ADAM-1564] Read VCF header from stream in VCFOutFormatter.

### LINKED ISSUE #1564: VCF header read from stream ignored in VCFOutFormatter
[VCFOutFormatter.scala#L59](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/variant/VCFOutFormatter.scala#L59) reads the VCF header from the stream and ignores it.

It should merge the header with `headerLines` specified in the ctr as [ADAMContext.scala#L1142](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala#L1142) does.

heuermh:
Fixes #1564 
Depends on #1563


adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -44,6 +44,7 @@ import org.apache.spark.rdd.MetricsContext._
 import org.apache.spark.rdd.RDD
 import org.apache.spark.storage.StorageLevel
 import org.bdgenomics.adam.converters._
+import org.bdgenomics.adam.converters.VariantContextConverter._
```
fnothaft: -1 on the `._` import, otherwise OK.

adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantContextConverter.scala
```diff
@@ -150,6 +153,74 @@ object VariantContextConverter {
       case GenotypeAllele.ALT                                => Allele.create(g.getVariant.getAlternateAllele)
     }
   }
+
+  private[adam] def cleanAndMixInSupportedLines(
```
fnothaft: Beyond moving this into `VariantContextConverter`, were there any changes to the method? If so, I'll make a second review pass, but for now I'm assuming it is just a move.

heuermh: Move only.  I don't really like it there to be honest, suggestions welcome.

fnothaft: Yeah, it's an awkward method; I don't think there's a perfect home for it, but this seems like a more reasonable place than it's former home.

fnothaft:
Merged! Thanks @heuermh!

PR #1587, opened 2017-07-04
[ADAM-1579] Add unit test coverage for BED12 format.

### LINKED ISSUE #1579: Add unit test coverage for BED12 parser and writer
Related to #1418 and #1405. The BED6 codepath is well covered, but BED12 is not covered on either the read and write path.

fnothaft:
Resolves #1579. Found small bugs that would lead to last 6 fields not getting written if the thickStart field wasn't written. Also, identified a bug where the strand would be set to unknown if an independent strand was provided.


heuermh:
Thank you, @fnothaft

Issue #1584, opened 2017-06-27
Support sorted knowledge for legacy files


fnothaft:
E.g., sorted SAM/BAM/CRAM, VCF.

Issue #1558, opened 2017-06-06
New model PartitionMap for Array[Option[(ReferenceRegion, ReferenceRegion)]]


heuermh:
After having to type `Option[Array[Option[(ReferenceRegion, ReferenceRegion)]]]` a few times I wonder if it might be worth creating a new model class `PartitionMap`. And then having said that, I don't like `Map` in the name, because `PartitionMap` ! is_a `Map`.

heuermh:
Closing as WontFix

Issue #1519, opened 2017-05-10
Support  ShuffleRegionJoins when an rdd is sorted by contig number


devin-petersohn:
Currently we only support lexicographically sorted sort-merge (ShuffleRegion) joins.

heuermh:
Closing as WontFix

PR #1397, opened 2017-02-21
[ADAM-882] R API

fnothaft:
WIP towards #882. Still bringing in tests, functions beyond reads. On top of #1391.

TODOs:

- [x] Full docs for loadAlignments
- [x] Add loadCoverage
- [x] Document catch 22 in build

Union method can't be added; necessary support in SparkR for dealing with lists of jobjs in R does not exist.

PR #1590, opened 2017-07-07
Add SortedGenomicRDD trait, refactor shuffle joins and pipe

### LINKED ISSUE #1583: Integrate sorted knowledge with pipe API

fnothaft:
Resolves #1583, #1584, #1519, #1558. Based on #1397.

1. Adds a SortedGenomicRDD trait, and concrete implementations of this trait for all datatypes. This trait eliminates the need for the partition map (#1558), by tracking the underlying partitioner used to partition the sorted data and an average record size for the sorted RDD. This can be used to obtain bounds similar to the partition map, but can be computed at a lower cost. Additionally, this data can be cheaply created by peeking at the first item on each partition, which enables us to support legacy formats including FASTA, coordinate sorted SAM/BAM/CRAM, and VCF where data is sorted on disk but no partition map is available (#1584). When we infer the partitioner, we can determine whether the partitioner is lexicographically sorted or sorted by index by looking at the reference sequence order in the partitioner and the sequence dictionary that is attached to the RDD. Since we then use the inferred partitioner to co-partition data during a shuffle region join, we can support both lexicographically and sequence index ordered sort orders in the region join (#1519).
3. Adds `sortAndFlank` method to GenomicRDD, which is overridden by `SortedGenomicRDD` with a variant that is shuffle-free. Refactors the `pipe` transformation to use the `sortAndFlank` method instead of shuffling with the `ManualRegionPartitioner`.

Still needs benchmarking, more unit tests in places, and more documentation.


fnothaft:
Test failure is due to a mismatch between Jenkins and the changes to jenkins-test.sh; specifically, #1397 changes the Spark 2 integration tests to look for Spark=2.1.0, while Jenkins is setting 2.0.0.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -1191,9 +1201,153 @@ private case class GenericGenomicRDD[T](
...
+        ((leftRdd.partitioner, rightRdd.partitioner) match {
+          case (IndexedGenomicRangePartitioner(leftBounds, _),
+            IndexedGenomicRangePartitioner(rightBounds, _)) => {
+            leftBounds.zip(rightBounds).forall(p => p._1 == p._2)
+          }
+          case (LexicographicalGenomicRangePartitioner(leftBounds),
+            LexicographicalGenomicRangePartitioner(rightBounds)) => {
+            leftBounds.zip(rightBounds).forall(p => {
+              p._1.referenceName == p._2.referenceName &&
+                p._1.start == p._2.start
```
devin-petersohn: This won't solve the problem of determining that the data is copartitioned properly, correct? We are doing overlap joins, so the starts being equivalent doesn't actually tell us whether all data that should be joined together is colocated together. Please let me know if I've missed something.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -1191,9 +1201,153 @@ private case class GenericGenomicRDD[T](
...
+    gRdd match {
+      case _: SortedGenomicRDD[T, U] => {
+        touchFile(sc, "%s/_sorted".format(fileName))
+      }
+      case _ => { /* no-op */ }
+    }
+  }
 
-    copy(rdd = newRdd, optPartitionMap = newPartitionMap)
+  def partitionersAreCompatible(leftRdd: SortedGenomicRDD[_, _],
```
devin-petersohn: I like the idea, but I'm not sure it's possible to make it simple enough to be worth computing. Please see my comment below regarding implementation.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -1191,9 +1201,153 @@ private case class GenericGenomicRDD[T](
...
+      false
+    }
+  }
+}
+
+trait SortedGenomicRDD[T, U <: GenomicRDD[T, U]] extends GenomicRDD[T, U] {
+
+  protected val partitioner: Partitioner
+
+  override final def prepareForShuffleRegionJoin[X, Y <: GenomicRDD[X, Y]](
```
devin-petersohn: I have an idea for resolving skew in both sides by sampling from both RDDs and partitioning accordingly. It would be much more accurate and balanced than the previous implementation.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -806,24 +819,23 @@ trait GenomicRDD[T, U <: GenomicRDD[T, U]] extends Logging {
    *   None, the number of partitions on the resulting RDD does not change.
    * @return a case class containing all the prepared data for ShuffleRegionJoins
    */
-  private def prepareForShuffleRegionJoin[X, Y <: GenomicRDD[X, Y], Z <: GenomicRDD[(T, X), Z]](
+  protected def prepareForShuffleRegionJoin[X, Y <: GenomicRDD[X, Y]](
```
devin-petersohn: See comment below in `SortedGenomicRDD` regarding balance of both RDDs.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
fnothaft: Yeah, we should be able to just pick a `RangePartitioner` that samples from both RDDs, no?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
fnothaft: I'd need to look back at the code to confirm 100%, but from memory, the way that this code should work is to use the estimated longest record size from the right RDD (I believe I call this "flank margin") to move records from partition `n + 1` to partition `n` if they are within `flank_margin` bases from the start of partition `n + 1`. This is sufficient to satisfy the copartitioning requirement we have.

I'd be glad to put together a formal writeup of this, if you'd like.

fnothaft:
> I am curious about the motivation for creating all the new classes/traits. Why do we need SortedGenomicRDD trait?

Adding a sorted trait was the original thing we'd tried doing, no? IIRC, we went down that road but then we ran into problems with mix-ins.

That said, I think the sorted trait is a natural fit because we can leverage sorted knowledge in more places than just the shuffle joins and the pipe API. E.g., in this PR, I added some simple functionality to the variant RDDs where we maintain sort order going from VariantContextRDD->VariantRDD,GenotypeRDD and VariantRDD->VariantContextRDD by overriding methods in the Sorted{Genotype,Variant,VariantContext}RDD traits. We'd want to extend this pattern anywhere that we have an `isSorted` parameter today.

heuermh:
@fnothaft Might be a herculean task, but could you attempt to rebase this?  If not possible, I'll try to pull out only the relevant changes in a new pull request.

heuermh:
Closing as WontFix.  While I would love to have these changes in, I am not able to push these changes forward.  Feel free to create a new PR after rebasing against git head.

PR #1591, opened 2017-07-07
[ADAM-1417] Removed unused Projection.apply method, add test for Filter.

### LINKED ISSUE #1417: Clean up possibly unused methods in Projection
There are several apply methods that overlap and two are uncovered by tests. The uncovered apply methods should be deleted.

fnothaft:
Resolves #1417.


heuermh:
Thank you, @fnothaft

PR #1595, opened 2017-07-08
[ADAM-1444] Ignore failed push to Coveralls.

### LINKED ISSUE #1444: Ignore failed push to Coveralls during CI builds
Coveralls frequently times out during CI when we're pushing coverage results. We should see if we can turn this into a warning instead of an error.

fnothaft:
Resolves #1444. Bumps to latest version of coveralls plugin and sets `failOnServiceError`, which ignores internal errors in Coveralls when pushing.


heuermh:
Thank you, @fnothaft!

PR #1596, opened 2017-07-09
Remove unused FeatureRDD.sortByReference.

fnothaft:
Found this when working on #1590.


heuermh:
GFF3 files like those provided by Ensembl with full gene models have deeply nested, overlapping, and identical-by-region features.  I found sort order to not be consistent without the deep feature ordering.

devin-petersohn:
@fnothaft What milestone should we have this set for?

fnothaft:
@devin-petersohn this is low priority; 0.24.0 is fine.

PR #1598, opened 2017-07-11
[ADAM-1576] Allow translation between two different GenomicRDD types.

### LINKED ISSUE #1576: Unable to save GenomicRDDs after a join.
There are two things at work here:
1.) We are upcasting to `GenomicRDD` after a join, which does not have any save functionality attached.
2.) `GenericGenomicRDD` extends `GenomicRDD`, but does not have any save functionality attached.

I think it would be ok to create a `GenomicRDD` that is specific to the result joins and other set operations, mimicking the [`PairRDDFunctions`](https://spark.apache.org/docs/2.1.1/api/java/org/apache/spark/rdd/PairRDDFunctions.html) in Spark. I don't see it as a problem for the user to be exposed to a `GenomicPairRDD` or something like that.

What type should a join return? How should we handle subsequent joins on a joined `GenomicRDD`?

fnothaft:
Resolves #1576. Adds `transmute` function, which is analogous to `transform`, but which allows any return type. This uses an implicit function a la `pipe` to build the output GenomicRDD.

Can I get a quick review pass on this? If this general direction looks good, I will flesh out the implicits for all the allowable transformations.

- [x] Extend to all primary types
- [x] Add docs

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
@@ -42,6 +42,7 @@ import org.apache.parquet.hadoop.util.ContextUtil
 import org.apache.spark.SparkContext
 import org.apache.spark.rdd.MetricsContext._
 import org.apache.spark.rdd.RDD
+import org.apache.spark.sql.Dataset
 import org.bdgenomics.adam.converters._
 import org.bdgenomics.adam.instrumentation.Timers._
 import org.bdgenomics.adam.io._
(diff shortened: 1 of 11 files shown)


fnothaft:
OK great! I'll flesh this out today. Don't merge as is!

fnothaft:
This is good to go from my end.

fnothaft:
Rebased and cleared conflicts.

fnothaft:
Rebased and cleared conflicts.

fnothaft:
Also, WRT @heuermh's comment, there's nothing that stops an enterprising user from breaking our preconditions today. This might make it easier, but not really any easier than just using `transform` today is.

fnothaft:
Looks like a transient s3 failure.

fnothaft:
Jenkins, test this please.

heuermh:
Thank you, @fnothaft!

PR #1545, opened 2017-05-22
[ADAM-1257] Add program record support for alignment/fragment files.

fnothaft:
Resolves #1257. Depends on https://github.com/bigdatagenomics/adam/pull/1523. WIP:

- [x] ~~Need to connect to `RecordGroup`~~ deferring for now
- [x] Need to look at `org.bdgenomics.adam.models.ProgramGroup`.

PR #1600, opened 2017-07-12
[ADAM-1599] Add explicit functions for updating GenomicRDD metadata.

### LINKED ISSUE #1599: Scala copy method not supported in abstract classes such as AlignmentRecordRDD
See issue https://github.com/bigdatagenomics/cannoli/issues/55

fnothaft:
Resolves #1599. Adds functions that replaces or appends to the sequence dictionary/record groups/header lines/sample metadata that is attached to a GenomicRDD. This is necessary as a59b8ed948db1bbf50260a61ed404b8325df0563 eliminated the `copy` methods that were attached to all implementations of `GenomicRDD` by making the exposed implementations abstract classes instead of case classes.

~One TODO:~
- [x] ~Extend to Python API~ Punted to #1604

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
@@ -32,9 +32,11 @@ import org.apache.spark.sql.{ DataFrame, Dataset }
 import org.apache.spark.storage.StorageLevel
 import org.bdgenomics.adam.instrumentation.Timers._
 import org.bdgenomics.adam.models.{
+  RecordGroup,
   RecordGroupDictionary,
   ReferenceRegion,
-  SequenceDictionary
+  SequenceDictionary,
+  SequenceRecord
 }
 import org.bdgenomics.formats.avro.{
   Contig,
@@ -121,6 +123,34 @@ trait GenomicRDD[T, U <: GenomicRDD[T, U]] extends Logging {
    * aligned to.
    */
   val sequences: SequenceDictionary
+
+  /**
+   * Replaces the sequence dictionary attached to a GenomicRDD.
+   *
+   * @param newSequences The new sequence dictionary to attach.
+   * @return Returns a new GenomicRDD with the sequences replaced.
+   */
+  def replaceSequences(newSequences: SequenceDictionary): U
+
+  /**
+   * Appends sequence metadata to the current RDD.
+   *
+   * @param sequencesToAdd The new sequences to append.
+   * @return Returns a new GenomicRDD with the sequences appended.
+   */
+  def addSequences(sequencesToAdd: SequenceDictionary): U = {
+    replaceSequences(sequences ++ sequencesToAdd)
+  }
+
+  /**
+   * Appends metadata for a single sequence to the current RDD.
+   *
+   * @param sequenceToAdd The sequence to add.
+   * @return Returns a new GenomicRDD with this sequence appended.
+   */
+  def addSequence(sequenceToAdd: SequenceRecord): U = {
+    addSequences(SequenceDictionary(sequenceToAdd))
+  }
 
   /**
    * The underlying RDD of genomic data, as a JavaRDD.
@@ -1258,6 +1288,11 @@ private case class GenericGenomicRDD[T](
     optPartitionMap: Option[Array[Option[(ReferenceRegion, ReferenceRegion)]]] = None)(
         implicit tTag: ClassTag[T]) extends GenomicRDD[T, GenericGenomicRDD[T]] {
 
+  def replaceSequences(
+    newSequences: SequenceDictionary): GenericGenomicRDD[T] = {
+    copy(sequences = newSequences)
+  }
+
   def union(rdds: GenericGenomicRDD[T]*): GenericGenomicRDD[T] = {
     val iterableRdds = rdds.toSeq
     GenericGenomicRDD(rdd.context.union(rdd, iterableRdds.map(_.rdd): _*),
@@ -1292,6 +1327,34 @@ trait MultisampleGenomicRDD[T, U <: MultisampleGenomicRDD[T, U]] extends Genomic
    * The samples who have data contained in this GenomicRDD.
    */
   val samples: Seq[Sample]
+
+  /**
+   * Replaces the sample metadata attached to the RDD.
+   *
+   * @param newSamples The new sample metadata to attach.
+   * @return A GenomicRDD with new sample metadata.
+   */
+  def replaceSamples(newSamples: Iterable[Sample]): U
+
+  /**
+   * Adds samples to the current RDD.
+   *
+   * @param samplesToAdd Zero or more samples to add.
+   * @return Returns a new RDD with samples added.
+   */
+  def addSamples(samplesToAdd: Iterable[Sample]): U = {
+    replaceSamples(samples ++ samplesToAdd)
+  }
+
+  /**
+   * Adds a single sample to the current RDD.
+   *
+   * @param sampleToAdd A single sample to add.
+   * @return Returns a new RDD with this sample added.
+   */
+  def addSample(sampleToAdd: Sample): U = {
+    addSamples(Seq(sampleToAdd))
+  }
 }
 
 /**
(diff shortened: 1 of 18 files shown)


adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -1335,6 +1398,36 @@ abstract class AvroReadGroupGenomicRDD[T <% IndexedRecord: Manifest, U <: Produc
    */
```
heuermh: Slightly unrelated, would it make sense to rename `AvroReadGroupGenomicRDD` &rarr; `AvroRecordGroupGenomicRDD`?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -1369,6 +1462,34 @@ abstract class MultisampleAvroGenomicRDD[T <% IndexedRecord: Manifest, U <: Prod
...
   val headerLines: Seq[VCFHeaderLine]
```
heuermh: This illustrates something I didn't notice before, the only subclass of `MultisampleAvroGenomicRDD` is `GenotypeRDD`.

`VariantContextRDD` extends `MultisampleGenomicRDD` and adds its own `headerLines` and `VariantRDD` extends `AvroGenomicRDD` and adds its own `headerLines`.

If `MultisampleAvroGenomicRDD` was intended to be a superclass for other data types, an association to VCF headers doesn't really make sense.

To address this, we could remove `MultisampleAvroGenomicRDD` and move `headerLines` to `GenotypeRDD`, or perhaps create something new for VCF `headerLines` that all of `VariantContextRDD`, `VariantRDD`, and `GenotypeRDD` could use.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -1292,6 +1327,34 @@ trait MultisampleGenomicRDD[T, U <: MultisampleGenomicRDD[T, U]] extends Genomic
...
+   */
+  def replaceSamples(newSamples: Seq[Sample]): U
+
+  /**
+   * Adds samples to the current RDD.
+   *
+   * @param samplesToAdd Zero or more samples to add.
+   * @return Returns a new RDD with samples added.
+   */
+  def addSamples(samplesToAdd: Seq[Sample]): U = {
```
devin-petersohn: Same note as above regarding `Iterable` vs `Seq`.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -1292,6 +1327,34 @@ trait MultisampleGenomicRDD[T, U <: MultisampleGenomicRDD[T, U]] extends Genomic
...
    */
   val samples: Seq[Sample]
+
+  /**
+   * Replaces the sample metadata attached to the RDD.
+   *
+   * @param newSamples The new sample metadata to attach.
+   * @return A GenomicRDD with new sample metadata.
+   */
+  def replaceSamples(newSamples: Seq[Sample]): U
```
devin-petersohn: Is `Iterable` better than `Seq` for the public APIs?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
fnothaft: Marking this as completed --> this was done by @heuermh in https://github.com/bigdatagenomics/adam/pull/1600/commits/79d26d54397672bf5dafafc4c30c928b80a2dc7a.

fnothaft:
@heuermh your comment RE: `MultisampleAvroGenomicRDD` got eaten by Github, so I'm copying it here:

> This illustrates something I didn't notice before, the only subclass of MultisampleAvroGenomicRDD is GenotypeRDD.
>
> VariantContextRDD extends MultisampleGenomicRDD and adds its own headerLines and VariantRDD extends AvroGenomicRDD and adds its own headerLines.
>
> If MultisampleAvroGenomicRDD was intended to be a superclass for other data types, an association to VCF headers doesn't really make sense.
>
> To address this, we could remove MultisampleAvroGenomicRDD and move headerLines to GenotypeRDD, or perhaps create something new for VCF headerLines that all of VariantContextRDD, VariantRDD, and GenotypeRDD could use.

Yeah I was noticing that as well. I think that got scoobied when we added support for non-default header lines, as the `MultisampleAvroGenomicRDD` trait existed prior to that. I agree that we should refactor this, but I'm reticent to take it on now. Would it be OK by you if we created an issue to track this and tabled it for 0.24.0?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -123,6 +125,34 @@ trait GenomicRDD[T, U <: GenomicRDD[T, U]] extends Logging {
...
+   */
+  def replaceSequences(newSequences: SequenceDictionary): U
+
+  /**
+   * Appends sequence metadata to the current RDD.
+   *
+   * @param sequencesToAdd The new sequences to append.
+   * @return Returns a new GenomicRDD with the sequences appended.
+   */
+  def addSequences(sequencesToAdd: SequenceDictionary): U = {
```
fnothaft: I mean, `SequenceDictionary` supports the `++` operator, so my preference is not to make the function variadic. Mostly, it adds a bit of error checking that we need to do (WRT being provided no sequencesToAdd).

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -123,6 +125,34 @@ trait GenomicRDD[T, U <: GenomicRDD[T, U]] extends Logging {
...
+    replaceSequences(sequences ++ sequencesToAdd)
+  }
+
+  /**
+   * Appends metadata for a single sequence to the current RDD.
+   *
+   * @param sequenceToAdd The sequence to add.
+   * @return Returns a new GenomicRDD with this sequence appended.
+   */
+  def addSequence(sequenceToAdd: SequenceRecord): U = {
```
fnothaft: My preference here is that if you have multiple records to add, you either provide them as a `SequenceDictionary` or chain multiple `addSequence` calls.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
fnothaft: Ah, good point. My preference is that we keep the type of the field `samples` as a `Seq` (because `Iterable` doesn't support indexed lookup, and it is useful to be able to easily query "which sample is sample 100?") but for these functions, `Iterable` is preferable.

fnothaft:
From a sequencing perspective, can we prioritize merging this? Once this is merged, then I'll make necessary changes to get #1545 compiling again.

devin-petersohn:
Absolutely. Are we still waiting on an update from `Seq` to `Iterable`, or merge as-is?

fnothaft:
@devin-petersohn yeah, I need to do Seq->Iterable, Python updates, and @heuermh needs to re-review.

heuermh:
Careful, two of my commits were accidentally force pushed to @fnothaft's branch, addressing comment https://github.com/bigdatagenomics/adam/pull/1600#issuecomment-314817889 above.

The additional commits move the `headerLine` stuff to `GenotypeRDD` at risk of breaking DRY and clean up the `saveMetadata` code path, fixing #1601.

fnothaft:
The Python bit needs more work to be useful, so I'm punting that to #1604/0.24.0.

fnothaft:
OK, I've cleaned up the history in @heuermh's commits and `Seq`->`Iterable`'d. This is good to go from my end.

heuermh:
@devin-petersohn Protocol question, in cases like these with one review approval and one or more review-but-not-approval (looks like yours was Comment not Request changes), should we wait for all approvals?

fnothaft:
I'd wait a short bit; @devin-petersohn is briefly OOO this PM but should be back online shortly.

devin-petersohn:
Sorry about that. Looks good to me.

PR #1606, opened 2017-07-13
[ADAM-1504] Fix transient sorting bug when sorting reads in index mode.

### LINKED ISSUE #1504: Alignment Record sort is not consistent with samtools
I have sorted platinum trio files using ADAM and saved them as a single file bam. When generating an index using samtools (samtools index NA12891.bam) I have to resort using samtools for some of the files. I get the following error:
`NO_COOR reads not in a single block at the end 0 -1`
Meaning that I have to resort.
The ADAM sorting complies with samtools for NA12878, but not NA12891 and NA12892.

fnothaft:
Resolves #1504. To handle unmapped reads—which must go at the end of the file, after all aligned reads—we assign them to a sequence index higher than the index of the highest read. However, to not put all reads at the same sequence index, we randomize using the hash code of the read name of the unmapped read. With the prior logic, reads whose names yielded negative hash codes would occasionally get indices that were valid contig indices.


fnothaft:
Thanks @heuermh! Before we merge this, let me add a unit test to catch this bug.

fnothaft:
Unit test added. This is good to go from my end.

heuermh:
Thank you, @fnothaft

PR #1608, opened 2017-07-17
[ADAM-1554] Support saving BGZF VCF output.

### LINKED ISSUE #1554: Save compressed single file VCF via HadoopBAM
See comment https://github.com/bigdatagenomics/adam/pull/1532#discussion_r119404589

fnothaft:
Resolves #1554. Additionally:

* Moves VCF file extension logic into ADAM, since Hadoop-BAM does not recognize .bgzf as a valid BGZF extension
* Refactored VCF output formats to remove need for ADAMHeaderlessVCFOutputFormat

This isn't completely done... Specifically, it produces output that bcftools thinks is good:

```
$ bcftools stats bqsr1.vcf.bgz 
# This file was produced by bcftools stats (1.3.1+htslib-1.3.1) and can be plotted using plot-vcfstats.
# The command line was:	bcftools stats  bqsr1.vcf.bgz
#
# Definition of sets:
# ID	[2]id	[3]tab-separated file names
ID	0	bqsr1.vcf.bgz
# SN, Summary numbers:
# SN	[2]id	[3]key	[4]value
SN	0	number of samples:	0
SN	0	number of records:	681
SN	0	number of no-ALTs:	0
SN	0	number of SNPs:	681
SN	0	number of MNPs:	0
SN	0	number of indels:	0
SN	0	number of others:	0
SN	0	number of multiallelic sites:	0
SN	0	number of multiallelic SNP sites:	0
```

But, that fails to load in properly when read back from ADAM:

```
scala> sc.loadVcf("bqsr1.vcf.bgz").rdd.count
res1: Long = 0
```

I'm still debugging this, but was wondering if anyone else had thoughts?

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/variant/VariantContextRDD.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/variant/VariantContextRDD.scala
@@ -18,9 +18,12 @@
 package org.bdgenomics.adam.rdd.variant
 
 import htsjdk.samtools.ValidationStringency
+import htsjdk.samtools.util.BlockCompressedOutputStream
 import htsjdk.variant.vcf.{ VCFHeader, VCFHeaderLine }
-import org.apache.hadoop.io.LongWritable
 import org.apache.hadoop.fs.Path
+import org.apache.hadoop.io.LongWritable
+import org.apache.hadoop.io.compress.CompressionCodec
+import org.apache.hadoop.mapreduce.lib.output.FileOutputFormat
 import org.apache.spark.rdd.RDD
 import org.bdgenomics.adam.converters.{
   DefaultHeaderLines,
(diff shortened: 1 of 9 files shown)


adam-core/src/main/scala/org/bdgenomics/adam/rdd/variant/VariantContextRDD.scala
```diff
@@ -227,8 +237,20 @@ case class VariantContextRDD(rdd: RDD[VariantContext],
...
+    val vcfFormat = VCFFormat.VCF
+    val isBgzip = FileExtensions.isGzip(filePath)
+    if (!FileExtensions.isVcfExt(filePath)) {
+      val msg = "Saw non-VCF extension for file %s. Extensions supported are .vcf<.gz, .bgz>"
+        .format(filePath)
+
+      if (stringency == ValidationStringency.STRICT) {
+        throw new IllegalArgumentException(msg)
+      } else {
+        log.warn(msg)
```
heuermh: Does it make sense to stringency this?  If we warn here, does the file still get written correctly?

fnothaft: I mean, you'll get a valid VCF file, but the file won't be BGZF'ed. That said, I'd be fine not stringencying this.

fnothaft:
OK, sorted out the issue; I was writing a BGZF EOF terminator at the end of the header. htslib is OK with this, htsjdk isn't. Also addressed @heuermh's comment RE validation stringency.

heuermh:
Thank you, @fnothaft

PR #1619, opened 2017-07-21
Update load method docs in Python and R.

heuermh:


heuermh:
Jenkins, retest this please.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2749,7 +2749,7 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
   /**
    * Load variants into a VariantRDD.
    *
-   * If the path name has a .vcf/.vcf.gz/.vcf.bgzf/.vcf.bgz extension, load as VCF format.
```
fnothaft: Due to Hadoop codec limitations, we don't support `.bgzf` on load. We support it on write since we choose the output codec, but there's no `.bgzf` codec and the codecs are autodetected on read from the file extensions.

heuermh: Yep, that is why I removed it.  Only `.bgz` is supported on load, right?

heuermh:
Jenkins, retest this please.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
fnothaft: Oh sorry about that, I misread the diff. `.bgz` and `.gz` are both supported.

fnothaft:
Merged! Thanks @heuermh!

heuermh:
Thank you!

PR #1620, opened 2017-07-21
add hive style partitioning for contigName

jpdna:
The relevant changes to review are in file:
`AlignmentRecordRDD.scala` the other changes are caused by our scripts to change to Spark2/scala 2.11

This PR implements Hive-style partitioning by `contigName` for AlignmentRecords when written to parquet through the Dataset/SparkSQL write path.  As described at [here](https://spark.apache.org/docs/2.1.1/sql-programming-guide.html#partition-discovery)

The output directory when the dataset is saved to Parquet looks like the following:
```
_SUCCESS
_common_metadata
_metadata
_rgdict.avro
_seqdict.avro
contigName=1
contigName=2
...
```
where the Parquet files for each contigName are written inside the `contigName=N` directories.

Note: in the future we will add another layer of hierarchy with 10 megabase bins within each chromosome as further subdirs

As per discussion in bigdatagenomics#651
such binning should allow a more efficient predicate pushdown of range queries than we may currently get from Parquet.

Below is described how to write and read back this data in the spark-shell,
and an error which occurs currently when reading back directly as an RDD is discussed.

Load data from SAM file, convert to Dataset, and save as Hive-partitioned parquet
```
import org.bdgenomics.adam.sql.{ AlignmentRecord => AlignmentRecordProduct }

val rdd = sc.loadAlignments("../adam/adam-core/src/test/resources/multi_chr.sam")
val x = rdd.transformDataset(ds => {  
  import ds.sqlContext.implicits._
      val df = ds.toDF()
       df.as[AlignmentRecordProduct]
      })

// save as Hive-style paritioned data by contigName
x.saveAsParquet("test_hivePartitions_1")
```

Read back in from disk as a Dataset
```
val read = sc.loadAlignments("test_hivePartitions_1")
read1: org.bdgenomics.adam.rdd.read.AlignmentRecordRDD =
ParquetUnboundAlignmentRecordRDD(org.apache.spark.SparkContext@2db4a84a,test_hivePartitions_1,SequenceDictionary{

// view fields from first record as a dataset
scala> read.dataset.collect()(0)
res38: org.bdgenomics.adam.sql.AlignmentRecord = org.bdgenomics.adam.sql.AlignmentRecord@64dba8f1

scala> read.dataset.collect()(0).contigName.get
res40: String = 1

scala> read.dataset.collect()(0).start.get
res41: Long = 26472783
```

Read back in directly as an RDD (this results incorrect contigName=null)
```
//examine the type
scala> read.rdd
res45: org.apache.spark.rdd.RDD[org.bdgenomics.formats.avro.AlignmentRecord] = MapPartitionsRDD[51] at map at ADAMContext.scala:126

// print the first record when read as an rdd
read.rdd.collect()(0)
res44: org.bdgenomics.formats.avro.AlignmentRecord = {"readInFragment": 0, "contigName": null, "start": 26472783, "oldPosition": null, "end": 26472858, "mapq": 60, "readName": "simread:1:26472783:false", "sequence": "GTATAAGAGCAGCCTTATTCCTATTTATAATCAGGGTGAAACACCTGTGCCAATGCCAAGACAGGGGTGCCAAGA", "qual": null, "cigar": "75M", "oldCigar": null, "basesTrimmedFromStart": 0, "basesTrimmedFromEnd": 0, "readPaired": false, "properPair": false, "readMapped": true, "mateMapped": false, "failedVendorQualityChecks": false, "duplicateRead": false, "readNegativeStrand": true, "mateNegativeStrand": false, "primaryAlignment": true, "secondaryAlignment": false, "supplementaryAlignment": false, "mismatchingPositions": null, "origQual": null, "attributes": "XS:i:0\tAS:i:75\tNM:i:0", "recordGroupName": null…
```
PROBLEM:  Note that above when read as an RDD the contigName field is NULL.  This is incorrect, the contigName=1 here.

We can get the correct `RDD[org.bdgenomics.formats.avro.AlignmentRecord]` though another way, by reading as a dataset and then converting to an RDD:
```
// read as dataset, convert to rdd of sql type then map to Avro bdg AlignmentRecord
val rdd_try2 = read.dataset.rdd.map(_.toAvro)
rdd_try2: org.apache.spark.rdd.RDD[org.bdgenomics.formats.avro.AlignmentRecord] = MapPartitionsRDD[58] at map at <console>:30

// checking the first record now, we find that contigName is correct
rdd_try2.collect()(0)
res46: org.bdgenomics.formats.avro.AlignmentRecord = {"readInFragment": 0, "contigName": "1", "start": 26472783, "oldPosition": null, "end": 26472858, "mapq": 60, "readName": "simread:1:26472783:false", "sequence": "GTATAAGAGCAGCCTTATTCCTATTTATAATCAGGGTGAAACACCTGTGCCAATGCCAAGACAGGGGTGCCAAGA", "qual": null, "cigar": "75M", "oldCigar": null, "basesTrimmedFromStart": 0, "basesTrimmedFromEnd": 0, "readPaired": false, "properPair": false, "readMapped": true, "mateMapped": false, "failedVendorQualityChecks": false, "duplicateRead": false, "readNegativeStrand": true, "mateNegativeStrand": false, "primaryAlignment": true, "secondaryAlignment": false, "supplementaryAlignment": false, "mismatchingPositions": null, "origQual": null, "attributes": "XS:i:0\tAS:i:75\tNM:i:0", "recordGroupName": null,..

```

Conclusion:

When using Hive style partitioning as implemented in this PR, the direct RDD read path
sc.loadAlignments("myParitionedSparkSQLParquetDir").rdd
will result in contigName erroneously being null because the direct RDD reading code is not handling the Hive-style partitioning correctly.  My guess is that the RDD read-path just isn't fully partition-aware and/or when partitioned parquet files are written by Spark-SQL they may null out the info that is redundant in the partitioning.

The cost of having to read Hive-partitioned parquet data in as a Dataset but then convert to an RDD seems like it may be a reasonable price to pay if indeed we find enough other benefits to the Hive-style partitioning, but we need to then disallow the direct rdd read-path so that users don't get this error of contigName=null


jpdna:
I've loaded a chr20 BAM file for a sample from 1000 genomes.  Hive partition dataset output to parquet with  contig/1MB  bins  writing and reading seems to be working fine.  

Each 1 megabase bin (1..60) under contig 20 has one parquet file in it.

Example queries which are working below:
```
val read = sc.loadAlignments("hg00096_chr20_1")
val df = read.dataset
df.filter("contigName = '20'").filter("posBin = '39'").count
```
```
df.createOrReplaceTempView("genomedata")
val sqlResult = spark.sql("SELECT * FROM genomedata where contigName='20' and posBin>39 and posBin<50")
sqlResult.count
```
Next, I will do some performance testing to  see if retrieval of various range sizes is faster or slower through the SQL path and the contigName/posBin columns compared  to: 
a) non hive partitioned dataset written to parquet 
b) non-hive partitioned parquet written from RDD

@fnothaft - note. the problem with posBin column that I thought I'd encountered earlier  appears to not be a problem after all, I think previously I was not actually using the dataset I had added that posBin column to - happily its seeming to just work now.

jpdna:
Some preliminary results:
On the bdg cluster, using 20 executors, reading a 14GB low cov whole genome BAM file:
Filtering this whole genome dataset  for a 1mb region (chr4:31000000-32000000) and then  counting or re-writing to parquet takes 25 seconds using the existing non-partitioned parquet ADAM files, but when reading a Hive-style partitioned parquet files, takes just 1 second.

Parsing the original BAM file and writing to paritioned parquet takes 2.5 minuets, while the existing non-paritioned parquet write takes 3.5 minutes, so surprising to me a bit of gain on write too.

The improvement in region filtering time for Hive-style partitioned data  I think can be attributed to eliminating the  scan time to touch and reject partitions based on min max  in the original parquet method.   Increased number of executors relative to the number of partitions in the of input dataset should reduce difference between methods in the limit of where this scan becomes fully parallel.   Single node or small clusters should see the most gains from the Hive partition strategy, though there seems to be little downside even on a larger cluster.

I'm writing up in detail the test's so far, and adding more.
In addition to counting and re-writing to disk tasks, I'll add k-mer counting.

@akmorrow13 - I look forward to seeing if this will be useful for retrieving region slices with lower latency for visualization in Mango.

jpdna:
can someone point out to me in the jenkins output why my tests are failing? - and suggest fixes.   This is successfully building with tests when I run `mvn clean package` on my local machine.

fnothaft:
@jpdna as an aside, I was thinking about the `null` issue when loading as Parquet into RDD-land and did a small amount of benchmarking. If we bind to a dataset and then do `.rdd` to convert back into an RDD, we're only about 20% slower (for a really trivial query). I think this overhead is pretty much in the noise, so we could bind Hive-partitioned files to Dataset on load and that'd resolve the `null` issue.

jpdna:
> bind Hive-partitioned files to Dataset on load

In that case, do we need to have a flag or otherwise explicitly detect when loading from Hive-Partitioned files?  Right now when reading, the partitioning is just auto-detected I believe.

fnothaft:
>> bind Hive-partitioned files to Dataset on load
>
> In that case, do we need to have a flag or otherwise explicitly detect when loading from Hive-Partitioned files? Right now when reading, the partitioning is just auto-detected I believe.

Yeah, my thought is that we'd either touch some file (e.g., `_hivePartitioned`) on write and then look for that file on load or we'd try to identify the partitioning scheme, whichever is the right combo of easy/robust.

fnothaft:
Jenkins, retest this please.

fnothaft:
@jpdna it was a jenkins glitch, I've just kicked off another test.

jpdna:
> look for that file on load or we'd try to identify the partitioning scheme

I've implemented this `_hivePartitioned` marker file so that now  `sc.loadAlignments("mypath")` when that file is present, returns a `DatasetBoundAlignmentRecordRDD` such that we prevent the possibility of the `null` cotnigName error discussed above when attempting to read as an rdd directly

jpdna:
Ready for initial review by others.

I added a parameter to `saveAsParquet` to enable/disable hive-style partitioning on write, and to control the genomic region bin size of  subdirs that parquet files are written into.  
```
rdd.saveAsParquet(outputPath, enableHivePartitioning = true, hivePartitioningGenomicBinSize = 1000000)
```
This is nice because in absence of explicitly enabling hivePartitioning here, behavior should not be altered.

As discussed above, previous commit includes a `_hivePartiioned` marker file written with parquet files and when detected on read automatically binds to a Dataset, ameliorating earlier issue when trying to read as rdd.

todo:
- I am suspicious as to why Jenkins says this passes Spark 1.6 when I believe I found that partitioning only worked with Spark 2.1.x, so need to make sure the tests are properly testing all the hive partitioning features.
- add tests to cover all the types `fragment`, etc.
- formally report the performance metrics in reading and writing hive-partitioned vs. non-partitioned dataset and rdd

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -1777,8 +1777,15 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
 
     (optPredicate, optProjection) match {
       case (None, None) => {
-        ParquetUnboundAlignmentRecordRDD(sc, pathName, sd, rgd, pgs)
+        val hiveFlagFile = new File(pathName, "_hivePartitioned")
```
fnothaft: This will only work for a file that is stored on a "local" (read: mounted POSIX compatible) file system. I'd suggest factoring this out into a function:

```
private def checkHiveFlag(fileName: String): Boolean = {
  val hivePath = new Path(pathName, "_hivePartitioned")
  val fs = hivePath.getFileSystem(sc.hadoopConfiguration)
  fs.exists(hivePath)
}
```

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -1848,7 +1848,9 @@ abstract class AvroGenomicRDD[T <% IndexedRecord: Manifest, U <: Product, V <: A
     blockSize: Int = 128 * 1024 * 1024,
     pageSize: Int = 1 * 1024 * 1024,
     compressCodec: CompressionCodecName = CompressionCodecName.GZIP,
-    disableDictionaryEncoding: Boolean = false) {
+    disableDictionaryEncoding: Boolean = false,
+    enableHivePartitioning: Boolean = false,
+    hivePartitioningGenomicBinSize: Int = 1000000) {
```
fnothaft: I would roll `enableHivePartitioning` and `hivePartitioningGenomicBinSize` into one parameter `optHivePartitioningBinSize`

adam-core/src/main/scala/org/bdgenomics/adam/rdd/feature/FeatureRDD.scala
```diff
@@ -284,14 +285,36 @@ case class DatasetBoundFeatureRDD private[rdd] (
...
+      val df = dataset.toDF()
+
+      df.withColumn("posBin", floor(df("start") / genomicBinSize))
+        .write
+        .partitionBy("contigName", "posBin")
+        .format("parquet")
+        .option("spark.sql.parquet.compression.codec", compressCodec.toString.toLowerCase())
+        .save(filePath)
+
+      val hiveFlagFile = new File(filePath, "_hivePartitioned")
```
fnothaft: These will also need to be rewritten to use either the Hadoop FileSystem API or the java.nio libraries.

jpdna:
> is to factor the saveAsParquet code out so that there's a single "save as binned" method and a "save unbinned"

@fnothaft can you elaborate a bit here, are you suggesting that instead of calling `saveAsParquet` we'd call `saveAsParquetUnbinned` or `saveAsParquetBinned` ?

heuermh:
As a first high level review, I think all references to "Hive" should be dropped.

Then I'm in agreement with @fnothaft in that boolean flags make babies cry.  I would prefer methods `saveAsParquet` and `saveAsBinnedParquet` or `saveAsPartitionedParquet`.

Should the marker file be called `_binnedByStartPosition` or `_partitionedByStartPosition`?  Are there any other binned `partitionBy` fields we would want to support?

I assume any new command line options will conflict with `-coalesce` and `-repartition`?

jpdna:
> Finally, is it possible to implement this as an external Maven module? Feels invasive to have to modify/extend all of the *RDD classes.

As best as I can tell - because I cut and pasted it anyhow, the code in *RDD classes implementing the over-ridden function saveAsParquet is the same in every *RDD class of ours, and I am wondering if the implementation can be factored out into the base class or elsewhere to be more DRY.  If so maybe it won't be so invasive.

I am not sure how to go about making this functionality a separate module - and if it works well I'd like to think partitioning parquet like this may become the standard mode when saving from a Dataset, as so far I don't see much of a performance penalty -  I'll proceed with the changes suggested above still within the current classes unless others suggest another way forward.

fnothaft:
> Finally, is it possible to implement this as an external Maven module? Feels invasive to have to modify/extend all of the *RDD classes.

-1. This is very reasonable functionality to have in the GenomicRDDs. Just to clarify in case there's a misunderstanding, this functionality isn't _only usable with Hive_, the "Hive" nomenclature comes from the fact that Hive was the tool that introduced this type of partitioning scheme.

> and if it works well I'd like to think partitioning parquet like this may become the standard mode when saving from a Dataset

+1!

heuermh:
> Just to clarify in case there's a misunderstanding, this functionality isn't only usable with Hive, the "Hive" nomenclature comes from the fact that Hive was the tool that introduced this type of partitioning scheme.

Agreed, in the Spark docs it is referred to as, "[an] approach used in systems like Hive."  Thus I don't see a need to mention Hive anywhere in field or method or file names.

> ...may become the standard mode when saving from a Dataset

Feels like there may be pathological cases where this falls down, say lots of unaligned reads, but perhaps it will be no worse than the default.

jpdna:
update wrt last push:
- fixed flag file writing to work with HDFS
- Now there are specific `saveAsPartitionedParquet` and `loadPartitionedParquet` functions.  The `load` functions replace what we previously an attempt to keep a single load function and vary behavior based on flag file.  The explicit approach is cleaner as a) the optional Parquet predicates and Filters parmeters will never be meaningful for the `Datasets` loaded from partitioned parquet because those filters/predicates will be applied via SQL/dataset syntax in subsequent steps after `loadParitionedParquet` returns its Dataset handle.
- All reference to `Hive` have been removed
- There is only a single implementation of `saveAsPartitionedParquet` appearing in the base class GenomicRDD.   Need to evaluate if this creates any problems for any subclases for which this function cannot work (if any)

todo:
- updates tests
- further collection of performance metrics
- consider some wrapper functions or other utilities/demos of integration of the partitioned/dataset reed path with algorithms

jpdna:
Ready for some further review here

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2904,4 +2970,16 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
       loadParquetFragments(pathName, optPredicate = optPredicate, optProjection = optProjection)
     }
   }
+
+  def writePartitionedParquetFlag(filePath: String): Boolean = {
+    val path = new Path(filePath, "_isPartitionedByStartPos")
+    val fs = path.getFileSystem(sc.hadoopConfiguration)
+    fs.createNewFile(path)
+  }
+  def checkPartitionedParquetFlag(filePath: String): Boolean = {
```
fnothaft: Nit: should have whitespace before this line.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2904,4 +2970,16 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+    val path = new Path(filePath, "_isPartitionedByStartPos")
+    val fs = path.getFileSystem(sc.hadoopConfiguration)
+    fs.exists(path)
+  }
+
```
fnothaft: Nit: shouldn't have whitespace here.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -1888,4 +1889,29 @@ abstract class AvroGenomicRDD[T <% IndexedRecord: Manifest, U <: Product, V <: A
...
+
+  def saveAsPartitionedParquet(filePath: String,
+                               blockSize: Int = 128 * 1024 * 1024,
+                               pageSize: Int = 1 * 1024 * 1024,
+                               compressCodec: CompressionCodecName = CompressionCodecName.GZIP,
+                               disableDictionaryEncoding: Boolean = false,
+                               partitionSize: Int = 1000000) {
+    log.warn("Saving directly as Hive-partitioned Parquet from SQL. " +
+      "Options other than compression codec are ignored.")
+    import org.apache.spark.sql.functions._
```
fnothaft: I would prefer moving this import up with all the other imports.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -1888,4 +1889,29 @@ abstract class AvroGenomicRDD[T <% IndexedRecord: Manifest, U <: Product, V <: A
   def saveAsParquet(filePath: java.lang.String) {
     saveAsParquet(new JavaSaveArgs(filePath))
   }
+
+  def saveAsPartitionedParquet(filePath: String,
+                               blockSize: Int = 128 * 1024 * 1024,
```
fnothaft: We should drop all the options other than compression codec and partition size, since they are ignored.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -1888,4 +1889,29 @@ abstract class AvroGenomicRDD[T <% IndexedRecord: Manifest, U <: Product, V <: A
...
+                               pageSize: Int = 1 * 1024 * 1024,
+                               compressCodec: CompressionCodecName = CompressionCodecName.GZIP,
+                               disableDictionaryEncoding: Boolean = false,
+                               partitionSize: Int = 1000000) {
+    log.warn("Saving directly as Hive-partitioned Parquet from SQL. " +
+      "Options other than compression codec are ignored.")
+    import org.apache.spark.sql.functions._
+    val genomicBinSize = partitionSize
+    val df = dataset.toDF()
```
fnothaft: You should be able to just call `toDF()` here.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -1888,4 +1889,29 @@ abstract class AvroGenomicRDD[T <% IndexedRecord: Manifest, U <: Product, V <: A
...
+  def saveAsPartitionedParquet(filePath: String,
+                               blockSize: Int = 128 * 1024 * 1024,
+                               pageSize: Int = 1 * 1024 * 1024,
+                               compressCodec: CompressionCodecName = CompressionCodecName.GZIP,
+                               disableDictionaryEncoding: Boolean = false,
+                               partitionSize: Int = 1000000) {
+    log.warn("Saving directly as Hive-partitioned Parquet from SQL. " +
+      "Options other than compression codec are ignored.")
+    import org.apache.spark.sql.functions._
+    val genomicBinSize = partitionSize
```
fnothaft: I wouldn't dupe the variable; I would prefer `s/genomicBinSize/partitionSize/g` throughout this function.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -1888,4 +1889,29 @@ abstract class AvroGenomicRDD[T <% IndexedRecord: Manifest, U <: Product, V <: A
...
+    val df = dataset.toDF()
+
+    df.withColumn("posBin", floor(df("start") / partitionSize))
+      .write
+      .partitionBy("contigName", "posBin")
+      .format("parquet")
+      .option("spark.sql.parquet.compression.codec", compressCodec.toString.toLowerCase())
+      .save(filePath)
+
+    df.sqlContext.sparkContext.writePartitionedParquetFlag(filePath)
```
fnothaft: `df.sqlContext.sparkContext` -> `rdd.context`.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -1888,4 +1889,29 @@ abstract class AvroGenomicRDD[T <% IndexedRecord: Manifest, U <: Product, V <: A
...
+    saveMetadata(filePath)
```
fnothaft: Nit: whitespace before this line, no whitespace afterwards.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -1789,6 +1789,24 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
     }
   }
 
+  def loadPartitionedParquetAlignments(pathName: String): AlignmentRecordRDD = {
+
+    if (!checkPartitionedParquetFlag(pathName)) {
```
fnothaft: You can replace this if statement with:

```
require(checkPartitionedParquetFlag(pathName),
  "Error: input Parquet files are not Partitioned")
```

fnothaft:
Hi @jpdna! Do you think you can have this cleaned up by the end of the week? If not, I will take this over. @heuermh and I would like to get this into 0.23.0.

jpdna:
Yes I can make progress this week, I'll have an update tomorrow.

One question below:
The reason for the separate `loadPartitionedParquetAlignments` you ask about above is that currently the optional predicate and projection parameters to `loadParquetAlignments` are not automatically translated into spark-sql and we have no way to use them directly that I see when reading via the spark-sql/dataset codepath, so the predicate and projection parameters would be silently ignored.   I was thinking of ways to translate them to spark-sql in general or way, or or more limited purposes, provide a function that takes a `ReferenceRegion` or other specific filters or turns projected fields into the fields selected via sql.   If you see another way to do this or to retain only one load function without this conundrum let me know.

jpdna:
I'm having trouble figuring out if and how to address the three failures noted by Jenkins after my last push.   Any thoughts? and can you point me to where in the output/logs you looked?  Thanks!

fnothaft:
We made a CI change for #1722. The PRB will not pass until #1722 merges. Also, there was an S3 outage that was impacting CI earlier today.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -1789,6 +1789,24 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
     }
   }
 
+  def loadPartitionedParquetAlignments(pathName: String): AlignmentRecordRDD = {
+
+    require(checkPartitionedParquetFlag(pathName),
+      "Error: input Parquet files are not Partitioned")
```
fnothaft: Drop the `Error:`.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2105,6 +2123,21 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
     }
   }
 
+  def loadPartitionedParquetGenotypes(pathName: String): GenotypeRDD = {
+    if (!checkPartitionedParquetFlag(pathName)) {
+      throw new IllegalArgumentException("Error: input Parquet files are not Partitioned")
```
fnothaft: Ditto RE: `error`

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2138,6 +2171,20 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
     }
   }
 
+  def loadPartitionedParquetVariants(pathName: String): VariantRDD = {
+    if (!checkPartitionedParquetFlag(pathName)) {
+      throw new IllegalArgumentException("Error: input Parquet files are not Partitioned")
```
fnothaft: Ditto RE `Error`

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2454,6 +2501,15 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
     }
   }
 
+  def loadPartitionedParquetFeatures(pathName: String): FeatureRDD = {
+    if (!checkPartitionedParquetFlag(pathName)) {
+      throw new IllegalArgumentException("Error: input Parquet files are not Partitioned")
```
fnothaft: Ditto RE: `Error`

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2486,6 +2542,16 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
     }
   }
 
+  def loadPartitionedParquetFragments(pathName: String): NucleotideContigFragmentRDD = {
+    if (!checkPartitionedParquetFlag(pathName)) {
+      throw new IllegalArgumentException("Error: input Parquet files are not Partitioned")
```
fnothaft: Ditto RE `Error`

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -1888,4 +1890,23 @@ abstract class AvroGenomicRDD[T <% IndexedRecord: Manifest, U <: Product, V <: A
   def saveAsParquet(filePath: java.lang.String) {
     saveAsParquet(new JavaSaveArgs(filePath))
   }
+
+  def saveAsPartitionedParquet(filePath: String,
+                               compressCodec: CompressionCodecName = CompressionCodecName.GZIP,
+                               partitionSize: Int = 1000000) {
+    log.warn("Saving directly as Hive-partitioned Parquet from SQL. " +
+      "Options other than compression codec are ignored.")
+    import org.apache.spark.sql.functions._
```
fnothaft: Move to preamble.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -1789,6 +1789,24 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+    val sd = loadAvroSequenceDictionary(pathName)
+
+    // convert avro to sequence dictionary
+    val rgd = loadAvroRecordGroupDictionary(pathName)
+
+    // load processing step descriptions
+    val pgs = loadAvroPrograms(pathName)
+    val reads = ParquetUnboundAlignmentRecordRDD(sc, pathName, sd, rgd, pgs)
+    DatasetBoundAlignmentRecordRDD(reads.dataset, reads.sequences, reads.recordGroups, reads.processingSteps)
+
```
fnothaft: Drop whitespace.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2138,6 +2171,20 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+      throw new IllegalArgumentException("Error: input Parquet files are not Partitioned")
+    }
+    val sd = loadAvroSequenceDictionary(pathName)
+
+    // load header lines
+    val headers = loadHeaderLines(pathName)
+
+    val variants = ParquetUnboundVariantRDD(sc, pathName, sd, headers)
+    DatasetBoundVariantRDD(variants.dataset, variants.sequences, headers)
+
```
fnothaft: Drop whitespace

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2486,6 +2542,16 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
   }
 
+  def loadPartitionedParquetFragments(pathName: String): NucleotideContigFragmentRDD = {
+    if (!checkPartitionedParquetFlag(pathName)) {
+      throw new IllegalArgumentException("Error: input Parquet files are not Partitioned")
+    }
+    val sd = loadAvroSequenceDictionary(pathName)
+    val nucleotideContigFragments = ParquetUnboundNucleotideContigFragmentRDD(sc, pathName, sd)
+    DatasetBoundNucleotideContigFragmentRDD(nucleotideContigFragments.dataset, nucleotideContigFragments.sequences)
+
```
fnothaft: Drop whitespace

heuermh:
@jpdna I think you'll need to rebase to pick up build related changes and resolve conflicts.

fnothaft:
+1 @heuermh

jpdna:
Hi @heuermh or @fnothaft  , I need help with something mysterious (to me) happening in my effort to rebase against master at  1302e07 
If I do
```
git clone https://github.com/jpdna/adam.git
```
As I expect I see my `master` is at 1302e07 because I just synced it with bdg repo.
Note, 1302e07 is on scala.version 2.10.6 in the pom.xml

next I checkout my PR branch
```
git checkout hive_partitions
```
which is at 78fcb1a89e46
and note, it is at scala.version 2.10.6

now if I try to rebase
```
git rebase origin/master
```
there are some conflicts.
Which I would work on resolving, but first, I am weirded out why
when looking at the the conflicts in my pom.xml
I see that suddenly in my middle pane of the mergtool I am seeting
```
<scala.version>2.11.8</scala.version>
```
I don't get where the scala 2.11.8 is coming from as I don't have it in my `hive_partitions` branch at 78fcb1a89 and its not in my origin master at 1302e07

What am I missing?

heuermh:
Right, that seems strange.  [Git blame](https://github.com/bigdatagenomics/adam/blame/master/pom.xml) shows that line (pom.xml, line 22) hasn't been modified in more than a year.

Perhaps somewhere in your intermediate commits it switches to Scala 2.11 and then back to Scala 2.10?  You may want to try going all the way through the rebase and see if you end up with what you expect.

jpdna:
OK - yes rebase did work, and squashed.

jpdna:
This branch will only work or compile on scala2.11, spark2,  so for purpose of this PR should it be moved to scala2.11/spark2  as it currently is, or do you want me to move it back?

fnothaft:
> or do you want me to move it back?

@jpdna please move it back. Thanks!

jpdna:
If you know a better way to setup a dataset api query using dataset columns than building up this string like here https://github.com/jpdna/adam/blob/217183e397aa97a67043cec8fe155dd3df2e19bf/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala#L1808
let me know.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2105,6 +2139,21 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
     }
   }
 
+  def loadPartitionedParquetGenotypes(pathName: String): GenotypeRDD = {
```
akmorrow13: Is there a reason there is no regions parameter here?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2138,6 +2187,35 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
     }
   }
 
+  def loadPartitionedParquetVariants(pathName: String, regions: Option[Iterable[ReferenceRegion]] = None, partitionSize: Int = 1000000): VariantRDD = {
```
akmorrow13: Is there any interest in combining this functionality into the existing load functions? It may be nice so the user does not have to know whether the files are partitioned when loading.

jpdna: I looked at that earlier, using a flag file to detect if the input was partitioned or not, but the trouble is that the existing load functions have a predicate and projection parameter that we don't right now have a way to directly translate into the something spark-sql/dataset can use.

jpdna:
I've done some single node benchmarking using a version of Mango that relies on this PR and uses a partitioned alignment dataset.
Random access of a 1000 base pair region:
unpartitioned (current) - adam files   8-9 seconds
partitioned ADAM file (this PR)   1.5 - 2.5 seconds

https://docs.google.com/presentation/d/1p6nA_vhydW2J2O7iFTJbsohbC_FAFEhpqfmzNtb8jRU/edit?usp=sharing

akmorrow13:
Awesome @jpdna! Can you make a PR in Mango so we can track these numbers there?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -18,15 +18,10 @@
 package org.bdgenomics.adam.rdd
 
 import java.io.{ File, FileNotFoundException, InputStream }
+
```
fnothaft: Can you remove this whitespace?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -18,15 +18,10 @@
...
 import htsjdk.samtools.{ SAMFileHeader, SAMProgramRecord, ValidationStringency }
 import htsjdk.samtools.util.Locatable
-import htsjdk.variant.vcf.{
-  VCFHeader,
-  VCFCompoundHeaderLine,
-  VCFFormatHeaderLine,
-  VCFHeaderLine,
-  VCFInfoHeaderLine
-}
+import htsjdk.variant.vcf.{ VCFCompoundHeaderLine, VCFFormatHeaderLine, VCFHeader, VCFHeaderLine, VCFInfoHeaderLine }
```
fnothaft: Can you undo this? We break at 80 chars or 4 items.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -42,72 +37,30 @@ import org.apache.parquet.hadoop.util.ContextUtil
...
-  Projection
-}
-import org.bdgenomics.adam.rdd.contig.{
-  DatasetBoundNucleotideContigFragmentRDD,
-  NucleotideContigFragmentRDD,
-  ParquetUnboundNucleotideContigFragmentRDD,
-  RDDBoundNucleotideContigFragmentRDD
-}
+import org.bdgenomics.adam.projections.{ FeatureField, Projection }
+import org.bdgenomics.adam.rdd.contig.{ DatasetBoundNucleotideContigFragmentRDD, NucleotideContigFragmentRDD, ParquetUnboundNucleotideContigFragmentRDD, RDDBoundNucleotideContigFragmentRDD }
```
fnothaft: Can you break this line? We indent at 80 chars/4 imports.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -42,72 +37,30 @@ import org.apache.parquet.hadoop.util.ContextUtil
...
-}
-import org.bdgenomics.adam.rdd.read.{
-  AlignmentRecordRDD,
-  DatasetBoundAlignmentRecordRDD,
-  RepairPartitions,
-  ParquetUnboundAlignmentRecordRDD,
-  RDDBoundAlignmentRecordRDD
-}
+import org.bdgenomics.adam.rdd.fragment.{ DatasetBoundFragmentRDD, FragmentRDD, ParquetUnboundFragmentRDD, RDDBoundFragmentRDD }
+import org.bdgenomics.adam.rdd.read.{ AlignmentRecordRDD, DatasetBoundAlignmentRecordRDD, ParquetUnboundAlignmentRecordRDD, RDDBoundAlignmentRecordRDD, RepairPartitions }
```
fnothaft: Can you break this line? We indent at 80 chars/4 imports.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -42,72 +37,30 @@ import org.apache.parquet.hadoop.util.ContextUtil
...
 import org.bdgenomics.adam.rich.RichAlignmentRecord
-import org.bdgenomics.adam.sql.{
-  AlignmentRecord => AlignmentRecordProduct,
-  Feature => FeatureProduct,
-  Fragment => FragmentProduct,
-  Genotype => GenotypeProduct,
-  NucleotideContigFragment => NucleotideContigFragmentProduct,
-  Variant => VariantProduct
-}
+import org.bdgenomics.adam.sql.{ AlignmentRecord => AlignmentRecordProduct, Feature => FeatureProduct, Fragment => FragmentProduct, Genotype => GenotypeProduct, NucleotideContigFragment => NucleotideContigFragmentProduct, Variant => VariantProduct }
```
fnothaft: Can you break this line? We indent at 80 chars/4 imports.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -42,72 +37,30 @@ import org.apache.parquet.hadoop.util.ContextUtil
...
-  Fragment,
-  Genotype,
-  NucleotideContigFragment,
-  ProcessingStep,
-  RecordGroup => RecordGroupMetadata,
-  Sample,
-  Variant
-}
+import org.bdgenomics.adam.util.{ GenomeFileReader, ReferenceContigMap, ReferenceFile, SequenceDictionaryReader, TwoBitFile }
+import org.bdgenomics.formats.avro.{ AlignmentRecord, Contig, Feature, Fragment, Genotype, NucleotideContigFragment, ProcessingStep, Sample, Variant, RecordGroup => RecordGroupMetadata }
```
fnothaft: Can you break this line? We indent at 80 chars/4 imports.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -1789,6 +1742,27 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
     }
   }
 
+  def loadPartitionedParquetAlignments(pathName: String, regions: Option[Iterable[ReferenceRegion]] = None, partitionSize: Int = 1000000): AlignmentRecordRDD = {
```
fnothaft: I don't think `partitionSize` is used, no?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2105,6 +2079,28 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
     }
   }
 
+  def loadPartitionedParquetGenotypes(pathName: String, regions: Option[Iterable[ReferenceRegion]] = None, partitionSize: Int = 1000000): GenotypeRDD = {
```
fnothaft: I don't think `partitionSize` is used, no?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
fnothaft: I think its fine if the predicate/projection parameters can't be translated over. In the main `loadParquet*` functions, what I would do is:

- Check for the partitionedParquetFlag
- If the flag is set, check if predicate/projection are provided. If projection is provided, log a warning. If predicate is provided, throw an exception.
- If we didn't throw an exception, call `loadPartitionedParquet*`

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2454,6 +2469,17 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
     }
   }
 
+  def loadPartitionedParquetFeatures(pathName: String): FeatureRDD = {
```
fnothaft: This is missing `regions`, no?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2454,6 +2469,17 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+    if (!checkPartitionedParquetFlag(pathName)) {
+      throw new IllegalArgumentException("input Parquet files are not Partitioned")
+    }
+    val sd = loadAvroSequenceDictionary(pathName)
+    val features = ParquetUnboundFeatureRDD(sc, pathName, sd)
+
+    DatasetBoundFeatureRDD(features.dataset, features.sequences)
+
```
fnothaft: Extra whitespace.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2486,6 +2512,15 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
     }
   }
 
+  def loadPartitionedParquetFragments(pathName: String): NucleotideContigFragmentRDD = {
+    if (!checkPartitionedParquetFlag(pathName)) {
+      throw new IllegalArgumentException("input Parquet files are not Partitioned")
```
fnothaft: Replace with:

```
require(!checkPartitionedParquetFlag(pathName),
  "Input Parquet files (%s) are not partitioned.".format(pathName))
```

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2486,6 +2512,15 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+    }
+    val sd = loadAvroSequenceDictionary(pathName)
+    val nucleotideContigFragments = ParquetUnboundNucleotideContigFragmentRDD(sc, pathName, sd)
+    DatasetBoundNucleotideContigFragmentRDD(nucleotideContigFragments.dataset, nucleotideContigFragments.sequences)
```
fnothaft: Break long line.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2105,6 +2079,28 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
     }
   }
 
+  def loadPartitionedParquetGenotypes(pathName: String, regions: Option[Iterable[ReferenceRegion]] = None, partitionSize: Int = 1000000): GenotypeRDD = {
+    if (!checkPartitionedParquetFlag(pathName)) {
+      throw new IllegalArgumentException("input Parquet files are not Partitioned")
```
fnothaft: Replace with:

```
require(!checkPartitionedParquetFlag(pathName),
  "Input Parquet files (%s) are not partitioned.".format(pathName))
```

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2138,6 +2134,25 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
     }
   }
 
+  def loadPartitionedParquetVariants(pathName: String, regions: Option[Iterable[ReferenceRegion]] = None, partitionSize: Int = 1000000): VariantRDD = {
+    if (!checkPartitionedParquetFlag(pathName)) {
+      throw new IllegalArgumentException("input Parquet files are not Partitioned")
```
fnothaft: Replace with:

```
require(!checkPartitionedParquetFlag(pathName),
  "Input Parquet files (%s) are not partitioned.".format(pathName))
```

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2454,6 +2469,17 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
     }
   }
 
+  def loadPartitionedParquetFeatures(pathName: String): FeatureRDD = {
+    if (!checkPartitionedParquetFlag(pathName)) {
+      throw new IllegalArgumentException("input Parquet files are not Partitioned")
```
fnothaft: Replace with:

```
require(!checkPartitionedParquetFlag(pathName),
  "Input Parquet files (%s) are not partitioned.".format(pathName))
```

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2105,6 +2079,28 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+
+    val genotypes = ParquetUnboundGenotypeRDD(sc, pathName, sd, samples, headers)
+
+    val datasetBoundGenotypeRDD: GenotypeRDD = regions match {
+      case Some(x) => DatasetBoundGenotypeRDD(genotypes.dataset
+        .filter(referenceRegionsToDatasetQueryString(x)), genotypes.sequences, genotypes.samples, genotypes.headerLines)
+      case _ => DatasetBoundGenotypeRDD(genotypes.dataset, genotypes.sequences, genotypes.samples, genotypes.headerLines)
+    }
+    datasetBoundGenotypeRDD
+
```
fnothaft: Extra whitespace.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2105,6 +2079,28 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+    // load header lines
+    val headers = loadHeaderLines(pathName)
+    // load sequence info
+    val sd = loadAvroSequenceDictionary(pathName)
+    // load avro record group dictionary and convert to samples
+    val samples = loadAvroSamples(pathName)
+
+    val genotypes = ParquetUnboundGenotypeRDD(sc, pathName, sd, samples, headers)
+
+    val datasetBoundGenotypeRDD: GenotypeRDD = regions match {
```
fnothaft: Instead of having `regions: Option[Iterable[ReferenceRegion]]` and running:

```
regions match {
  case Some(x) => DatasetBoundGenotypeRDD(genotypes.dataset.filter(referenceRegionsToDatasetQueryString(x))
  DatasetBoundGenotypeRDD(genotypes.dataset)
}
```

Change the type of `regions` to `Iterable[ReferenceRegion]` and then apply the region filters by running:

```
regions.foldLeft(genotypes)(p => p._1.transformDataset(_.filter(referenceRegionsToDatasetQueryString(p._2))))

File: adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala [hunk 41]
```diff
@@ -2904,4 +2939,29 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
       loadParquetFragments(pathName, optPredicate = optPredicate, optProjection = optProjection)
     }
   }
+
+  def writePartitionedParquetFlag(filePath: String): Boolean = {
```
fnothaft: This should be private (and go in `GenomicRDD.scala`, methinks?)

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2904,4 +2939,29 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+    val path = new Path(filePath, "_isPartitionedByStartPos")
+    val fs = path.getFileSystem(sc.hadoopConfiguration)
+    fs.createNewFile(path)
+  }
+
+  def checkPartitionedParquetFlag(filePath: String): Boolean = {
```
fnothaft: This should be private.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2904,4 +2939,29 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+    val path = new Path(filePath, "_isPartitionedByStartPos")
+    val fs = path.getFileSystem(sc.hadoopConfiguration)
+    fs.exists(path)
+  }
+
+  def referenceRegionsToDatasetQueryString(x: Iterable[ReferenceRegion], partitionSize: Int = 1000000): String = {
```
fnothaft: This should be private.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2904,4 +2939,29 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+    var regionQueryString = "(contigName=" + "\'" + x.head.referenceName.replaceAll("chr", "") + "\' and posBin >= \'" +
```
fnothaft: Why are we replacing "chr"s?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
jpdna: I think I understand you want to use the more elegant fold left over a potentially empty regions list, with default being the unfiltered dataset if `regions` is empty.

However, I'm not sure that the code you suggest implements the OR logic that is intended for this region filter, as won't the consecutive applications of filter as the list is folded result in `AND` logic?

I'm incline to leave the clunky `regions match` code for now if we think it is correct, as it seems to work.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
jpdna: I actually find a need to sue this from Mango library in order to determine if dataset is partitioned to see if it should use the partitioned read functions, since we needed to keep them separate.  Perhaps a better option would be some wrapper function like `maybeReadPartitioned` but for now I prefer to keep it public and give client code chance to explicitly check.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -2476,4 +2478,29 @@ abstract class AvroGenomicRDD[T <% IndexedRecord: Manifest, U <: Product, V <: A
   def saveAsParquet(filePath: java.lang.String) {
     saveAsParquet(new JavaSaveArgs(filePath))
   }
+
+  def writePartitionedParquetFlag(filePath: String): Boolean = {
+    val path = new Path(filePath, "_isPartitionedByStartPos")
```
akmorrow13: would there be a way to write a _partitioned file with the fields it is partitioned by, instead of hard coding to start position?

jpdna:
Closing this PR in favor of:
https://github.com/bigdatagenomics/adam/pull/1864

PR #1636, opened 2017-07-26
[ADAM-1635] Eliminate passing FASTQ splittable status via config.

### LINKED ISSUE #1635: FastqInputFormat.FILE_SPLITTABLE in conf not getting passed properly
Added in https://github.com/bigdatagenomics/adam/commit/985e5d8fac0436d5f4ee24017367bd20bb6552ca. I made some tweak between when I tested this last before merging and when we merged it that borked this. A BGZF'ed file will get properly split by the input format, but then the record reader will read the config and see `false` for the FILE_SPLITTABLE flag, and read the whole file.

fnothaft:
Resolves #1635. Instead of passing whether a FASTQ was splittable via config, checks to see if the compression codec is splittable. This is more reliable. In the case of a .gz file, the BGZFEnhancedGZipCodec properly handles this edge case by checking the stream type; this coupled with us explicitly checking the stream when split picking ensures that we don't try to create an invalid GZIP split. Additionally, I identified and fixed an error in the old FASTQ code that did a seek on the uncompressed input stream to backtrack if seeing a line of quality scores that began with @ when identifying the position of the first valid record in a split. Instead, we check for two successive lines that start with an @, which indicates that the first line contains quality scores, while the second line contains read names.

### DIFF
--- a/adam-core/src/main/java/org/bdgenomics/adam/io/FastqRecordReader.java
+++ b/adam-core/src/main/java/org/bdgenomics/adam/io/FastqRecordReader.java
@@ -145,22 +145,22 @@ abstract class FastqRecordReader extends RecordReader<Void, Text> {
         CompressionCodecFactory codecFactory = new CompressionCodecFactory(conf);
         CompressionCodec codec = codecFactory.getCodec(file);
 
-	// if our codec is splittable, we can (tentatively) say that
-	// we too are splittable.
-	//
-	// if we get a bgzfenhancedcodec, the codec might not actually
-	// be splittable. however, if we get a non-splittable gz file,
-	// several things happen:
-	//
-	// 1. the input format will detect this, and will not split the
-	//    file
-	// 2. the bgzfenhancedcodec will check the underlying data type
-	//    (BGZF vs GZIP) at input stream creation time, and will
-	//    apply the appropriate codec.
-	//
-	// if we get an unsplittable codec, really all that we do differently
-	// is skip the positioning check, since we know that we're at the
-	// start of the file and can get to reading immediately
+        // if our codec is splittable, we can (tentatively) say that
+        // we too are splittable.
+        //
+        // if we get a bgzfenhancedcodec, the codec might not actually
+        // be splittable. however, if we get a non-splittable gz file,
+        // several things happen:
+        //
+        // 1. the input format will detect this, and will not split the
+        //    file
+        // 2. the bgzfenhancedcodec will check the underlying data type
+        //    (BGZF vs GZIP) at input stream creation time, and will
+        //    apply the appropriate codec.
+        //
+        // if we get an unsplittable codec, really all that we do differently
+        // is skip the positioning check, since we know that we're at the
+        // start of the file and can get to reading immediately
         isSplittable = (codec instanceof SplittableCompressionCodec);
         
         if (codec == null) {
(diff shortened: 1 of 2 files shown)


adam-core/src/main/java/org/bdgenomics/adam/io/FastqRecordReader.java
```diff
@@ -145,12 +145,23 @@ protected FastqRecordReader(final Configuration conf,
         CompressionCodecFactory codecFactory = new CompressionCodecFactory(conf);
         CompressionCodec codec = codecFactory.getCodec(file);
 
-        // we expect this record reader to be used only with input formats that
-        // extend FastqInputFormat. the behavior of this record reader depends on
-        // whether the input stream is splittable or not. FastqInputFormat maintains
-        // a contract where it will put the file's splittable status into the hadoop
-        // configuration object.
-        isSplittable = conf.getBoolean(FastqInputFormat.FILE_SPLITTABLE, false);
+	// if our codec is splittable, we can (tentatively) say that
```
heuermh: Yer editor accidentally used tabs for some of these lines

adam-core/src/main/java/org/bdgenomics/adam/io/FastqRecordReader.java
```diff
@@ -206,20 +217,19 @@ protected final int positionAtFirstRecord(final FSDataInputStream stream,
...
+	    // Advance to the start of the first record that ends with /1
+	    // We use a temporary LineReader to read lines until we find the
+	    // position of the right one.  We then seek the file to that position.
+	    stream.seek(start);
             reader = new LineReader(stream);
         } else {
             // see above note about 
             // SplittableCompressionCodec.createInputStream needing the stream
             // to be at offset 0
-            stream.seek(0);
```
heuermh: the comment above this line can be removed

adam-core/src/main/java/org/bdgenomics/adam/io/FastqRecordReader.java
fnothaft: Thanks for catching this; I was editing this patch on a different computer from my usual and I was wondering why the diff looked weird.

adam-core/src/main/java/org/bdgenomics/adam/io/FastqRecordReader.java
fnothaft: I think this is still useful info to keep around, but I'll update the comment to better reflect the changed code.

fnothaft:
Pushed a commit addressing reviewer comments.

BTW @heuermh do you think it would be worthwhile to add something to our CI that would flag any tabs in our source and fail the build? I would've missed those if you hadn't caught them.

heuermh:
> do you think it would be worthwhile to add something to our CI that would flag any tabs in our source and fail the build? I would've missed those if you hadn't caught them.

We have a linter that runs on the scala source, this made it through because it was a java source file.  I don't think we can put a CI check on the whole repo because some of our test resources require tab characters.

heuermh:
Thank you, @fnothaft

heuermh:
Sorry, wrong button, I should've squashed.

fnothaft:
> We have a linter that runs on the scala source, this made it through because it was a java source file. I don't think we can put a CI check on the whole repo because some of our test resources require tab characters.

I mean, sure, but we could do something like:

```
find adam-*/src -name "*.java" -exec ./scripts/failIfHasTabs.sh {} \;
find adam-*/src -name "*.R" -exec ./scripts/failIfHasTabs.sh {} \;
find adam-*/src -name "*.py" -exec ./scripts/failIfHasTabs.sh {} \;
```

heuermh:
+1, add `*.pom`, `*.sh`

PR #1648, opened 2017-07-30
[ADAM-1647] Fix bad JAR discovery grep in bin/pyadam.

### LINKED ISSUE #1647: incorrect grep for assembly jars in bin/pyadam
the pyadam script has the incorrect grep for the assembly jars at https://github.com/bigdatagenomics/adam/blob/master/bin/pyadam#L33

Should
`grep "^adam_[0-9A-Za-z\.-]*\.jar$"` 
be
`^adam-assembly_[0-9A-Za-z\.-]*\.jar$" 
`?

Unless I am missing the jar it is supposed to be accessing.

fnothaft:
Resolves #1647.

### DIFF
--- a/bin/pyadam
+++ b/bin/pyadam
@@ -30,14 +30,14 @@ else
   ASSEMBLY_DIR="$SCRIPT_DIR/adam-assembly/target"
 fi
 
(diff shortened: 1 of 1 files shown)


heuermh:
Thank you, @fnothaft.  Good catch, @akmorrow13!

PR #1665, opened 2017-08-11
[ADAM-1663] Enable read groups with repeated names when unioning.

### LINKED ISSUE #1663: Can't merge BAM files containing the same sample
We outlawed this in our glob loader code. However, this comes up if you want to do something like load in BAM files that are split out by sample. I had a benchmarking workflow where I had split a single aligned dataset into a single file per chromosome for debug, and got this on load:

```
User class threw exception: java.lang.IllegalArgumentException: requirement failed: Read group dictionary contains multiple samples with identical read group names.
java.lang.IllegalArgumentException: requirement failed: Read group dictionary contains multiple samples with identical read group names.
	at scala.Predef$.require(Predef.scala:233)
	at org.bdgenomics.adam.models.RecordGroupDictionary.<init>(RecordGroupDictionary.scala:85)
	at org.bdgenomics.adam.models.RecordGroupDictionary.$plus$plus(RecordGroupDictionary.scala:97)
	at org.bdgenomics.adam.rdd.ADAMContext$$anonfun$loadBam$1$$anonfun$13.apply(ADAMContext.scala:1485)
	at org.bdgenomics.adam.rdd.ADAMContext$$anonfun$loadBam$1$$anonfun$13.apply(ADAMContext.scala:1484)
```

fnothaft:
Resolves #1663. When unioning two read groups together, calls distinct after appending the two RecordGroupDictionaries.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/models/RecordGroupDictionary.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/models/RecordGroupDictionary.scala
@@ -94,7 +94,7 @@ case class RecordGroupDictionary(recordGroups: Seq[RecordGroup]) {
    * @return The merged record group dictionary.
    */
   def ++(that: RecordGroupDictionary): RecordGroupDictionary = {
-    new RecordGroupDictionary(recordGroups ++ that.recordGroups)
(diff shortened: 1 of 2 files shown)


heuermh:
Thank you, @fnothaft!

PR #1671, opened 2017-08-18
[ADAM-1670] Add ability to selectively project VCF fields.

### LINKED ISSUE #1670: Support loading a subset of VCF fields
Parsing all admissible VCF fields may be more expensive than is warranted for all uses. We should allow users to project a set of VCF fields on load.

fnothaft:
Resolves #1670. Adds a method loadVcfWithProjection to ADAMContext. This method takes a list of INFO/FORMAT field names, which are then applied as filters to the set of header lines used when creating the VariantContextConverter.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
@@ -2021,6 +2021,73 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
   }
 
   /**
+   * Load variant context records from VCF into a VariantContextRDD.
+   *
+   * Only converts the core Genotype/Variant fields, and the fields set in the
+   * requested projection. Core variant fields include:
+   *
+   * * Names (ID)
+   * * Filters (FILTER)
+   *
+   * Core genotype fields include:
+   *
+   * * Allelic depth (AD)
+   * * Read depth (DP)
+   * * Min read depth (MIN_DP)
+   * * Genotype quality (GQ)
+   * * Genotype likelihoods (GL/PL)
+   * * Strand bias components (SB)
+   * * Phase info (PS,PQ)
+   *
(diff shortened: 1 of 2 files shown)


fnothaft:
This is not ready to merge; please hold for testing.

heuermh:
Curious how you are performance testing this method, and how we might separate cost on our side vs that incurred in htsjdk, because isn't htsjdk parsing all the fields anyway?

fnothaft:
@heuermh will benchmark this and report back.

heuermh:
In single node mode, `loadVcf` is 2.5x the runtime of a streaming implementation, whereas `loadVcfWithProjections` is 1.3x.

In clustered mode on Yarn with the data on HDFS, `loadVcf` is 2.4x the runtime of a streaming implementation, whereas `loadVcfWithProjections` is 1.4x.

`loadVcfWithProjections` has an added benefit in that it allows the user to filter out known errors in the VCF file.

```scala
$ dsh-bio filter-vcf --filter -i IN -o OUT

$ adam-shell -i filter-vcf.scala

val variants = sc.loadVcf("IN").toVariants()
val filtered = variants.transform(rdd => { rdd.filter(_.getFiltersPassed()) })
filtered.toVariantContexts().saveAsVcf("OUT", asSingleFile = true, deferMerging = false, disableFastConcat = false, stringency = htsjdk.samtools.ValidationStringency.SILENT)
System.exit(0)

$ adam-shell -i filter-filtered-vcf.scala

val variants = sc.loadVcfWithProjection("IN", Set.empty, Set.empty).toVariants()
val filtered = variants.transform(rdd => { rdd.filter(_.getFiltersPassed()) })
filtered.toVariantContexts().saveAsVcf("IN", asSingleFile = true, deferMerging = false, disableFastConcat = false, stringency = htsjdk.samtools.ValidationStringency.SILENT)
System.exit(0)
```

heuermh:
Thank you, @fnothaft

PR #1681, opened 2017-08-24
Remove adam- prefix from module directory names.

heuermh:
Pet peeve.


fnothaft:
I'm a -0.5. I'd like all of our projects to be consistent, and cannoli is the only one that lacks prefixes. My main motivation for preferring the prefixes is that then the module name is a closer match to the artifact name. As an aside, there doesn't seem to be consistency across all multi-module projects; e.g., Spark has no prefixes, but Hadoop and Parquet both have prefixes on their modules.

heuermh:
ADAM isn't even internally consistent (`distribution` module). I don't like the extra keystrokes when tab completing.

fnothaft:
Yeah, I agree RE: `distribution` being an exception (I don't think that was intentional, though), but again, I'd rather change 1 project and one module to achieve consistency, rather than changing `n - 2` projects and `m - 1` modules.

heuermh:
Closing as wontfix.  Created new issue #1716 to track `distribution` module name.

PR #1684, opened 2017-08-25
[ADAM-1682] Add variant quality field.

### LINKED ISSUE #1682: Convert and store variant quality field.
Needed for https://github.com/bigdatagenomics/avocado/issues/253. Depends on https://github.com/bigdatagenomics/bdg-formats/issues/146.

fnothaft:
Resolves #1682. Depends on https://github.com/bigdatagenomics/bdg-formats/pull/147.


heuermh:
LGTM.  Could you add an additional `log.info` line at or around

https://github.com/fnothaft/adam/blob/bd697e38e681a1402acc4f9a91da1a3532ccaa6b/adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantContextConverter.scala#L311

to complement this one

https://github.com/fnothaft/adam/blob/bd697e38e681a1402acc4f9a91da1a3532ccaa6b/adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantContextConverter.scala#L275

with a message about splitting alternate alleles, and how some fields such as `Variant.quality` may no longer be particularly relevant or correct?

fnothaft:
Jenkins, test this please.

fnothaft:
This is not mergeable until the 0.11.2 release of bdg-formats is pushed.

fnothaft:
This is good to go from my side.

@heuermh WRT your comment:

> LGTM. Could you add an additional log.info line at or around...

I just pushed a second commit that actually removes the original log.info. I was working with someone recently who was seeing really slow performance when working with VCF data from ADAM. A bit of profiling revealed that specific line to be the culprit. This also leads to a lot of log spamming. Since the info that line outputs could be obtained by looking at the input and output of the conversion, I feel like it isn't worth the configuration hassle of keeping it in place.

heuermh:
> I just pushed a second commit that actually removes the original log.info. I was working with someone recently who was seeing really slow performance when working with VCF data from ADAM. A bit of profiling revealed that specific line to be the culprit. This also leads to a lot of log spamming. Since the info that line outputs could be obtained by looking at the input and output of the conversion, I feel like it isn't worth the configuration hassle of keeping it in place.

+1 to removing that log line.

We need something somewhere that the `Variant.quality` field is suspect when a multi-allelic row is split.  Is that documentation?  What if we emitted a single log message only on the first instance? Added a new INFO binary flag?  Added `splitFromMultiAllelic` field to `Variant` or `VariantAnnotation`?

Additionally, and we may want to scan across the codebase for other instances of this, `log.info` does not have the `log.isInfoEnabled()` guard, whereas `logInfo()` does.

https://github.com/bigdatagenomics/utils/blob/master/utils-misc/src/main/scala/org/bdgenomics/utils/misc/Logging.scala#L43

heuermh:
bdg-formats 0.11.3 has been released, will update when it shows up on Maven Central

fnothaft:
The release has pushed: http://search.maven.org/#artifactdetails%7Corg.bdgenomics.bdg-formats%7Cbdg-formats%7C0.11.3%7Cjar, updating this now.

heuermh:
Rebase and merge as three commits, or would you like me to squash?

fnothaft:
Please keep as 3.

heuermh:
Thank you, @fnothaft

PR #1687, opened 2017-08-27
[ADAM-1679] Unmapped items should not get caught in requirement when sorting

### LINKED ISSUE #1679: Shuffle region join throws multimapped exception for unmapped reads
If you have an unmapped read, you get the (confusing) exception:

```
java.lang.IllegalArgumentException: requirement failed: Cannot sort RDD containing a multimapped element. 
```

fnothaft:
Resolves #1679.


heuermh:
Thank you, @fnothaft

PR #1689, opened 2017-08-28
Migrate bdg-formats to new adam-formats module.

heuermh:


fnothaft:
I'm -0.5 to moving this here. We don't rev bdg-formats that often, and we'll want bdg-formats to stay even more stable than ADAM after we go to ADAM 1.0.0. Additionally, moving this in here would lose all of the revision control history that we've got in bdg-formats. What're your motivations for the move, @heuermh ?

heuermh:
If bdg-formats doesn't have a life of its own (afaik the only projects that depend on it without also depending on ADAM are mine) and is typically released in lock-step with ADAM, there is not much reason for it to be in a separate repo.

My latest itch is around documentation; I'd like to host javadocs and mapping docs, and it doesn't feel like those merit standing on their own separate from ADAM docs.

> Additionally, moving this in here would lose all of the revision control history that we've got in bdg-formats.

I don't know, there may be some git magick to import the source and save history but yeah, the issues and pull requests would get lost.

fnothaft:
> (afaik the only projects that depend on it without also depending on ADAM are mine)

@ryan-williams as well. With the work on SNAP and bowtie2 in cannoli, I would still like to attempt the dream eventually of using the schemas from C++ land.

> I don't know, there may be some git magick to import the source and save history but yeah, the issues and pull requests would get lost.

It appears that there is a way to import the commits (https://stackoverflow.com/a/8396318), but we'd lose the issues and pull request history, which is displeasing.

ryan-williams:
I don't think I currently depend on bdg-formats anywhere that I don't also depend on adam core, fwiw.

However, it seems like bdg-formats would still be released as its own artifact even if it was put into the ADAM repo, so that wouldn't close the door on anyone seeking to depend on formats and not core, right?

Is formats usually released in lockstep with ADAM, though? That wasn't really my impression.

Anyway there's definitely costs to this move and I don't have much insight into what is improved by it, so my drive-by reaction is not very enthusiastic but I defer to y'all. I don't share the posture that things should be put in the same repo by default.

fnothaft:
> I don't think I currently depend on bdg-formats anywhere that I don't also depend on adam core, fwiw.

I think your sbt infrastructure may pull it in by default? IIRC, I saw it show up in the transitive dependency tree of genome-loci? That may've been an error on my side.

> Anyway there's definitely costs to this move and I don't have much insight into what is improved by it, so my drive-by reaction is not very enthusiastic but I defer to y'all. I don't share the posture that things should be put in the same repo by default.

+1 @ryan-williams

ryan-williams:
> I think your sbt infrastructure may pull it in by default? IIRC, I saw it show up in the transitive dependency tree of genome-loci? That may've been an error on my side.

That should be untrue on both counts. `sbt 'show fullClasspath' | grep formats` should (have always) come up blank, as one way to inspect. Definitely lmk if you see otherwise.

fnothaft:
@ryan-williams yeah, you're right. I just reran `mvn dependency:tree` and didn't see the `bdg-formats` dep this time around. Not sure why I saw it earlier... must've been bleary eyed late at night...

heuermh:
Rebased.

heuermh:
Closing as WontFix.  Feel free to create a new PR after rebasing against git head.

PR #1690, opened 2017-08-28
[ADAM-1688] Add bdg-formats exclusion to org.hammerlab:genomic-loci dependency.

heuermh:
Fixes #1688

PR #1692, opened 2017-08-29
[ADAM-1688] Remove dependency on org.hammerlab:genomic-loci.

### LINKED ISSUE #1688: Conflict in bdg-formats dependency version due to org.hammerlab:genomic-loci
```
$ mvn dependency:tree
...
[INFO] +- org.hammerlab:genomic-loci_2.10:jar:1.4.4:compile
[INFO] |  +- org.hammerlab:args4s_2.10:jar:1.0.0:compile
[INFO] |  +- org.bdgenomics.bdg-formats:bdg-formats:jar:0.10.0:compile
[INFO] |  +- org.hammerlab:iterator_2.10:jar:1.0.0:compile
[INFO] |  |  \- org.spire-math:spire_2.10:jar:0.11.0:compile
[INFO] |  |     +- org.spire-math:spire-macros_2.10:jar:0.11.0:compile
[INFO] |  |     +- org.typelevel:machinist_2.10:jar:0.4.1:compile
[INFO] |  |     \- org.scalamacros:quasiquotes_2.10:jar:2.0.1:compile
[INFO] |  \- org.hammerlab:string-utils_2.10:jar:1.0.0:compile
```

heuermh:
Fixes #1688, alternative to #1690.

### DIFF
--- a/adam-core/pom.xml
+++ b/adam-core/pom.xml
@@ -184,11 +184,6 @@
       <scope>compile</scope>
     </dependency>
     <dependency>
-      <groupId>org.hammerlab</groupId>
-      <artifactId>genomic-loci_${scala.version.prefix}</artifactId>
-      <scope>compile</scope>
-    </dependency>
-    <dependency>
       <groupId>com.esotericsoftware.kryo</groupId>
       <artifactId>kryo</artifactId>
       <scope>compile</scope>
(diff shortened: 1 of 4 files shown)


fnothaft:
I have a preference for #1690. Why do you prefer this?

heuermh:
> Why do you prefer this?

I'd rather have these +24 lines of code than have to deal with dependency problems.

fnothaft:
The dependency you're worried about is bdg-formats, right? From a grep of the source, I don't think genomic-loci actually use bdg-formats, so an exclusion should be sufficient.

heuermh:
No, I'm worried about org.hammerlab generally.  I was hesitant to bring it in initially, and after further consideration, I would rather not have the dependency.

fnothaft:
BTW, have you identified any issues in genomic-loco WRT GRCh38/HLA contig names? If so, let's patch the issue upstream, no?

fnothaft:
Closing; resolved by #1690.

PR #1698, opened 2017-08-29
[ADAM-1694] Add short readable descriptions for toString in subclasses of GenomicRDD.

### LINKED ISSUE #1694: Make GenomicRDD.toString more adam-shell friendly

heuermh:
Fixes #1694

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
@@ -115,6 +115,9 @@ private[rdd] object GenomicRDD {
  */
 trait GenomicRDD[T, U <: GenomicRDD[T, U]] extends Logging {
 
+  override def toString = "%s with %d reference sequences"
+    .format(getClass.getSimpleName, sequences.size)
+
   /**
    * The RDD of genomic data that we are wrapping.
    */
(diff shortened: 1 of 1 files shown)


adam-core/src/main/scala/org/bdgenomics/adam/rdd/contig/NucleotideContigFragmentRDD.scala
```diff
@@ -490,4 +490,7 @@ sealed abstract class NucleotideContigFragmentRDD extends AvroGenomicRDD[Nucleot
       (p._1, p._2: java.lang.Long)
     }).toJavaRDD()
   }
+
+  override def toString = "%s with %d reference sequences"
```
fnothaft: You should be able to factor this into the `GenomicRDD` trait, no?

heuermh: I think it is possible ... but I read some odd things about toString on traits and figured the sealed abstract classes would be the best place.  Then the messages are slightly different so it isn't exactly breaking DRY.

fnothaft: What odd things did you see?

fnothaft:
Merged! Thanks @heuermh!

Issue #1703, opened 2017-08-30
Allow validation stringency to be set in AnySAMOutFormatter


fnothaft:
Needed to do with GRCh38 `@SQ` headerline wonkyness in bigdatagenomics/cannoli#65.

fnothaft:
Associated issue: https://github.com/amplab/snap/issues/107

Issue #1702, opened 2017-08-30
InterleavedFastqInFormatter should sort by readInFragment


fnothaft:
Otherwise, you can get reads that are flipped in order. From WIP on https://github.com/bigdatagenomics/cannoli/issues/65:

```
SNAP exited with exit code 1 from line 404 of file SNAPLib/FASTQ.cpp
PairedInterleavedFASTQReader: second read of batch doesn't have ID ending with /2: 'H06JUADXX130110:1:2112:13939:74406/1'
PairedInterleavedFASTQReader: second read of batch doesn't have ID ending with /2: 'H06HDADXX130110:2:1212:7495:26889/1'
SNAP exited with exit code 1 from line 404 of file SNAPLib/FASTQ.cpp
PairedInterleavedFASTQReader: first read of batch doesn't have ID ending with /1: 'H06HDADXX130110:1:2112:18137:64299/2'
SNAP exited with exit code 1 from line 404 of file SNAPLib/FASTQ.cpp
SNAP exited with exit code 1 from line 399 of file SNAPLib/FASTQ.cpp
PairedInterleavedFASTQReader: first read of batch doesn't have ID ending with /1: 'H06JUADXX130110:1:1115:17025:85774/2'
```

PR #1704, opened 2017-08-30
Misc cleanup needed for bigdatagenomics/cannoli#65

### LINKED ISSUE #1701: Allow silencing the # of reads in fragment warning in InterleavedFastqInFormatter
If you're realigning a file that was already aligned, [this](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/fragment/InterleavedFASTQInFormatter.scala#L62-L68) warning creates lots of log noise.

fnothaft:
Resolves #1701, #1702, #1703.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/fragment/InterleavedFASTQInFormatter.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/fragment/InterleavedFASTQInFormatter.scala
@@ -80,8 +80,18 @@ class InterleavedFASTQInFormatter private (
         outputOriginalBaseQualities = writeOriginalQualities) + "\n"
 
       // write both to the output stream
-      os.write(fastq1.getBytes)
-      os.write(fastq2.getBytes)
+      // ensure that reads are ordered properly if ordering is known (see #1702)
+      if (read1.getReadInFragment == 0 &&
+        read2.getReadInFragment == 1) {
+        os.write(fastq1.getBytes)
+        os.write(fastq2.getBytes)
+      } else if (read1.getReadInFragment == 1 &&
+        read2.getReadInFragment == 0) {
+        os.write(fastq2.getBytes)
+        os.write(fastq1.getBytes)
+      } else {
+        log.warn("Improper pair of reads in fragment %s. Dropping...".format(p))
+      }
     })
   }
 }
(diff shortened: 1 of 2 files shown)


adam-core/src/main/scala/org/bdgenomics/adam/rdd/fragment/InterleavedFASTQInFormatter.scala
```diff
@@ -59,11 +60,11 @@ class InterleavedFASTQInFormatter private (
     iter.flatMap(frag => {
       val reads = converter.convertFragment(frag).toSeq
 
-      if (reads.size < 2) {
+      if (enableLogging && reads.size < 2) {
```
heuermh: would simply dropping to `log.info` accomplish the same goal?

fnothaft: I mean, it really is a warning message. However, if you're doing development of a piped tool and you want to read your executor logs to figure out why the tool running under the pipe is blowing up, the warnings get in the way of reading the logs.

If you're not 100% sold on this change, I'm OK with dropping it. It was just for my convenience when developing.

heuermh: I don't have data to test these particular checks (though I suppose I could invent some) and am not all that confident our logging configuration works.  Could you try adding
```
log4j.logger.org.bdgenomics.adam.rdd.fragment.InterleavedFASTQInFormatter=ERROR
```
to your log4j.properties and see if it does the trick?

fnothaft: Yeah, agreed. I've rebased this and dropped the commit that resolves #1701.

fnothaft:
This is good to go from my side.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AnySAMOutFormatter.scala
```diff
@@ -29,7 +29,7 @@ import scala.collection.mutable.ListBuffer
  * An OutFormatter that automatically infers whether the piped input is SAM or
  * BAM. Autodetecting streamed CRAM is not currently supported.
  */
-class AnySAMOutFormatter extends OutFormatter[AlignmentRecord] {
+case class AnySAMOutFormatter(stringency: ValidationStringency = ValidationStringency.STRICT) extends OutFormatter[AlignmentRecord] {
```
heuermh: From the Jenkins failures, looks like the default ctr parameter value isn't available to Python.  Could you add a no-arg `apply` method or update the Python code?

heuermh:
Thank you, @fnothaft

Issue #1740, opened 2017-09-21
Standardize regionFn for ShuffleJoin returned objects


devin-petersohn:
Some of the `regionFn`s have the value as the first set of regions, and the key second, but I need them to all be the same. Because all the broadcast joins have the left first, the ShuffleJoins should follow that pattern.

devin-petersohn:
Resolved in #1741.

PR #1741, opened 2017-09-21
Enabled thresholding for joins and standardized regionFn

### LINKED ISSUE #1739: Shuffle, Broadcast Joins with threshold
Allowing a threshold for joins to allow for joining nearby, but not necessarily overlapping records. Needed in both Shuffle and Broadcast joins.

devin-petersohn:
Resolves #1739, #1740.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
@@ -1176,7 +1176,7 @@ trait GenomicRDD[T, U <: GenomicRDD[T, U]] extends Logging {
    * @return Returns a new genomic RDD containing all pairs of keys that
    *   overlapped in the genomic coordinate space.
    */
-  def shuffleRegionJoin[X, Y <: GenomicRDD[X, Y]](
+  private[rdd] def shuffleRegionJoin[X, Y <: GenomicRDD[X, Y]](
     genomicRdd: GenomicRDD[X, Y],
     optPartitions: Option[Int],
     flankSize: Long)(
@@ -1199,31 +1199,6 @@ trait GenomicRDD[T, U <: GenomicRDD[T, U]] extends Logging {
         getReferenceRegions(kv._1).map(_.pad(-1 * flankSize)) ++
           genomicRdd.getReferenceRegions(kv._2)
       })
-  }
-
-  /**
-   * Performs a sort-merge inner join between this RDD and another RDD.
-   *
-   * In a sort-merge join, both RDDs are co-partitioned and sorted. The
-   * partitions are then zipped, and we do a merge join on each partition.
-   * The key equality function used for this join is the reference region
-   * overlap function. Since this is an inner join, all values who do not
-   * overlap a value from the other RDD are dropped.
-   *
-   * @param genomicRdd The right RDD in the join.
-   * @param optPartitions Optionally sets the number of output partitions. If
-   *   None, the number of partitions on the resulting RDD does not change.
-   * @return Returns a new genomic RDD containing all pairs of keys that
-   *   overlapped in the genomic coordinate space.
-   */
-  def shuffleRegionJoin[X, Y <: GenomicRDD[X, Y]](
-    genomicRdd: GenomicRDD[X, Y],
-    optPartitions: Option[Int])(
-      implicit tTag: ClassTag[T],
-      xTag: ClassTag[X],
-      txTag: ClassTag[(T, X)]): GenericGenomicRDD[(T, X)] = {
-
-    shuffleRegionJoin(genomicRdd, optPartitions, 0L)
   }
 
   /**
@@ -1293,7 +1268,7 @@ trait GenomicRDD[T, U <: GenomicRDD[T, U]] extends Logging {
    *   overlapped in the genomic coordinate space, and all keys from the
    *   right RDD that did not overlap a key in the left RDD.
    */
-  def rightOuterShuffleRegionJoin[X, Y <: GenomicRDD[X, Y]](
+  private[rdd] def rightOuterShuffleRegionJoin[X, Y <: GenomicRDD[X, Y]](
     genomicRdd: GenomicRDD[X, Y],
     optPartitions: Option[Int],
     flankSize: Long)(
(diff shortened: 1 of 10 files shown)


adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -757,18 +757,21 @@ trait GenomicRDD[T, U <: GenomicRDD[T, U]] extends Logging {
    * @see broadcastRegionJoinAgainst
    */
   def broadcastRegionJoin[X, Y <: GenomicRDD[X, Y], Z <: GenomicRDD[(T, X), Z]](
-    genomicRdd: GenomicRDD[X, Y])(
+    genomicRdd: GenomicRDD[X, Y],
+    threshold: Long = 0L)(
```
heuermh: `threshold` param needs scaladoc here and elsewhere.

`flankSize`, `padding`, and `threshold` all have similar meanings, is there any reason to choose one over the other?

Finally, adding a parameter with a Scala default value may make life more difficult for the Java API and downstream in Python and R.

devin-petersohn: Thanks for the scaladoc reminder. I need to push the tests too.

I'm not tied to `threshold`, but for a join, `padding` and `flankSize` don't seem to capture the meaning of the input: to find nearby regions within some threshold.

I can rewrite it to separate the optional parameter into two different methods if that's better.

fnothaft: +1 @heuermh RE consistency. My preference is `flankSize`, as that is the amount of flanking sequence to consider when running an operation. Not a huge deal one way or the other, as long as we're consistent.

>> Finally, adding a parameter with a Scala default value may make life more difficult for the Java API and downstream in Python and R.
>
> I can rewrite it to separate the optional parameter into two different methods if that's better.

+1 @heuermh and +1 at separating into two methods

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -815,25 +820,29 @@ trait GenomicRDD[T, U <: GenomicRDD[T, U]] extends Logging {
...
     // key the RDDs and join
     GenericGenomicRDD[(Option[T], X)](RightOuterTreeRegionJoin[T, X]().broadcastAndJoin(
-      buildTree(flattenRddByRegions()),
+      buildTree(flattenRddByRegions().map(f => (f._1.pad(threshold), f._2))),
       genomicRdd.flattenRddByRegions()),
       sequences ++ genomicRdd.sequences,
       kv => {
-        Seq(kv._1.map(v => getReferenceRegions(v))).flatten.flatten ++
+        Seq(kv._1.map(v => getReferenceRegions(v)
+          .map(_.pad(-1 * threshold)))).flatten.flatten ++
```
fnothaft: I did a double-take on the lines that pad by a negative threshold. Can you add documentation to explain that we are removing the padding that we added earlier?

devin-petersohn: Alternatively we could add an `unpad` and put the documentation there. It doesn't have to be public, but it would make this cleaner.

devin-petersohn:
Rebased on current master and overloaded methods with optionals. I haven't changed the `threshold` name yet. I want to make sure it passes with @AmplabJenkins first.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
heuermh: I think a `//` style single line comment would be sufficient

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
heuermh: Now sometimes we have four
```scala
def join(GenomicRDD)
def join(GenomicRDD, Option[Int])
def join(GenomicRDD, Long)
def join(GenomicRDD, Option[Int], Long)
```

I think it is ok to have only the simplest and longest forms, especially if one is an option
```scala
def join(GenomicRDD)
def join(GenomicRDD, Option[Int], Long)
```

devin-petersohn: I have mixed feelings on that. I don't think 4 is too many, but I do see the issue. I don't know if we should restrict users in that way.

Related: How do you pass an `Option` from R/Python? 

We may just keep the method signatures with the `optPartitions` parameter `private[adam]` (because hopefully in the future we can deterministically set the partition number) and go with `genomicRDD: GenomicRDD, flankSize: Long` for the public API. Do users need the power to set the partition count?

heuermh: >  How do you pass an Option from R/Python?

From Java I import `scala.Option`/`scala.Some` directly.

I found a library that did Java `Optional` &rarr; Scala `Option` conversions but didn't feel it was worth adding another dependency.

devin-petersohn: Here is what I am proposing:

```scala
public def join(GenomicRDD)
public def join(GenomicRDD, Long)
private[rdd] def join(GenomicRDD, Option[Int], Long)
```

I just don't see users caring about the number of partitions for their join, and we can probably do better deterministically setting it.

devin-petersohn: `public` specified for clarity.

fnothaft: That SGTM.

devin-petersohn:
Rebased and made changes to architecture reflecting [this](https://github.com/bigdatagenomics/adam/pull/1741#discussion_r141141314) comment.

heuermh:
Squashed and merged as commit 77a16f2.  Thank you, @devin-petersohn

PR #1748, opened 2017-09-26
[ADAM-1747] Fixed subtract bug and tests

### LINKED ISSUE #1747: ReferenceRegion.subtract eliminating valid regions
The implementation is too strict when regions are overlapping the start. Somehow it passed the tests, so I am updating those at the same time as this fix.

devin-petersohn:
Resolves #1747.


devin-petersohn:
I remember the logic getting confusing when I first wrote this, but somehow even the tests were wrong.

heuermh:
Thank you, @devin-petersohn

PR #1749, opened 2017-09-26
update to _replaceRdd with tests

akmorrow13:


adam-python/bdgenomics/adam/rdd.py
```diff
@@ -692,6 +692,11 @@ def _inferConversionFn(self, destClass):
 class FeatureRDD(GenomicDataset):
 
 
+    def _replaceRdd(self, newRdd):
```
heuermh: Is this only necessary in `class FeatureRDD`?

akmorrow13:
@heuermh if it is not present transform will fail (_replaceRdd is not a function)

adam-python/bdgenomics/adam/rdd.py
akmorrow13: @heuermh if it is not present transform will fail (_replaceRdd is not a function)

akmorrow13: https://github.com/akmorrow13/adam/blob/eec00a545dca3e4cd4e258023c99a1afa8a8974e/adam-python/bdgenomics/adam/rdd.py#L104 is defined in GenomicRDD, calling _replaceRDD, which FeatureRDD does not have

heuermh: Thanks!

heuermh:
Thank you, @akmorrow13

PR #1754, opened 2017-10-07
[ADAM-1680] Eliminate non-determinism in the ShuffleRegionJoin.

### LINKED ISSUE #1680: Region join shows non-determinism
Running on a WGS read file vs. the GTF of gene coordinates, I see non-determinism when running the shuffle region join. Specifically, I see about a 0.03% difference in the number of records that I get at the output of the join.

fnothaft:
Resolves #1680. In the existing shuffle region join code, the partition start/stop boundaries are determined by sorting the data and looking at the coordinates of the first and last record on each partition. This is done via a sampling process, which is fundamentally non-deterministic. Once these partition bounds are picked, we replicate records from the right side of the join into the partitions they overlap. There was a bug in this step that led to records being dropped from the first partition of each contig (except for the first contig). This was due to a bug in how the IntervalArray was being used to search for repartitioning bounds.

This PR replaces the interval array with a map between contig names and the partitions that contain their data. The first and last partition of each contig are extended to the start/end of each contig.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
@@ -1870,51 +1870,106 @@ trait GenomicRDD[T, U <: GenomicRDD[T, U]] extends Logging {
     // number of partitions we will have after repartition
     val numPartitions = destinationPartitionMap.length
 
-    // here we create a partition map with a single ReferenceRegion that spans
-    // the entire range, however we have to handle the case where the partition
-    // spans multiple referenceNames because of load balancing.
-    val adjustedPartitionMapWithIndex =
+    val partitionRegionsAndIndices: Map[String, Seq[(ReferenceRegion, Int)]] = {
 
       // the zipWithIndex gives us the destination partition ID
-      destinationPartitionMap.flatten.zipWithIndex.map(g => {
-        val (firstRegion, secondRegion, index) = (g._1._1, g._1._2, g._2)
+      val pMap: Map[String, Seq[(ReferenceRegion, Int)]] =
+        destinationPartitionMap.zipWithIndex.flatMap(g => {
+          val (optRegions, index) = (g._1, g._2)
 
-        // in the case where we span multiple referenceNames using
-        // IntervalArray.get with requireOverlap set to false will assign all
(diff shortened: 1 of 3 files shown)


adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -1870,51 +1870,106 @@ trait GenomicRDD[T, U <: GenomicRDD[T, U]] extends Logging {
...
+
+      // the above map does not contain sequences who exist in the sequence
+      // dictionary but who are not seen in a record at the start/end of a
+      // partition
+      //
+      // here, we loop over the sequence records, check if they are in the map,
+      // and create a record if the sequence is not in the map
+      var lastIdx = 0
+      val missingSequences: Map[String, Seq[(ReferenceRegion, Int)]] =
+        sequences.records
```
heuermh: there wasn't a dependency on `sequences` before, will this still work if both RDDs have empty sequence dictionaries?  I can give it a try with two `FeatureRDD`s

fnothaft:
Ping for review.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
fnothaft: Yeah, we can't guarantee correct behavior if you don't have correctly defined sequence dictionaries.

fnothaft:
I can go back and rerun if you need numbers, but the perf impact was reasonably negligible.

fnothaft:
Can we merge this?

heuermh:
> Can we merge this?

I've been running joins of features to features, to see if behavior changed per https://github.com/bigdatagenomics/adam/pull/1754#discussion_r143518233.  No answer to that question yet, unfortunately

fnothaft:
My opinion is that we should require the user to provide valid sequence dictionaries. GIGO.

fnothaft:
Pinging for a merge.

heuermh:
I've been trying to cause troubles by joining various features with no sequence dictionaries and haven't found any, in fact the answers are correct whereas before they were not always correct.

heuermh:
Thank you, @fnothaft.  Thank you for the review, @devin-petersohn.

PR #1769, opened 2017-10-19
[ADAM-1768] Add InFormatter for unpaired FASTQ.

### LINKED ISSUE #1768: Add SingleFASTQInFormatter
I've got a need to use the pipe API to align single ended data.

fnothaft:
Resolves #1768. Still testing; not ready for merge. Thoughts on how to handle the configuration properties that are shared with `InterleavedFASTQInFormatter`? I'd rather not duplicate them.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDD.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDD.scala
@@ -101,6 +101,20 @@ private[adam] class AlignmentRecordArraySerializer extends IntervalArraySerializ
 object AlignmentRecordRDD extends Serializable {
 
   /**
+   * Hadoop configuration path to check for a boolean value indicating whether
+   * the current or original read qualities should be written. True indicates
+   * to write the original qualities. The default is false.
+   */
+  val WRITE_ORIGINAL_QUALITIES = "org.bdgenomics.adam.rdd.read.AlignmentRecordRDD.writeOriginalQualities"
+
+  /**
+   * Hadoop configuration path to check for a boolean value indicating whether
+   * to write the "/1" "/2" suffixes to the read name that indicate whether a
+   * read is first or second in a pair. Default is false (no suffixes).
+   */
+  val WRITE_SUFFIXES = "org.bdgenomics.adam.rdd.read.AlignmentRecordRDD.writeSuffixes"
+
+  /**
    * Converts a processing step back to the SAM representation.
    *
    * @param ps The processing step to convert.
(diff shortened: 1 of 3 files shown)


adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/SingleFASTQInFormatter.scala
```diff
@@ -0,0 +1,67 @@
...
+import org.bdgenomics.adam.converters.AlignmentRecordConverter
+import org.bdgenomics.adam.rdd.{ InFormatter, InFormatterCompanion }
+import org.bdgenomics.adam.rdd.fragment.FragmentRDD
+import org.bdgenomics.formats.avro.AlignmentRecord
+import org.bdgenomics.utils.misc.Logging
+
+/**
+ * InFormatter companion that creates an InFormatter that writes FASTQ.
+ */
+object SingleFASTQInFormatter extends InFormatterCompanion[AlignmentRecord, AlignmentRecordRDD, SingleFASTQInFormatter] {
```
heuermh: Single is kinda redundant, how about `FASTQInFormatter`?

heuermh:
> Thoughts on how to handle the configuration properties that are shared with InterleavedFASTQInFormatter? I'd rather not duplicate them.

The way you have them now, keyed by fields on `FragmentRDD`, seems fine to me.  Or are you asking a different question?

fnothaft:
> The way you have them now, keyed by fields on FragmentRDD, seems fine to me. Or are you asking a different question?

No, that's exactly what I'm asking. I guess what I don't like about having them in `FragmentRDD` is that now you're using them with `AlignmentRecordRDD` in addition to `FragmentRDD`. Just seems a bit counterintuitive/awkward for a user.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/SingleFASTQInFormatter.scala
fnothaft: SGTM!

heuermh:
Created https://github.com/fnothaft/adam/pull/21 to address comment above

fnothaft:
> Alternatively, we could put them on `GenotypeRDD` and then refer by specific subclass (`AlignmentRecordRDD.WRITE_SUFFIXES`) but that is somewhat poor code style.

I assume that by `GenotypeRDD`, you meant `RecordGroupGenomicRDD`?

fnothaft:
I'm fine with fnothaft#21  for now, so let's go with that.

fnothaft:
@heuermh Do you want me to squash down your commit into mine, or leave it separate?

heuermh:
Squash is ok

fnothaft:
> Squash is ok

Sounds good! I am going to test this later today, and will squash this down once it is OK to merge.

fnothaft:
I've tested this and it is good to go. Squashing now.

heuermh:
Thank you, @fnothaft

PR #1771, opened 2017-10-20
[ADAM-1770] Genotype should only store core variant fields.

### LINKED ISSUE #1770: Genotypes should only contain the core variant fields
Currently, we save all of the variant annotations into the genotype records. This is not intended and bloats the genotype records by 2.5x on disk.

fnothaft:
Resolves #1770.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantContextConverter.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantContextConverter.scala
@@ -289,9 +289,9 @@ class VariantContextConverter(
     try {
       vc.getAlternateAlleles.toList match {
         case List(NON_REF_ALLELE) | List() => {
-          val variant = variantFormatFn(vc, None, 0, false)
+          val (coreVariant, variant) = variantFormatFn(vc, None, 0, false)
           val genotypes = vc.getGenotypes.map(g => {
-            genotypeFormatFn(g, variant, NON_REF_ALLELE, 0, Some(1), false)
+            genotypeFormatFn(g, coreVariant, NON_REF_ALLELE, 0, Some(1), false)
           })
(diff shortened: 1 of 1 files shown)


heuermh:
I'm not so sure about this.  For the use case where someone converts to `Genotype`s in Avro+Parquet format and archives the VCF to low cost storage, we don't want to lose data.  This change doesn't appear to give the user any choice.

See also comment https://github.com/bigdatagenomics/bdg-formats/pull/108#issuecomment-336207184, which refers to the thread at https://github.com/bigdatagenomics/bdg-formats/pull/108#discussion_r112486393.

fnothaft:
Yeah, the premise is that you would save the VCF to separate Parquet Variant and Genotype files and save both. This is what the Genotype converter did prior to the VariantAnnotation refactor, which was intentional. I had thought that we had kept that behavior, but was wrong.

heuermh:
> Yeah, the premise is that you would save the VCF to separate Parquet Variant and Genotype files and save both.

If you read them in separately, how do you join them back together?

fnothaft:
Short answer is with a region join; it'd be similar to the join we used to have with DatabaseVariantAnnotations. When we clean up #1590, data will be marked as sorted on import from VCF, so the join will be fast.

heuermh:
The [filter here](https://github.com/chapmanb/bcbio-nextgen/blob/master/bcbio/variation/vardict.py#L176) (which is also messy in a number of different reasons) uses a combination of genotype VCF FORMAT attributes, variant VCF INFO attributes, and `Variant.quality`.

The genotype refactor in https://github.com/bigdatagenomics/bdg-formats/pull/108 cleans up the separation between `Variant` and `Genotype`, meaning that one would have to filter `VariantRDD` and then join to `GenotypeRDD` to continue the filter. How might such a filter be implemented in the meantime?

heuermh:
Thank you, @fnothaft

Issue #1666, opened 2017-08-15
Unable to run pyadam, SQLException: Failed to start database 'metastore_db'


heuermh:
Ran into this when testing #1651, need to confirm if it is a broader issue
```
$ ./bin/pyadam 
...
Welcome to
      ____              __
     / __/__  ___ _____/ /__
    _\ \/ _ \/ _ `/ __/  '_/
   /__ / .__/\_,_/_/ /_/\_\   version 2.2.0
      /_/

Using Python version 2.7.13 (default, Dec 20 2016 23:05:08)
SparkSession available as 'spark'.
>>> from bdgenomics.adam.adamContext import ADAMContext
>>> ac = ADAMContext(sc)
>>> reads = ac.loadAlignments("adam-core/src/test/resources/small.sam")
>>> df = reads.toDF()
2017-08-15 13:48:10 ERROR Schema:125 - Failed initialising database.
Unable to open a test connection to the given database. JDBC url = jdbc:derby:;databaseName=metastore_db;create=true, username = APP. Terminating connection pool (set lazyInit to true if you expect to start your database after your app). Original Exception: ------
java.sql.SQLException: Failed to start database 'metastore_db' with class loader org.apache.spark.sql.hive.client.IsolatedClientLoader$$anon$1@21afd1a0, see the next exception for details.
...
Caused by: ERROR XJ040: Failed to start database 'metastore_db' with class loader org.apache.spark.sql.hive.client.IsolatedClientLoader$$anon$1@21afd1a0, see the next exception for details.
	at org.apache.derby.iapi.error.StandardException.newException(Unknown Source)
	at org.apache.derby.impl.jdbc.SQLExceptionFactory.wrapArgsForTransportAcrossDRDA(Unknown Source)
	... 115 more
Caused by: ERROR XSDB6: Another instance of Derby may have already booted the database /Users/heuermh/working/adam/metastore_db.
	at org.apache.derby.iapi.error.StandardException.newException(Unknown Source)
	at org.apache.derby.iapi.error.StandardException.newException(Unknown Source)
	at org.apache.derby.impl.store.raw.data.BaseDataFileFactory.privGetJBMSLockOnDB(Unknown Source)
	at org.apache.derby.impl.store.raw.data.BaseDataFileFactory.run(Unknown Source)
	at java.security.AccessController.doPrivileged(Native Method)
[...]

PR #1777, opened 2017-10-24
[ADAM-1666] SQLContext creation fix for Spark 2.x

akmorrow13:
Fixes https://github.com/bigdatagenomics/adam/issues/1666

Only for Spark 2.x. However, this breaks for Spark 1.x because SparkSession does not exist. Any ideas?

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
@@ -42,7 +42,7 @@ import org.apache.parquet.hadoop.util.ContextUtil
 import org.apache.spark.SparkContext
 import org.apache.spark.rdd.MetricsContext._
 import org.apache.spark.rdd.RDD
-import org.apache.spark.sql.Dataset
+import org.apache.spark.sql.{ SparkSession, Dataset }
 import org.bdgenomics.adam.converters._
 import org.bdgenomics.adam.instrumentation.Timers._
 import org.bdgenomics.adam.io._
@@ -998,6 +998,17 @@ object ADAMContext {
 
   // Add ADAM Spark context methods
   implicit def sparkContextToADAMContext(sc: SparkContext): ADAMContext = new ADAMContext(sc)
+
+  /**
+   * Creates an ADAMContext from a SparkSession. Sets active session, including SQLContext, to input session.
+   *
+   * @param ss SparkSession
+   * @return ADAMContext
+   */
+  def ADAMContextFromSession(ss: SparkSession): ADAMContext = {
+    SparkSession.setActiveSession(ss)
+    new ADAMContext(ss.sparkContext)
+  }
 
   // Add generic RDD methods for all types of ADAM RDDs
   implicit def rddToADAMRDD[T](rdd: RDD[T])(implicit ev1: T => IndexedRecord, ev2: Manifest[T]): ConcreteADAMRDDFunctions[T] = new ConcreteADAMRDDFunctions(rdd)
--- a/adam-python/bdgenomics/adam/adamContext.py
+++ b/adam-python/bdgenomics/adam/adamContext.py
@@ -33,7 +33,7 @@ class ADAMContext(object):
     """
 
 
-    def __init__(self, sc):
+    def __init__(self, ss):
         """
         Initializes an ADAMContext using a SparkContext.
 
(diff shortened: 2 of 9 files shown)


fnothaft:
Hi @akmorrow13! Can you give a bit of background on why the fix necessitates moving to SparkSession from SparkContext?

akmorrow13:
Sure @fnothaft ! SQLContext is not backwards compatible in Spark 2.X, so trying to pass the sparkContext from pyspark to scala will not work. In this case, scala tries to create an SQLContext, which fails, because the metastore_db is already locked by the SQLContext created in pyspark (this throws an error). SparkSessions, however, can be passed back to scala. The SparkSession can then be used to activate scala's session using SparkSession.setActiveSession(ss). This way, scala will not try to re-instantiate an SQLContext.

fnothaft:
I guess I'm a little curious as to why this happens in local mode, but doesn't occur in any of our unit tests. Any thoughts there?

akmorrow13:
I think the issue is that all the unit tests use SparkContext for the entry point, but the Dataset API uses SparkSession, which instantiates SQLContext. I am can recreate (not quite the same) issue in the unit tests if I create a SparkSession in the unit tests:

```
    def setUp(self):
        self._old_sys_path = list(sys.path)
        class_name = self.__class__.__name__
        self.ss = SparkSession.builder.master('local[4]').appName(class_name).getOrCreate()
        print(self.ss)
        self.sc = self.ss.sparkContext
```

This throws the error:
` Another SparkContext is being constructed (or threw an exception in its constructor).  This may indicate an error, since only one SparkContext may be running in this JVM (see SPARK-2243).`

fnothaft:
Would it be possible to make a lower impact/backwards compatible change by using [`SparkSession.sparkContext`](http://spark.apache.org/docs/2.1.0/api/python/pyspark.sql.html?highlight=sparksession#pyspark.sql.SparkSession.sparkContext)? From the Scala API docs, I believe this solely a wrapper around SparkSession, and would avoid re-instantiating the SparkSession.

akmorrow13:
@fnothaft do you mean pass back ss.sparkContext._jsc.sc() to ADAMContext instead of SparkSession? This will cause the same issue, because SQLContext is not being communicated.

akmorrow13:
Ok I think I misunderstood you. I updated `ADAMContextFromSparkSession()` to get the SparkContext from the session. Let me know if this is what you meant @fnothaft .

akmorrow13:
Jenkins, test this please.

fnothaft:
Jenkins, retest this please.

fnothaft:
Actually, @akmorrow13 could you rebase this and clean conflicts?

akmorrow13:
Jenkins, retest this please.

akmorrow13:
Jenkins, retest this please.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -999,6 +999,17 @@ object ADAMContext {
   // Add ADAM Spark context methods
   implicit def sparkContextToADAMContext(sc: SparkContext): ADAMContext = new ADAMContext(sc)
 
+  /**
+   * Creates an ADAMContext form SparkSession. Sets active session, including SQLContext, to input session.
```
fnothaft: `form` -> `from`

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -999,6 +999,17 @@ object ADAMContext {
...
+   *
+   * @param ss SparkSession
+   * @return ADAMContext
+   */
+  def ADAMContextFromSession(ss: SparkSession): ADAMContext = {
+    SparkSession.setActiveSession(ss)
```
fnothaft: OOC, why do you need the `setActiveSession` call?

akmorrow13: This resets the sparkContext to read the session passed in. This fixes the issue if one sparkContext has already been started in the scala backend, then it will be replaced.

akmorrow13:
@fnothaft I haved fixed and squashed so we should be set

fnothaft:
Thanks @akmorrow13! I added a comment explaining the `setActiveSession` and touched up the commit message and merge this manually as 4223f56b3272a9b9a3af201653d7ecaf7b8c1ddf.

PR #1778, opened 2017-10-24
added adam notebook

akmorrow13:


fnothaft:
This is awesome! Would it be possible to add an example notebook, and to add some docs around this?

akmorrow13:
Definitely! Should In Mango, I have an example-files where I put some example-notebooks based on test resource files. Would we prefer to do that or reference the our test s3 bucket?

fnothaft:
My preference is to add a `scripts/examples`, and to then run `adam-notebook` with the example script from within `scripts/jenkins_test`. Two birds, one stone.

akmorrow13:
I can do that. What I have been doing is converting example-notebooks to py files and testing them in the unit tests. That way, the local tests also catch errors. Either way works though.

fnothaft:
Unless we have unit test coverage gaps that the example notebook would catch, I'd rather have the example notebook run through the `adam-notebook` command so we can catch any regressions in `bin/adam-notebook`.

akmorrow13:
I'm not sure how you would run an ipython notebook in jenkins test.

heuermh:
Python on Juypter is one flavor of ADAM notebook, do we want to commit to that as the `adam-notebook`?

akmorrow13:
Jenkins, retest this please.

akmorrow13:
There are 2 issues here:
1. I cannot test adam-notebook in the regression tests. This is because it spawns a new process and is accessible through the browser, which jenkins cannot access.
2. @fnothaft asked to add some example notebooks. However, I don't know what data to access for examples. We can't use data in the resource directories because then the examples will not work in the distribution. Any ideas?

akmorrow13:
Closing, as duplication in https://github.com/bigdatagenomics/mango/blob/master/bin/mango-notebook

PR #1785, opened 2017-10-27
[ADAM-1784] loadIndexedBam should pass the raw globbed path to Hadoop-BAM

### LINKED ISSUE #1784: loadIndexedBam does an unnecessary union
See [these lines](https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala#L1574-L1577). Hadoop-BAM supports the use of an index with globs. Using union can cause worse performance in the wild.

fnothaft:
Resolves #1784.


heuermh:
Thank you, @fnothaft

PR #1790, opened 2017-10-31
[ADAM-1789] Move scala-lang to provided scope.

### LINKED ISSUE #1789: scala-lang should be a provided dependency
Including scala-lang as a compile scoped dependency causes issues when running on some Spark distributions.

fnothaft:
Resolves #1789. Also includes utils-misc in adam-core. Previously, utils-misc was specified as just being test scoped in adam-core.


heuermh:
Thank you, @fnothaft

PR #1794, opened 2017-11-07
[ADAM-1793] Adding vcf2adam and adam2vcf that handle separate variant and genotype data.

### LINKED ISSUE #1793: Add `transformVariantContexts` or similar to cli
Per #1771 it would be useful to load VCF and save both `Variant`s and `Genotype`s separately to Parquet+Avro.

heuermh:
Fixes #1793 

Rough first cut, needs review.

### DIFF
--- a/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Adam2Vcf.scala
+++ b/adam-cli/src/main/scala/org/bdgenomics/adam/cli/Adam2Vcf.scala
@@ -1,0 +1,125 @@
+/**
+ * Licensed to Big Data Genomics (BDG) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The BDG licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


adam-cli/src/main/scala/org/bdgenomics/adam/cli/Vcf2Adam.scala
```diff
@@ -0,0 +1,126 @@
...
+
+  def run(sc: SparkContext) {
+    require(!(args.sort && args.sortLexicographically),
+      "Cannot set both -sort_on_save and -sort_lexicographically_on_save.")
+
+    val variantContexts = sc.loadVcf(
+      args.inputPath,
+      stringency = stringency)
+
+    // todo: cache variantContexts?  sort variantContexts first?
```
fnothaft: I would present this as an option. It's fine to convert twice if you're working with a small dataset, but you'll want to persist if you're working with a large data.

adam-cli/src/main/scala/org/bdgenomics/adam/cli/Adam2Vcf.scala
```diff
@@ -0,0 +1,125 @@
...
+      stringency = stringency)
+
+    val genotypes = sc.loadGenotypes(
+      args.genotypesInputPath,
+      optPredicate = None,
+      optProjection = None,
+      stringency = stringency)
+
+    val join = variants.shuffleRegionJoin(genotypes)
+    val updatedGenotypes = join.rdd.map(pair => {
```
fnothaft: Reading through this a second time, what I'd do here is not a shuffleRegionJoin (think of the multiallelic sites) but instead a join on contig/start/end/ref/alt. You might be able to do this through the dataset API for better (?) performance.

heuermh:
Move to close this in favor of #1865.

heuermh:
Closing in favor of #1865.

PR #1800, opened 2017-11-08
[ADAM-1795] Map -DskipTests=true to exec.skip for Python and R tests.

### LINKED ISSUE #1795: Run python and R tests conditionally in build
Scala and Java unit tests use an in-process SparkContext, whereas the python and R tests depend on an external Spark process.  Consider supporting `-DskipTests=true` as a parameter.

heuermh:
Fixes #1795.


fnothaft:
Merged! Thanks @heuermh!

PR #1805, opened 2017-11-15
Added Filter error when file loaded does not match schema

akmorrow13:


fnothaft:
Merged! Thanks @akmorrow13!

PR #1812, opened 2017-12-01
[ADAM-1807] Check for empty partition when running a piped command.

### LINKED ISSUE #1807: Pipe command can fail with empty partitions
We should drop the empty partitions when running a pipe call.

fnothaft:
Resolves #1807.


heuermh:
LGTM, once test count is updated to include new VCF file

fnothaft:
Jenkins, test this please.

heuermh:
Thank you, @fnothaft!

PR #1820, opened 2017-12-06
Fix link to documentation

nzachow:


heuermh:
Thanks!  Good catch, @nzachow

PR #1828, opened 2017-12-12
Adding loadPairedFastqAsFragments method

ffinfo:
During some talk o gitter I missing this method and looks simple enough to add it myself ;).

Some open issue here are how to do validation if the 2 files are indeed in sync. I can use `stringency: ValidationStringency` like in `loadPairedFastq`. Will follow up this on gitter also.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
@@ -2357,8 +2357,8 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
           val r2Fragment = fastqRecordConverter.convertFragment(r2)
           stringency match {
             case ValidationStringency.STRICT | ValidationStringency.LENIENT =>
-              val r1Name = r1Fragment.getReadName.split(" ").head.stripSuffix("/1")
-              val r2Name = r2Fragment.getReadName.split(" ").head.stripSuffix("/2")
+              val r1Name = r1Fragment.getReadName.takeWhile(_ != ' ').stripSuffix("/1")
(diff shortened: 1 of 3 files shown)


heuermh:
Jenkins, test this please.

ffinfo:
This should be done for now, could someone look at this?

heuermh:
I'm +1 on adding the new method, but would like some time to benchmark the implementation here against a few other possible ones.  Will take a look after cutting the 0.23.0 release.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2306,6 +2306,58 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+    val recordsR2 = readFastq(pathNameR2)
+
+    // convert records
+    val fastqRecordConverter = new FastqRecordConverter
+    // Zip will fail if R1 and R2 has an different number of reads
+    // Checking this explicitly like in loadPairedFastq is not required and not blocking anymore
+    val rdd = recordsR1.zip(recordsR2)
+      .map {
+        case (r1, r2) =>
+          val pairText = new Text(r1._2.toString + r2._2.toString)
```
heuermh: Note zipping the records together assumes that they are sorted in the same order in each file.

ffinfo: Yes indeed, this should always the case when a pair of fastq files. When this is not the case the fastq files are corrupt, at least as paired reads.
Adding a check add more computation, seems like a waste

heuermh:
Thank you for the contribution, @ffinfo!  Closing this in favor of #1866, which as you point out makes fewer assumptions at some performance cost.

Issue #1830, opened 2017-12-13
loadIndexed bam lazily throws Exception if index does not exist


akmorrow13:
If bai file is missing, an error will only occur when performing a collect on the underlying rdd. It would be nice to immediately throw and error if the index does not exist.

heuermh:
Fixed by #1831

PR #1831, opened 2017-12-13
Added check for bam index files in loadIndexedBam

akmorrow13:


akmorrow13:
Fixes https://github.com/bigdatagenomics/adam/issues/1830

heuermh:
We hadn't checked in binary files as test resources before, as then diffs in git aren't meaningful.  If we're going to break that convention, and this is as good a time as any, then we should add an index for `small.bam` and unit test(s) for the happy paths.

Do you need this for the 0.23.0 release?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -1562,11 +1562,29 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+    // select index files.
+    val indexPath = if (pathName.endsWith(".bam"))
+      new Path(pathName + ".bai")
+    else
+      path
+
     // todo: can this method handle SAM and CRAM, or just BAM?
     val bamFiles = getFsAndFiles(path).filter(p => p.toString.endsWith(".bam"))
 
+    val indexFiles = getFsAndFiles(indexPath).filter(p => p.toString.endsWith(".bai"))
```
fnothaft: Use `getFsAndFilesWithFilter`.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -1562,11 +1562,29 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
     val path = new Path(pathName)
+
+    // If pathName is a directory of bam files, scan all files to
+    // select index files.
+    val indexPath = if (pathName.endsWith(".bam"))
+      new Path(pathName + ".bai")
+    else
+      path
+
     // todo: can this method handle SAM and CRAM, or just BAM?
```
fnothaft: Can you update this comment to `// currently only supports BAM` and cross reference https://github.com/bigdatagenomics/adam/issues/1833.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -1562,11 +1562,29 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
     stringency: ValidationStringency = ValidationStringency.STRICT)(implicit s: DummyImplicit): AlignmentRecordRDD = LoadIndexedBam.time {
 
     val path = new Path(pathName)
+
+    // If pathName is a directory of bam files, scan all files to
+    // select index files.
+    val indexPath = if (pathName.endsWith(".bam"))
```
fnothaft: Use `{}`'s.

Also, its not clear that this code and the comment match up. Can you add more to the comment to clarify how this bit of code is necessary for directory support?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -1562,11 +1562,29 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
     val bamFiles = getFsAndFiles(path).filter(p => p.toString.endsWith(".bam"))
 
+    val indexFiles = getFsAndFiles(indexPath).filter(p => p.toString.endsWith(".bai"))
+      .map(r => r.toString)
+
     require(bamFiles.nonEmpty,
       "Did not find any BAM files at %s.".format(path))
+
+    // Make sure each bam file has an index
+    bamFiles.foreach(file => {
```
fnothaft: Nit: if multiple files are missing an index, this will just throw for the first one. Instead, I'd suggest:

```
val missingIndices = bamFiles.filter(f => indexFiles.contains(f.toString + ".bai"))
require(missingIndices.isEmpty,
  "Missing indices for BAMs:\n%s".format(missingIndices.mkString("\n"))
```

adam-core/src/test/scala/org/bdgenomics/adam/rdd/ADAMContextSuite.scala
```diff
@@ -435,6 +435,14 @@ class ADAMContextSuite extends ADAMFunSuite {
     assert(reads.rdd.count == 4)
   }
 
+  sparkTest("loadIndexedBam should throw exception without an index file") {
```
fnothaft: Can you add a test case for the directory/glob case?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -1562,11 +1562,33 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+    val indexPath = if (pathName.endsWith(".bam")) {
+      new Path(pathName + ".bai")
+    } else {
+      path
+    }
+
+    // currently only supports BAM files, see https://github.com/bigdatagenomics/adam/issues/1833
     val bamFiles = getFsAndFiles(path).filter(p => p.toString.endsWith(".bam"))
 
+    val indexFiles = getFsAndFiles(indexPath).filter(p => p.toString.endsWith(".bai"))
```
akmorrow13: @fnothaft getFsAndFilesWith filter wasn't able to find .bai files unless just the directory was specified

fnothaft: Yeah makes sense.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -1562,11 +1562,33 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
     stringency: ValidationStringency = ValidationStringency.STRICT)(implicit s: DummyImplicit): AlignmentRecordRDD = LoadIndexedBam.time {
 
     val path = new Path(pathName)
-    // todo: can this method handle SAM and CRAM, or just BAM?
+
+    // If pathName is a single file or *.bam, append .bai to find all bam indices.
+    // Otherwise, pathName is a directory and the entire path must be searched
+    // for indices.
```
fnothaft: Can you note:
```
// TODO: support BAMs where the index is located at s/.bam/.bai/g
```

These aren't common (`myFile.bam` + `myFile.bai` vs `myFile.bam.bai`) but you do see them in the wild occasionally.

fnothaft: Maybe create an issue for that after this merges.

fnothaft: Actually, if you do `new Path(pathName.toString.replace(".bam", "*.bai"))` here and make a change below (will drop a comment), I think you can handle both cases.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -1562,11 +1562,33 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
     val bamFiles = getFsAndFiles(path).filter(p => p.toString.endsWith(".bam"))
 
+    val indexFiles = getFsAndFiles(indexPath).filter(p => p.toString.endsWith(".bai"))
+      .map(r => r.toString)
+
     require(bamFiles.nonEmpty,
       "Did not find any BAM files at %s.".format(path))
+
+    val missingIndices = bamFiles.filterNot(f => {
+      indexFiles.contains(f.toString + ".bai")
```
fnothaft: If you do `indexFiles.contains(f.toString + ".bai") || indexFiles.contains(f.toString.dropRight(3) + "bai"))` you should handle both the `myFile.bam.bai` and `myFile.bai` cases.

adam-core/src/test/scala/org/bdgenomics/adam/rdd/ADAMContextSuite.scala
fnothaft: Ping! Also, add a test case for the `myFile.bai` case.

akmorrow13:
@heuermh and @fnothaft can we merge this?

heuermh:
I'll give it one last review check after the 0.23.0 release is cut

fnothaft:
Yeah, this is good to go from my side but I have a strong preference for it being in 0.24.0 instead of 0.23.0.

heuermh:
Thank you, @akmorrow13!

PR #1839, opened 2017-12-19
[ADAM-1838] Make populating variant.annotation field in Genotype configurable

### LINKED ISSUE #1838: Allow variant annotations to be loaded into genotypes
In #1771 we stripped variant annotation loading out of genotypes, in favor of keeping the variant annotations stored in the accompanying variant (c.f., `VariantContext`). However, if you're working with single sample data (e.g., gVCFs), having the variant annotation nested under the genotype may be useful.

fnothaft:
Resolves #1838. Modifies the behavior of #1771, which disabled populating the `variant.annotation` field in the `Genotype` record. Now, this field is not populated by default. To enable populating it, a user can set the property `org.bdgenomics.adam.converters.VariantContextConverter.NEST_ANN_IN_GENOTYPES` to true.

### DIFF
--- a/adam-core/src/test/scala/org/bdgenomics/adam/rdd/variant/VariantContextRDDSuite.scala
+++ b/adam-core/src/test/scala/org/bdgenomics/adam/rdd/variant/VariantContextRDDSuite.scala
@@ -181,7 +181,7 @@ class VariantContextRDDSuite extends ADAMFunSuite {
     val records = rdd.rdd.count
 
     implicit val tFormatter = VCFInFormatter
-    implicit val uFormatter = new VCFOutFormatter
+    implicit val uFormatter = new VCFOutFormatter(sc.hadoopConfiguration)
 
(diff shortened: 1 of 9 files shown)


adam-core/src/test/scala/org/bdgenomics/adam/rdd/variant/VariantContextRDDSuite.scala
```diff
@@ -210,7 +210,7 @@ class VariantContextRDDSuite extends ADAMFunSuite {
     val rdd: VariantContextRDD = sc.loadVcf(freebayesVcf)
 
     implicit val tFormatter = VCFInFormatter
```
heuermh: How do we provide `conf` to the `VCFInFormatter`?

fnothaft: It gets it through the `rdd.context.hadoopConfiguration`.

heuermh:
Thank you, @fnothaft

Issue #1597, opened 2017-07-10
RFC: Removing Spark 1.x, Scala 2.10 support in 0.24.0 release


fnothaft:
[From the spark-dev list, it looks like Spark 2.3.0 will add support for Scala 2.12 and drop support for Scala 2.10.](http://apache-spark-developers-list.1001551.n3.nabble.com/RFC-removing-Scala-2-10-td21116.html) As such, I think that we should drop Spark 1.x and Scala 2.10 support in the first ADAM release after Spark 2.3.0, which for now, I am assuming to be ADAM 0.24.0. Thoughts?

heuermh:
+1

For me, the stragglers are cgcloud and dockerized Spark for Toil, both of which are still our responsibility to some extent.  I haven't heard from any external users still wanting Spark 1.x.

antonkulaga:
+1

antonkulaga:
Maybe for 0.23 release, you can at least make Spark 2 and Scala 2.11 the default one so people will not need to run 2.11/spark2 bash scripts each time the clone adam?

heuermh:
> Maybe for 0.23 release, you can at least make Spark 2 and Scala 2.11 the default one...

Sorry, misread this originally.  We don't want to change the release scripts before the 0.23.0 release, so Spark 1.x/Scala 2.10 will remain the default.

PR #1859, opened 2018-01-04
[ADAM-1858] Use getattr instead of Class.forName from python API.

### LINKED ISSUE #1858: `_jvm.java.lang.Class.forName` does not work for certain configurations
This impacts the python `transmute` API. Specifically, `transmute` gets the converter between two types by calling `sc._jvm.java.lang.Class.forName` and constructing the class. This does not work for all py4j setups, but I'm not entirely sure why. However, doing:

```
getattr(sc._jvm, "path.to.my.class")
```

Does seem to work instead.

fnothaft:
Resolves #1858.

### DIFF
--- a/adam-python/bdgenomics/adam/rdd.py
+++ b/adam-python/bdgenomics/adam/rdd.py
@@ -18,6 +18,7 @@
 
 import logging
 
+from py4j.java_gateway import get_java_class
 from pyspark.rdd import RDD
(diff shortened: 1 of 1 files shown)


fnothaft:
Needs #1597.

fnothaft:
Rebased.

heuermh:
Thank you, @fnothaft

PR #1865, opened 2018-01-09
Add command line option for populating nested variant.annotation field in Genotype records.

heuermh:


fnothaft:
Merged! Thanks @heuermh!

PR #1881, opened 2018-01-22
[ADAM-1875] Add ability to timeout a piped command.

### LINKED ISSUE #1875: Allow piped commands to timeout
This is useful for some commands where runtime is not consistent across all regions.

fnothaft:
Resolves #1875. Adds an optional timeout parameter to the `GenomicRDD.pipe` function. If this is provided, a command will be killed (but without failing the Spark task) if it runs for longer than the timeout parameter.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
@@ -454,6 +454,10 @@ trait GenomicRDD[T, U <: GenomicRDD[T, U]] extends Logging {
    * @param environment A map containing environment variable/value pairs to set
    *   in the environment for the newly created process. Default is empty.
    * @param flankSize Number of bases to flank each command invocation by.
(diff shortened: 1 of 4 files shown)


adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -454,21 +454,27 @@ trait GenomicRDD[T, U <: GenomicRDD[T, U]] extends Logging {
    * @param environment A map containing environment variable/value pairs to set
    *   in the environment for the newly created process. Default is empty.
    * @param flankSize Number of bases to flank each command invocation by.
+   * @param optTimeout An optional parameter specifying how long to let a single
```
heuermh: Add doc that this is in seconds, or alternatively allow the caller to specify the `TimeUnit`

fnothaft:
Jenkins, test this please.

heuermh:
Thank you, @fnothaft

PR #1897, opened 2018-02-01
Update for  Spark version 2.3.0

jpdna:
Adds Kryo registration for two classes which is are required for Spark 2.3.0.rc2
With these changes ADAM compiles and passes all tests for Spark-2.3.0.rc2

~~These changes fail to compile with Spark-2.2.1, so must be accompanied by Spark version bump.~~

 If there is a better way to note these changes than a PR against master, let me know.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/serialization/ADAMKryoRegistrator.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/serialization/ADAMKryoRegistrator.scala
@@ -283,9 +283,11 @@ class ADAMKryoRegistrator extends KryoRegistrator with Logging {
     try {
       kryo.register(Class.forName("org.apache.spark.internal.io.FileCommitProtocol$TaskCommitMessage"))
       kryo.register(Class.forName("org.apache.spark.sql.execution.datasources.FileFormatWriter$WriteTaskResult"))
+      kryo.register(Class.forName("org.apache.spark.sql.execution.datasources.BasicWriteTaskStats"))
+      kryo.register(Class.forName("org.apache.spark.sql.execution.datasources.ExecutedWriteSummary"))
     } catch {
       case cnfe: java.lang.ClassNotFoundException => {
-        log.info("Did not find Spark internal class. This is expected for Spark 1.")
+        log.info("Did not find Spark internal class. This is expected for earlier Spark versions.")
       }
     }
     kryo.register(classOf[org.apache.spark.sql.catalyst.expressions.UnsafeRow])


heuermh:
This sucks.  Do you think it might be enough to hold up the 2.3.0 RC process?

heuermh:
E.g. add them to this block
https://github.com/jpdna/adam/blob/e57133aa5ced41d0f36b2fb14f172c9a1409e829/adam-core/src/main/scala/org/bdgenomics/adam/serialization/ADAMKryoRegistrator.scala#L284

and then update the `log.info` message to say "...earlier Spark versions".

jpdna:
Now this works for Spark 2.3.0rc2 and also works for Spark 2.2.1
Thanks @heuermh - I put in `try` as you suggested

jpdna:
I think there is a Jenkins test configuration error of some kind causing build to fail above, but I don't know how to fix it.
I've confirmed again that all tests pass when I run locally, with the Spark 2.2.1 dependency.

fnothaft:
@shaneknapp can you take a look at these builds in @AmplabJenkins? We're getting errors checking out the branch.

heuermh:
Note this can't be merged until 2.3.0 is released; voting is still taking place on RC2, which will fail. RC3 may be available next week.

adam-core/src/main/scala/org/bdgenomics/adam/serialization/ADAMKryoRegistrator.scala
```diff
@@ -283,9 +283,11 @@ class ADAMKryoRegistrator extends KryoRegistrator with Logging {
     try {
       kryo.register(Class.forName("org.apache.spark.internal.io.FileCommitProtocol$TaskCommitMessage"))
       kryo.register(Class.forName("org.apache.spark.sql.execution.datasources.FileFormatWriter$WriteTaskResult"))
+      kryo.register(Class.forName("org.apache.spark.sql.execution.datasources.BasicWriteTaskStats"))
+      kryo.register(Class.forName("org.apache.spark.sql.execution.datasources.ExecutedWriteSummary"))
     } catch {
       case cnfe: java.lang.ClassNotFoundException => {
-        log.info("Did not find Spark internal class. This is expected for Spark 1.")
+        log.info("Did not find Spark internal class. This is expected for earlier Spark versions.")
```
akmorrow13: Can you say spark < 2.3.0 or is that incorrect?

jpdna: So the first two `kryo.register` in this `try`  were already there from a Spark1 to Spark2 incompatibility and the two I add i this PR are for a <2.3 to a 2.3 incompatibility.
But I think just saying <2.3 as you suggest is probably fine, will do unless @heuermh comments otherwise.

shaneknapp:
it looks like it's able to check the branch out now...?

adam-core/src/main/scala/org/bdgenomics/adam/serialization/ADAMKryoRegistrator.scala
heuermh: TBH the errors with Spark 1 won't happen any more, since we don't support it.  There are a few other places in this file that could be updated as well, but that is a low priority.

heuermh:
Moved to 0.25.0 Milestone, can't have 0.24.0 wait for the Spark 2.3.0 release.

heuermh:
Sorry for the noise, I see that the pom.xml change is no longer in this pull request.

heuermh:
Jenkins, retest this please.

heuermh:
Thank you, @jpdna!

PR #1899, opened 2018-02-02
[ADAM-1898] Support converting INFO fields set to empty ('.').

### LINKED ISSUE #1898: Support INFO fields set to missing
While the best practice for missing VCF INFO fields is to omit them from the INFO column for the variant with the missing field, the VCF spec is actually quite loose about this being the correct way to do things:

> The exact format of each INFO sub-field should be specified in the meta-information (as described above). Example for an INFO field: DP=154;MQ=52;H2. Keys without corresponding values are allowed in order to indicate group membership (e.g. H2 indicates the SNP is found in HapMap 2). It is not necessary to list all the properties that a site does NOT have, by e.g. H2=0. See below for additional reserved INFO sub-fields used to encode structural variants.

We throw if we see an INFO field with a missing value `.`.

fnothaft:
Resolves #1898. Also fixes an issue with converting negative phred values.


heuermh:
Thank you, @fnothaft

Issue #1719, opened 2017-09-12
Confirm all I/O resources created by ADAM are closed properly


heuermh:
From discussion re: https://github.com/HadoopGenomics/Hadoop-BAM/pull/152

fnothaft:
@heuermh I'd love to see this in 0.24.0, but realistically I don't know if anyone is planning to take it. Should this bump to 0.25.0?

heuermh:
+1 to 0.25.0

PR #1901, opened 2018-02-03
[ADAM-1900] Close stream after reading VCF header.

### LINKED ISSUE #1900: InputStream passed to VCFHeaderReader does not get closed
See https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala#L1153. This causes connection pool exhaustion when reading a large number of VCFs directly from S3.

fnothaft:
Resolves #1900.


heuermh:
Progress towards #1719.

heuermh:
Thank you, @fnothaft

PR #1878, opened 2018-01-19
[ADAM-651] Hive-style partitioning of parquet files by genomic position

jpdna:
Fixes #651

Manually merged changes for the "hive-style" partitioning branch as a single commit on top of master.

PR #1911, opened 2018-02-13
[ADAM-651] Hive-style partitioning of parquet files by genomic position

### LINKED ISSUE #651: Support Hive-style partitioning

jpdna:
Fixes #651.

Uses genomic range binning to write partitioned parquet files, readable by Spark dataset API.
Significantly improves latency when reading filtering by genomic regions.
Replaces #1878


jpdna:
Please review @heuermh, @akmorrow13 and others, and let me know changed needed.

todo: Have not yet resolved the documented ambiguous behavior if only some directories in a glob are partitioned.  I suggest we leave with the documented ambiguous behavior if merge is possible for v24 this weeks and update later.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/AlignmentRecordRDD.scala
```diff
@@ -26,7 +26,7 @@ import java.nio.file.Paths
 import org.apache.hadoop.fs.Path
 import org.apache.hadoop.io.LongWritable
 import org.apache.parquet.hadoop.metadata.CompressionCodecName
-import org.apache.spark.SparkContext
+import org.apache.spark.{ SparkContext }
```
akmorrow13: remove brackets

jpdna:
My recent commits here are an effort to better deal with the parameter `partitionedBinSize`.
This is a parameter that the user can set at write time defining the genomic bin size used when writing partitioned parquet datasets.

Prior to the previous commit, user would need to keep track of what partition size they used when they wrote a dataset, if they used other than the default of 1 mb. 
This value is required at read time in order to form the query that takes advantage of the binning to improve performance.   

The tracking issue is solved now by writing the partition size as an integer to the flag file `_isPartitionedByStartPos` which is written when a partitioned dataset is saved.

The next goal I have is for user at read time to not need to retrieve or supply the partitionBinSize when working with the dataset, because this is already set for the dataset at write time and the dataset should know its own partitionBinSize and use it.  For example, `filterDatasetByOverlappingRegion` in the public API should not need a partionBinSize parameter.

The best way I see to deal with this is to have `partitionedBinSize` be a member of the DatasetBound object like `DatasetBoundAlignmentRecordRDD` as it is a property of the persisted dataset.
I've implemented that here, and it works fine.

My concern though is that this a specific implementation detail for partitioned parquet backed datasets, and we may not want it adding noise to the  `DatasetBoundAlignmentRecordRDD` constructor even as an optional parameter.

Let me know if you see a better way.

pinging @fnothaft and all for this and general review

jpdna:
5c29dde passes all tests for me locally, any ideas why it fails above?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -1866,6 +1866,35 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+   * @return Returns an AlignmentRecordRDD.
+   */
+  def loadPartitionedParquetAlignments(pathName: String,
+                                       regions: Iterable[ReferenceRegion] = Iterable.empty): AlignmentRecordRDD = {
+
+    val partitionedBinSize = getPartitionedBinSize(pathName)
+    val reads = loadParquetAlignments(pathName)
+
+    val datasetBoundAlignmentRecordRDD = if (regions.nonEmpty) {
+      DatasetBoundAlignmentRecordRDD(reads.dataset, reads.sequences, reads.recordGroups, reads.processingSteps, Some(partitionedBinSize))
```
fnothaft: Nit: Long line, should be broken up.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -1866,6 +1866,35 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+        .filterByOverlappingRegions(regions)
+    } else {
+      DatasetBoundAlignmentRecordRDD(reads.dataset, reads.sequences, reads.recordGroups, reads.processingSteps)
```
fnothaft: Nit: Long line, should be broken up.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2249,6 +2278,30 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+   * @return Returns a GenotypeRDD.
+   */
+  def loadPartitionedParquetGenotypes(pathName: String,
+                                      regions: Iterable[ReferenceRegion] = Iterable.empty): GenotypeRDD = {
+
+    val partitionedBinSize = getPartitionedBinSize(pathName)
+    val genotypes = loadParquetGenotypes(pathName)
+
+    val datasetBoundGenotypeRDD = if (regions.nonEmpty) {
+      DatasetBoundGenotypeRDD(genotypes.dataset, genotypes.sequences, genotypes.samples, genotypes.headerLines, Some(partitionedBinSize))
```
fnothaft: Nit: Long line, should be broken up.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2249,6 +2278,30 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+        .filterByOverlappingRegions(regions)
+    } else {
+      DatasetBoundGenotypeRDD(genotypes.dataset, genotypes.sequences, genotypes.samples, genotypes.headerLines, Some(partitionedBinSize))
```
fnothaft: Nit: Long line, should be broken up.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2282,6 +2335,30 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+   * @return Returns a VariantRDD
+   */
+  def loadPartitionedParquetVariants(pathName: String,
+                                     regions: Iterable[ReferenceRegion] = Iterable.empty): VariantRDD = {
+
+    val partitionedBinSize = getPartitionedBinSize(pathName)
+    val variants = loadParquetVariants(pathName)
+
+    val datasetBoundVariantRDD = if (regions.nonEmpty) {
+      DatasetBoundVariantRDD(variants.dataset, variants.sequences, variants.headerLines, Some(partitionedBinSize))
```
fnothaft: Nit: Long line, should be broken up.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2282,6 +2335,30 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+        .filterByOverlappingRegions(regions)
+    } else {
+      DatasetBoundVariantRDD(variants.dataset, variants.sequences, variants.headerLines, Some(partitionedBinSize))
```
fnothaft: Nit: Long line, should be broken up.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2598,6 +2675,30 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+   * @return Returns a FeatureRDD.
+   */
+  def loadPartitionedParquetFeatures(pathName: String,
+                                     regions: Iterable[ReferenceRegion] = Iterable.empty): FeatureRDD = {
+
+    val partitionedBinSize = getPartitionedBinSize(pathName)
+    val features = loadParquetFeatures(pathName)
+
+    val datasetBoundFeatureRDD = if (regions.nonEmpty) {
+      DatasetBoundFeatureRDD(features.dataset, features.sequences, Some(partitionedBinSize))
```
fnothaft: Nit: Long line, should be broken up.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2598,6 +2675,30 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+        .filterByOverlappingRegions(regions)
+    } else {
+      DatasetBoundFeatureRDD(features.dataset, features.sequences, Some(partitionedBinSize))
```
fnothaft: Nit: Long line, should be broken up.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2630,6 +2731,30 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+   * @return Returns a NucleotideContigFragmentRDD
+   */
+  def loadPartitionedParquetContigFragments(pathName: String,
+                                            regions: Iterable[ReferenceRegion] = Iterable.empty): NucleotideContigFragmentRDD = {
+
+    val partitionedBinSize = getPartitionedBinSize(pathName)
+    val contigs = loadParquetContigFragments(pathName)
+
+    val datasetBoundNucleotideContigFragmentRDD = if (regions.nonEmpty) {
+      DatasetBoundNucleotideContigFragmentRDD(contigs.dataset, contigs.sequences, Some(partitionedBinSize))
```
fnothaft: Nit: Long line, should be broken up.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -2630,6 +2731,30 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+        .filterByOverlappingRegions(regions)
+    } else {
+      DatasetBoundNucleotideContigFragmentRDD(contigs.dataset, contigs.sequences, Some(partitionedBinSize))
```
fnothaft: Nit: Long line, should be broken up.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -3051,4 +3176,27 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+   * @return Return true if the specified path of Parquet + Avro files is partitioned.
+   *
+   * If a glob is used, all directories within the blog must be partitioned, and must have been saved
+   * using the same partitioned bin size.  Behavior is undefined if this requirement is not met.
+   */
+  def getPartitionedBinSize(pathName: String): Int = {
+
+    val partitionSize = try {
+      val f = getFsAndFilesWithFilter(pathName, new FileFilter("_isPartitionedByStartPos")).head
+      f.getFileSystem(sc.hadoopConfiguration).open(f).readInt()
```
fnothaft: You need to close the stream after you open it. E.g.:

```
val is = f.getFileSystem(sc.hadoopConfiguration).open(f)
val bucketSize = is.readInt()
is.close()
```

Additionally, I'd prefer that we enforced that all partitioned files had the same bin size. E.g.,

```
val partitionSizes = getFsAndFilesWithFilter(pathName, new FileFilter("_isPartitionedByStartPos")).map(f => {
    val is = f.getFileSystem(sc.hadoopConfiguration).open(f)
    val partitionSize = is.readInt()
    is.close()
    partitionSize
  }).toSet

require(partitionSizes.nonEmpty, "Input Parquet files (%s) are not partitioned.".format(pathName))
require(partitionSizes.size == 1, "Found multiple partition sizes (%s).".format(partitionSizes.mkString(", "))

partitionSizes.head
```

You shouldn't ever actually get a `FileNotFoundException`; if there's no `_isPartitionedByStartPos` files, then `getFsAndFilesWithFilter` shouldn't return any files.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
```diff
@@ -3051,4 +3176,27 @@ class ADAMContext(@transient val sc: SparkContext) extends Serializable with Log
...
+  /**
+   * Return true if the specified path of Parquet + Avro files is partitioned.
+   *
+   * @param pathName Path in which to look for partitioned flag.
+   * @return Return true if the specified path of Parquet + Avro files is partitioned.
+   *
+   * If a glob is used, all directories within the blog must be partitioned, and must have been saved
+   * using the same partitioned bin size.  Behavior is undefined if this requirement is not met.
+   */
+  def getPartitionedBinSize(pathName: String): Int = {
```
fnothaft: This function should be private, or at least `private[rdd]`.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -2531,6 +2546,23 @@ abstract class AvroGenomicRDD[T <% IndexedRecord: Manifest, U <: Product, V <: A
...
+        "\' and positionBin >= \'" + ((scala.math.floor(r.start / partitionSize).toInt) - partitionedLookBackNum) +
+        "\' and positionBin < \'" + (scala.math.floor(r.end / partitionSize).toInt + 1) +
+        "\' and (end > " + r.start + " and start < " + r.end + "))")
+        .mkString(" or ")
+    } else { // if no positionBin field is found then construct query without bin optimization
+      regions.map(r => "(contigName=" + "\'" + r.referenceName +
+        "\' and (end > " + r.start + " and start < " + r.end + "))")
+        .mkString(" or ")
+    }
+
```
fnothaft: Nit: extra whitespace.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -2324,6 +2326,19 @@ trait GenomicDataset[T, U <: Product, V <: GenomicDataset[T, U, V]] extends Geno
...
+  /**
+   * Filters and replaces the underlying dataset based on overlap with any of a Seq of ReferenceRegions.
+   *
+   * @param querys ReferencesRegions to filter against
+   * @param optPartitionedLookBackNum Optional number of parquet position bins to look back to find start of a
+   *                                  ReferenceRegion, defaults to 1
+   * @return Returns a new DatasetBoundGenomicRDD with ReferenceRegions filter applied.
+   */
+  def filterDatasetByOverlappingRegions(querys: Iterable[ReferenceRegion],
```
fnothaft: I'd looove this to move into the new `DatasetBoundGenomicDataset` trait which came in @akmorrow13's https://github.com/bigdatagenomics/adam/commit/60513214660b4f7de9f38c6df9cdd8ede9b25a47, and then this would override `filterByOverlappingRegions`. However, this has a slightly different signature due to the addition of `optPartitionedLookBackNum`. I propose that we either:

* Move this into `DatasetBoundGenomicDataset` and make this method protected, and then override `filterByOverlappingRegions(querys)` to call `filterDatasetByOverlappingRegions(querys)`.
* Or, keep this as is now but open a ticket to close up this API in 0.25.0.

Thoughts?

Unrelated nit: `querys` should be `queries`.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
```diff
@@ -2531,6 +2546,23 @@ abstract class AvroGenomicRDD[T <% IndexedRecord: Manifest, U <: Product, V <: A
     saveSequences(filePath)
   }
 
+  protected def referenceRegionsToDatasetQueryString(regions: Iterable[ReferenceRegion], partitionSize: Int = 1000000, partitionedLookBackNum: Int = 1): String = {
```
fnothaft: I see that a lot of downstream classes that implement `GenomicDataset` wind up overriding `filterDatasetByOverlappingRegions` with `transformDataset(ds => ds.filter(referenceRegionsToDatasetQueryString(...))`. I get why they have to do this, and I think fixing it is out of scope for this PR. However, can you open a ticket for us to move that code into a trait in a future release?

adam-core/src/main/scala/org/bdgenomics/adam/rdd/fragment/FragmentRDD.scala
```diff
@@ -227,6 +228,20 @@ case class DatasetBoundFragmentRDD private[rdd] (
...
+   *
+   * @param querys ReferencesRegions to filter against
+   * @param optPartitionedLookBackNum Optional number of parquet position bins to look back to find start of a
+   *                                  ReferenceRegion, defaults to 1
+   * @return Returns a new DatasetBoundFragmentRDD with ReferenceRegions filter applied.
+   */
+  override def filterDatasetByOverlappingRegions(querys: Iterable[ReferenceRegion],
+                                                 optPartitionedLookBackNum: Option[Int] = Some(1)): FragmentRDD = {
+    transformDataset((d: Dataset[org.bdgenomics.adam.sql.Fragment]) =>
+      d.filter(referenceRegionsToDatasetQueryString(querys, partitionedBinSize.get, optPartitionedLookBackNum.get)))
```
fnothaft: This won't work for fragment, no? `Fragment` doesn't have `contigName`, `start`, or `end` fields.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
jpdna: Alternatively - upon thinking about it, I don't like this `optPartitionedLookbacknum` as a parameter to the filterByOverlap function anyhow, as this is a config parameter tied to the dataset just like `partitionedBinSize` that should only need be set once when the dataset is created, so we could add it as another optional parameter to the DatasetBoundtype constructor.
 
If we did that, `optPartitionedLookbacknum`,  the parameter could go away making overriding `filterByOverlappingRegions` override in `DatasetBoudGenomicDataset` work cleanly.

fnothaft: SGTM!

adam-core/src/main/scala/org/bdgenomics/adam/rdd/fragment/FragmentRDD.scala
jpdna: you are right, removed Fragment.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/GenomicRDD.scala
jpdna: In update the `filterDatasetByOverlappingRegion` is implemented generically in trait `DatasetBoundGenomicDataset` , thus no more overrideing

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
jpdna: Done

adam-cli/src/main/scala/org/bdgenomics/adam/cli/TransformAlignments.scala
```diff
@@ -562,8 +564,12 @@ class TransformAlignments(protected val args: TransformAlignmentsArgs) extends B
       mergedSd
     }
 
-    outputRdd.save(args,
-      isSorted = args.sortReads || args.sortLexicographically)
+    if (args.saveAsDataset) {
+      outputRdd.saveAsPartitionedParquet(args.outputPath)
```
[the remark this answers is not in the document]
jpdna: It will save all data to partition bin number 0, and doing so will not result in any benefit.
A warning stating that has been added with `log.warn`

adam-cli/src/main/scala/org/bdgenomics/adam/cli/TransformAlignments.scala
```diff
@@ -131,6 +131,8 @@ class TransformAlignmentsArgs extends Args4jBase with ADAMSaveAnyArgs with Parqu
   var storageLevel: String = "MEMORY_ONLY"
   @Args4jOption(required = false, name = "-disable_pg", usage = "Disable writing a new @PG line.")
   var disableProcessingStep = false
+  @Args4jOption(required = false, name = "-save_as_dataset", usage = "EXPERIMENTAL: Save as a Parquet format Spark-SQL dataset")
```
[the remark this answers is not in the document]
jpdna: Resolved - and changed parameter to an integer so user can specify bin size.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala
jpdna: Done, but had to add a separate public `isPartitioned` function as because Mango needs to be able to check to see if data is partitioned.

jpdna:
Replaced by #1922

Issue #1349, opened 2017-01-12
adam-shell does not allow additional jars via Spark jars argument


heuermh:
If I have this right, `adam-shell` does not allow additional jars via Spark `--jars` argument
```bash
$ adam-shell --jars my-jar.jar
...
scala> import org.bdgenomics.adam.rdd.ADAMContext
<console>:24: error: object bdgenomics is not a member of package org
       import org.bdgenomics.adam.rdd.ADAMContext
                  ^
```

`adam-shell` uses `--jars` to specify the ADAM cli jar
```bash
# submit the job to Spark
"$SPARK_SHELL" \
    --conf spark.serializer=org.apache.spark.serializer.KryoSerializer \
    --conf spark.kryo.registrator=org.bdgenomics.adam.serialization.ADAMKryoRegistrator \
    --jars ${ADAM_CLI_JAR} \
    "$@"
```

fnothaft:
IIRC, you can only have `--jars`once, and first one wins.

heuermh:
Right, so how do we fix it?  :)

fnothaft:
Well, I can't promise an answer that you'll like, but I'd say either we mark that as a won't fix, or we parse out a `--jars` argument in the `bin/adam-submit` script.

heuermh:
If one tries to cheat by specifying a custom jar and the adam fat jar (here coming from a homebrew installation) in `--jars`, it appears ok for a while but does not work out in the end
```
$ adam-shell --jars target/dsh-bio-adam-1.0-SNAPSHOT.jar:/usr/local/Cellar/adam/0.21.0/libexec/repo/adam_2.11-0.21.0.jar
Using SPARK_SHELL=/usr/local/bin/spark-shell
...
scala> import org.dishevelled.bio.adam.DishevelledAdamContext
import org.dishevelled.bio.adam.DishevelledAdamContext

scala> import org.bdgenomics.adam.rdd.ADAMContext
import org.bdgenomics.adam.rdd.ADAMContext

scala> val dac = new DishevelledAdamContext(sc)
dac: org.dishevelled.bio.adam.DishevelledAdamContext = org.dishevelled.bio.adam.DishevelledAdamContext@4a977146

scala> val features = dac.loadGff3("/Users/heuermh/working/adam/adam-core/src/test/resources/dvl1.200.gff3")
[...]

PR #1918, opened 2018-02-16
Add Google Cloud documentation

Georgehe4:


heuermh:
Jenkins, test this please.

docs/deploying/gcp.rst
```diff
@@ -0,0 +1,73 @@
...
+
+The prebuilt jar can be found at `https://oss.sonatype.org/content/repositories/releases/com/google/cloud/google-cloud-nio/0.22.0-alpha/google-cloud-nio-0.22.0-alpha-shaded.jar <https://oss.sonatype.org/content/repositories/releases/com/google/cloud/google-cloud-nio/0.22.0-alpha/google-cloud-nio-0.22.0-alpha-shaded.jar>`__
+
+::
+  
+    adam-submit \
+      transformAlignments \
+      gs://my_bucket/my_reads.adam \
+      gs://my_bucket/my_new_reads.adam \
+      --jars google-cloud-nio-0.22.0-alpha-shaded.jar
```
heuermh: Since this jar is in Maven Central, would `--package` be more appropriate?  Note also we have an issue with `adam-shell` that prevents using `--jars` (https://github.com/bigdatagenomics/adam/issues/1349).

fnothaft: +1 towards using `--packages`

Georgehe4:
Adding the dependency via --packages would be ideal except that the NIO file readers rely on Guava 20.0+.

hadoop 2.6+ breaks with guava 17.0+ (https://github.com/bigdatagenomics/adam/blob/49cbdb72e7b4090d4b2840dc72d06b9c68618a3a/pom.xml#L600)

So I am using the shaded jar is the only way to prevent naming conflicts.

I don't believe we can shade using the --packages flag, but let me know if anyone knows of a workaround.

Georgehe4:
To clarify, the error that pops up when using --packages is: https://gist.github.com/Georgehe4/2897f2ead685a9fcf014cbf01cee4375.

Root cause:
https://stackoverflow.com/questions/45247193/nosuchmethoderror-com-google-common-base-preconditions-checkargumentzljava-lan

Georgehe4:
Running with --jars solves the above issue:
https://gist.github.com/Georgehe4/a6ff7e37e3a8d78a3ca7dd52846bed17

fnothaft:
Thanks for the detailed reply and gists @Georgehe4! It looks like `--jars` is working fine for you, that had been broken for a while earlier; apparently we fixed it without noticing it! This is good to go from my side, ping @heuermh to approve and merge.

heuermh:
> It looks like --jars is working fine for you, that had been broken for a while earlier; apparently we fixed it without noticing it

`--jars` works ok for `adam-submit` but not for `adam-shell`.

heuermh:
Thank you, @Georgehe4

PR #1929, opened 2018-02-28
[ADAM-1927] Increase visibility for util.FileExtensions for use downstream.

### LINKED ISSUE #1927: util.FileExtensions should be public for use downstream in Cannoli

heuermh:
Fixes #1927


fnothaft:
Merged! Thanks @heuermh!

PR #1935, opened 2018-03-03
Ignore VCF INFO fields with number=G when stringency=LENIENT

jpdna:
Ignores, and logs a warning, for  non-specifically-supported INFO fields with Number=G when Stringency = LENIENT
rather than  the current behavior of throwing IllegalArgumentException

Needed because I want to be able to parse directly the gnomAD VCF file contains tags such as:
```
INFO=<ID=GC_Male,Number=G,Type=Integer,Description="Count of Male individuals for each genotype">
```
These INFO number=G tags don't have a simple/general  meaning for multi-allelic variants which are split when loaded to ADAM, which is why in status quo this produces `IllegalArgumentException("Number=G INFO lines are not supported in split-allelic model: %s".` 
Simply allowing the user to choose to ignore INFO tags of number=G seems a workable solution to me.

Aside: 
At the least, there is not a general solution for INFO Number=G fields that can be implemented in `lineToVariantContextExtractor`.
If we do decide that we want to do something specific for an INFO Number=G  tag like `GC_Male` we can consider adding it as a supported tag and adding custom logic to parse it later.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantContextConverter.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantContextConverter.scala
@@ -1439,46 +1439,46 @@ class VariantContextConverter(
   }
 
   private def lineToVariantContextExtractor(
-    headerLine: VCFInfoHeaderLine): ((HtsjdkVariantContext, Int, Array[Int]) => Option[(String, String)]) = {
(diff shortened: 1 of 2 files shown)


jpdna:
Can one f you point me to where in the Jenkins output the relevant error is?
Tests pass for me locally.

heuermh:
@jpdna This also passes tests for me locally, will try retriggering the Jenkins build.

heuermh:
Jenkins, retest this please.

fnothaft:
Merged! Thanks @jpdna!

PR #1936, opened 2018-03-03
[ADAM-1934] Clip phred values to 3233, instead of Int.MaxValue.

### LINKED ISSUE #1934: Clip phred scores below Int.MaxValue
While #1569 improved the resolution that our phred score converter covered, it still clips around Phred 3233. We should saturate there, instead of at Int.MaxValue.

fnothaft:
Resolves #1934. When phred values underflow log space double precision floating point numbers, clip them to the lowest phred score representable as a log valued double (3233), instead of Int.MaxValue.


heuermh:
Thank you, @fnothaft!

PR #1937, opened 2018-03-05
[ADAM-1932] Throw file not found exception if directory is empty.

### LINKED ISSUE #1932: ADAMContext.getFsAndFilesWithFilter should throw exception if paths null or empty
Shouldn't this code block be throwing an exception?

```scala
    val paths = if (fs.isDirectory(path)) {
      val paths = fs.listStatus(path)

      if (paths == null || paths.isEmpty) {
        s"Couldn't find any files matching ${path.toUri}"
      }
      fs.listStatus(path, filter)
}
```
https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/rdd/ADAMContext.scala#L1371

heuermh:
Fixes #1932


fnothaft:
Merged! Thanks @heuermh!

PR #1946, opened 2018-03-07
[ADAM-1914] Python profile needs to be specified for egg to be in distribution.

### LINKED ISSUE #1914: cannot run pyadam from adam distribution 0.23.0
generates error org.apache.spark.SparkException: Added file file:adam-distribution-spark2_2.11-0.23.0/repo is a directory and recursive is not turned on.

This is because the adam egg file does not exist in the repo directory.

fnothaft:
Resolves #1914.

### DIFF
--- a/scripts/release/release.sh
+++ b/scripts/release/release.sh
@@ -27,9 +27,18 @@ git push origin ${branch}
 
 # do spark 2, scala 2.11 release
 git checkout -b maint_spark2_2.11-${release} ${branch}
+
+cd adam-python
+virtualenv adam-release
+. adam-release/bin/activate
+make develop
+deactivate
+rm -rf adam-release
+cd ..
+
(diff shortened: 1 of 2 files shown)


scripts/release/release.sh
```diff
@@ -29,7 +29,7 @@ git push origin ${branch}
 git checkout -b maint_spark2_2.11-${release} ${branch}
 git commit -a -m "Modifying pom.xml files for Spark 2, Scala 2.11 release."
 mvn --batch-mode \
-  -P distribution \
+  -P python,r,distribution \
```
heuermh: On further consideration, this won't work, will it?  The only way I get the python and r profiles to build for me locally is to run separate mvn steps after first running mvn install, and then specify all the `PY4J_ZIP PYSPARK_SUBMIT_ARGS` etc. environment variables.

heuermh:
Thank you, @fnothaft

PR #1960, opened 2018-03-22
[ADAM-1959] Support VCF rows at position 0.

### LINKED ISSUE #1959: Support VCF lines with positions at 0
According to the VCF spec, these are admissible:

```
2. POS - position: The reference position, with the 1st base having position 1. Positions are sorted numerically,
in increasing order, within each reference sequence CHROM. It is permitted to have multiple records with the
same POS. Telomeres are indicated by using positions 0 or N+1, where N is the length of the corresponding
chromosome or contig. (Integer, Required)
```

We allow these in our schema and converter, but we create an invalid reference position for them.

fnothaft:
Resolves #1959.


heuermh:
Thank you, @fnothaft

PR #1961, opened 2018-03-22
[ADAM-1957] Don't initialize missing likelihoods to MinValue.

### LINKED ISSUE #1957: Don't initialize non-ref values to Int.MinValue
https://github.com/bigdatagenomics/adam/blob/master/adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantContextConverter.scala#L1077

fnothaft:
Resolves #1957. Introduces a PhredUtils maximum value constant.


heuermh:
Thank you, @fnothaft

PR #1965, opened 2018-03-23
[ADAM-1964] Lower point where phred conversions are done using log code.

### LINKED ISSUE #1964: Phred values from 156–254 do not round trip properly between log space

fnothaft:
Resolves #1964. There was an underflow somewhere causing round trip from log space values to Phred to mismatch for values between 164 and 254. This change lowers the point where we switch from using the precomputed table to where we use the log math. Also, we add a unit test checking round trip values using PhredUtils.


heuermh:
Thank you, @fnothaft

PR #1966, opened 2018-03-26
Adding a BamLoader class to have only 1 header parse for multiple ind…

ffinfo:
Adding a BamLoader class to have only 1 header parse for multiple independent regions


ffinfo:
This also reduces some code duplication for header validation.
The warning from codacy what already in there at an other.

heuermh:
Closing as WontFix for now.  Such a change might be useful to consider when migrating to Disq (#2111).

PR #1968, opened 2018-03-26
Fix partitioning code to use Long instead of Int

jpdna:
Change to use Long instead of Int


fnothaft:
Closing in favor of #1980.

PR #1998, opened 2018-06-13
[ADAM-1997] Nest list of region join types in joins doc.

### LINKED ISSUE #1997: Unordered list of region join types in doc is missing nested levels
This unordered list in `joins.rst` is missing nested levels:

ADAM has a variety of region join types that you can perform on your data, and all are called in a similar way:

- Joins implemented across both shuffle and broadcast
- Inner join
- Right outer join
- Shuffle-only joins
- Full outer join
- Inner join and group by left
- Left outer join
- Right outer join and group by left
- Broadcast-only joins
- Inner join and group by right
- Right outer join and group by right

heuermh:
Fixes #1997.

### DIFF
--- a/docs/api/joins.rst
+++ b/docs/api/joins.rst
@@ -55,16 +55,21 @@ ADAM has a variety of region join types that you can perform on your
 data, and all are called in a similar way:
 
 -  Joins implemented across both shuffle and broadcast
--  Inner join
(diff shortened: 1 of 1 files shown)


fnothaft:
Merged! Thanks @heuermh!

PR #2008, opened 2018-07-04
[ADAM-2006] Support loading 0-length reads as FASTQ.

### LINKED ISSUE #2006: Support 0 length FASTQ reads
These can come out of trimming tools like CutAdapt, and are technically valid reads. We currently barf on them.

fnothaft:
Resolves #2006.


heuermh:
Thank you, @fnothaft!

PR #2010, opened 2018-07-05
[ADAM-1697] Expand Illumina metadata regex to cover interleaved index sequences.

### LINKED ISSUE #1697: FASTQ input format needs to support index sequences
Related to #1585. There's another Illumina FASTQ spec out there that puts the index sequences in the read name. From the ["home of the FASTQ spec"](https://en.wikipedia.org/wiki/FASTQ_format#Illumina_sequence_identifiers):

```
@EAS139:136:FC706VJ:2:2104:15343:197393 1:Y:18:ATCACG
```

I've also seen something like this for paired sequences:

```
@EAS139:136:FC706VJ:2:2104:15343:197393 1:Y:18:ATCACG+ATCACG
```

This is close to but slightly different from the variant in #1585.

heuermh:
Fixes #1697.  Thank you to @zhangnew for the suggestion.


fnothaft:
Merged! Thanks @heuermh!

PR #2019, opened 2018-07-22
Don't discard unmapped reads in indel realignment

pauldwolfe:
When doing some QC on ADAM outputs we noticed unmapped reads were dropped in indel realignment. We need to keep all reads throughout our pipeline, mapped or no.

Improvement comments on the implementation welcome, its a bit quick and dirty but at least gives us something to talk about. The idea now is to just union the unmapped reads when realignment is complete. Could sort at this point as well but decided to leave that to the client. I'm not sure indel realignment returns a sorted RDD in the first place.


pauldwolfe:
Just realized I made a bit of a mess of this pull request by using an old fork. Let me know if we can salvage this one or if you'd like me to make a new fork from scratch

heuermh:
Thank you for the contribution, @pauldwolfe!

No where in the scaladoc does it say that unmapped reads are filtered out; we should certainly add that.

Can the select for unmapped reads and later union happen outside the call to `realignIndels`?  I wonder if that might be cleaner than adding another flag to the method call.

pauldwolfe:
Hey @heuermh , see your point, adding more options and conditionals always increases complexity. I don't mind pulling the fix out into my code, it's pretty straightforward.  

That said, you might want to consider going the other direction, and having realign indels always return the unmapped reads. Looking at the commit that introduced this behaviour, seems like an internal performance improvement with a functional side-effect rather than an intentional behaviour change. Also from a client perspective it's easier to filter the unmapped after realignment than to do the filter + union.

Just a thought, let me know what you think. Otherwise I'll go ahead and handle in my own code.

fnothaft:
Jenkins, test this please.

heuermh:
Closing as WontFix; doc fixes should be handled separately (#2022).

PR #2045, opened 2018-09-07
[ADAM-2023] Implemented Duplicate Marking algorithm in Spark SQL

### LINKED ISSUE #2023: Mark Duplicates implemented in Spark SQL
I believe that there are performance improvements to be had by implementing duplicate marking in Spark SQL.

jonpdeaton:
With these changes, I see as much as a 30% speedup for large datasets. Fixes #2023

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/MarkDuplicates.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/MarkDuplicates.scala
@@ -214,9 +214,12 @@ private[rdd] object MarkDuplicates extends Serializable with Logging {
   private def findDuplicates(fragmentDf: DataFrame): DataFrame = {
     import fragmentDf.sparkSession.implicits._
 
+    val filteredDf = fragmentDf
+      .filter('read1contigName.isNotNull and 'read1fivePrimePosition.isNotNull and 'read1strand.isNotNull)
+
     // this DataFrame has an extra column "groupCount" which is the number of distinct
     // right reference positions for fragments grouped by left reference position
-    val withGroupCount = calculateGroupCounts(fragmentDf)
+    val withGroupCount = calculateGroupCounts(filteredDf)
 
     // Window into fragments grouped by left and right reference positions
     val positionWindow = Window.partitionBy(
@@ -228,10 +231,8 @@ private[rdd] object MarkDuplicates extends Serializable with Logging {
     // duplicates are those fragments which are not the highest scoring fragment among those with the same
     // left and right reference positions or those with unmapped right position and group count is equal to zero
     val duplicatesDf = withGroupCount.withColumn("duplicateFragment",
-      ('read1contigName.isNotNull and 'read1fivePrimePosition.isNotNull and 'read1strand.isNotNull)
-        and (
-          row_number.over(positionWindow) =!= 1
-          or ('read2contigName.isNull and 'read2fivePrimePosition.isNull and 'read2strand.isNull and 'groupCount > 0)))
+      row_number.over(positionWindow) =!= 1
+        or ('read2contigName.isNull and 'read2fivePrimePosition.isNull and 'read2strand.isNull and 'groupCount > 0))
 
     // result is just the relation between fragment and duplicate status
     duplicatesDf.select("recordGroupName", "readName", "duplicateFragment")
@@ -266,6 +267,7 @@ private[rdd] object MarkDuplicates extends Serializable with Logging {
 
   /**
    * Calculates the number of distinct right positions for each group of left reference positions
+   *
    * @param fragmentsDf DataFrame of fragments containing left and right reference positions
    * @return DataFrame indicating for each left-reference position the group "groupCount" which is equal to the
    *         number of distinct right reference positions among all fragments with the same left
@@ -283,6 +285,7 @@ private[rdd] object MarkDuplicates extends Serializable with Logging {
    * from a DataFrame specifying which fragments found in the collection of alignment records are duplicates.
    * Each read will be marked as a duplicate if it is a primary alignments and part of a duplicate fragment
    * or if it is a mapped read but not a primary alignment.
+   *
    * Unmapped reads will not be marked as duplicate.
    * @param alignmentRecords Dataset of AlignmentRecords
    * @param duplicatesDf DataFrame containing information about each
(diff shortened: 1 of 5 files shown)


heuermh:
Jenkins, test this please

heuermh:
Jenkins, add to whitelist

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/MarkDuplicates.scala
```diff
@@ -17,153 +17,419 @@
...
+  /**
+   * Marks alignment records as PCR duplicates.
+   * This class marks duplicates all read pairs that have the same pair alignment locations,
+   * and all unpaired reads that map to the same sites. Only the highest scoring
+   * read/read pair is kept, where the score is the sum of all quality scores minus 33 in
+   * the read that are greater than 15.
+   * @param alignmentRecords GenomicRDD of alignment records
+   * @return RDD of alignment records with the "duplicateRead" field marked appropriately
+   */
+  def apply(alignmentRecords: AlignmentRecordRDD): RDD[AlignmentRecord] = {
```
heuermh: We should refactor the caller here so that we don't force the conversions between RDD and Dataset.  In other words, only perform the conversions if `alignmentRecords` has been realized as `RDDBoundAlignmentRecordRDD`.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/MarkDuplicates.scala
```diff
@@ -17,153 +17,419 @@
...
+    import alignmentRecords.sparkSession.implicits._
+    alignmentRecords.join(duplicatesDf, Seq("readName", "recordGroupName"), "left")
   }
 
-  def apply(rdd: FragmentRDD): RDD[Fragment] = {
+  /**
+   * Case class which merely extends the Fragment Schema by a single column "duplicateFragment" so that
+   * a DataFrame with fragments having been marked as duplicates can be cast back into a DataSet
+   */
+  private case class FragmentDuplicateSchema(readName: Option[String] = None,
```
heuermh: Do you think it would be useful to add `duplicateFragment` or similarly named field to the Avro schema definition for `Fragment`, or is this flag only useful in a temporary context?

jondeaton: I'm not sure. I was considering it while developing but then I realized that it really only seems useful in a temporary context since each of the alignment records contained within the fragment also contain a duplicate flag so the schema has redundancies.

jonpdeaton: I'm not sure. I was considering it while developing but then I realized that it really only seems useful in a temporary context since each of the alignment records contained within the fragment also contain a duplicate flag so the schema has redundancies.

adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/MarkDuplicates.scala
jonpdeaton: I see, so there should be another `apply` method which takes `Dataset[AlignmentRecord]` and this `apply` method's signature should be changed to take `RDDBoundAlignmentRecordRDD`?

jonpdeaton: Or do you mean that this method should just be changed to take `Dataset[AlignmentRecord]`. In this case would it also make sense to refactor so that the other `apply` method also just takes `Dataset[Fragment]` and do a similar refactoring of the caller?

heuermh: Does this look right?  We would want the caller to be
```scala
abstract class AlignmentRecordRDD ... {
  def markDuplicates(): AlignmentRecordRDD = {
    replaceRdd(MarkDuplicates(this.rdd, this.recordGroups))
  }
}

case class DatasetBoundAlignmentRecordRDD ... {
  override def markDuplicates(): AlignmentRecordRDD = {
    replaceDataset(MarkDuplicates(this.dataset, this.recordGroups))
  }
}

abstract class FragmentRDD ... {
  def markDuplicates(): FragmentRDD = {
    replaceRdd(MarkDuplicates(this.rdd, this.recordGroups))
  }
}

case class DatasetBoundFragmentRDD ... {
  override def markDuplicates(): FragmentRDD = {
    replaceDataset(MarkDuplicates(this.dataset, this.recordGroups))
  }
}
```

so then the `apply` methods might be
```scala
import org.bdgenomics.formats.avro.{
  AlignmentRecord,
  Fragment
}
import org.bdgenomics.adam.sql.{
  AlignmentRecord => AlignmentRecordProduct,
  Fragment => FragmentProduct
}

object MarkDuplicates {
  def apply(RDD[AlignmentRecord], RecordGroupDictionary): RDD[AlignmentRecord] = { }
  def apply(RDD[Fragment], RecordGroupDictionary): RDD[Fragment] = { }
  def apply(Dataset[AlignmentRecordProduct], RecordGroupDictionary): Dataset[AlignmentRecordProduct] = { }
  def apply(Dataset[FragmentProduct], RecordGroupDictionary): Dataset[FragmentProduct] = { }
}
```

jonpdeaton: Alright sounds good. Because of type erasure there will need to be a single `apply[T]` for `RDD` and for `Dataset` for fragments/alignment-records.

heuermh: Ah, good point.  Is `RecordGroupDictionary` actually required for the fragment cases?  If not, that could be the discriminator.

jonpdeaton: I believe that it is.

jonpdeaton: Would it be better to take `RDDBoundAlignmentRecordRDD` and `DatasetBoundAlignmentRecordRDD` so that the code for converting from `RDD` to `Dataset` that is already implemented in `RDDBoundAlignmentRecordRDD` is not duplicated in `MarkDuplicates`?

heuermh:
Closing as WontFix.  Performance testing these changes were inconclusive.  Feel free to create a new PR after rebasing against git head.

Issue #1498, opened 2017-04-19
Compute coverage by sample


fnothaft:
From a set of `AlignmentRecord`s.

heuermh:
Fixed by #2051

PR #2050, opened 2018-09-16
Multi-sample coverage

akmorrow13:


adam-core/src/main/scala/org/bdgenomics/adam/models/Coverage.scala
```diff
@@ -33,8 +33,8 @@ private[adam] object Coverage {
    * @param count Coverage count for each base pair in region
    * @return Coverage spanning the specified ReferenceRegion
    */
-  def apply(region: ReferenceRegion, count: Double): Coverage = {
-    Coverage(region.referenceName, region.start, region.end, count)
+  def apply(region: ReferenceRegion, count: Double, name: Option[String]): Coverage = {
```
akmorrow13: should we have this be source instead of name?

heuermh:
See #1498.

heuermh:
Closing in favor of #2051.

PR #2065, opened 2018-10-15
[ADAM-2064] Avoid collision between @RG FO and KS fields.

### LINKED ISSUE #2064: @RG Flow Order field truncated
SAM format has a read group field called "FO" which lists the order dNTP flows during a sequencing run.

image

For Ion Torrent runs in particular this could be 500 characters or more.  I have seen that saveAsSam is truncating this field to 18 or 19 characters.

fnothaft:
Resolves #2064. We were writing the KS (key sequence) read group attribute back into the FO (flow order) attribute. This commit resolves this error and adds a unit test touching all supported @RG attributes.


heuermh:
Thank you, @fnothaft!

PR #2077, opened 2018-11-05
[ADAM-1992] Make maximum FASTQ read length configurable.

### LINKED ISSUE #1992: Fastq reader clips long reads at 10,000 bp
E.g. ftp://ftp-trace.ncbi.nlm.nih.gov/giab/ftp/data/AshkenazimTrio/HG002_NA24385_son/Ultralong_OxfordNanopore/

```
$ dsh-bio fastq-sequence-length -i combined_2018-05-18.fastq.gz | sort -n -r | head
322099
298433
283861
262525
247501
```

```
scala> val reads = sc.loadAlignments("combined_2018-05-18.fastq.gz")
reads: org.bdgenomics.adam.rdd.read.AlignmentRecordRDD =
ParquetUnboundAlignmentRecordRDD with 0 reference sequences, 0 read groups,
and 0 processing steps

scala> val lengths = reads.rdd.map(_.sequence.length()).collect()
lengths: Array[Int] = Array(641, 3082, 10000, 10000, 10000, ...

scala> lengths.sortBy(-1 * _).take(10)
res3: Array[Int] = Array(10000, 10000, 10000, 10000, 10000, 10000, 10000, 10000, ...
```

heuermh:
Fixes #1992


heuermh:
@akmorrow13 Ping for review and merge

adam-core/src/main/java/org/bdgenomics/adam/io/FastqRecordReader.java
```diff
@@ -60,6 +60,24 @@
      * application.  We'll see if someone complains in other applications.
      */
 
+    /** Default maximum read length, <code>10,000</code> bp. */
+    public static final int DEFAULT_MAX_READ_LENGTH = 10000;
+
+    /** Maximum read length property name. */
+    public static final String MAX_READ_LENGTH_PROPERTY = "org.bdgenomics.adam.io.FastqRecordReader.MAX_READ_LENGTH";
```
akmorrow13: Is there a way to set this in the cli or just the shell?

heuermh: Hmm, good point.

It is a property in the Hadoop configuration, so I suppose
```
adam-submit \
  --conf org.bdgenomics.adam.io.FastqRecordReader.MAX_READ_LENGTH=42 \
  --
```
might work

adam-cli/src/main/scala/org/bdgenomics/adam/cli/TransformAlignments.scala
```diff
@@ -139,6 +140,8 @@ class TransformAlignmentsArgs extends Args4jBase with ADAMSaveAnyArgs with Parqu
   var partitionByStartPos: Boolean = false
   @Args4jOption(required = false, name = "-partition_bin_size", usage = "Partition bin size used in Hive-style partitioning. Defaults to 1Mbp (1,000,000) base pairs).")
   var partitionedBinSize = 1000000
+  @Args4jOption(required = false, name = "-max_read_length", usage = "Maximum FASTQ read length, defaults to 10,000 base pairs (bp).")
+  var maxReadLength: Int = 0
```
akmorrow13: Should the default be 10,000?

heuermh: The default is 10,000, as set in `FastqRecordReader`.
Since it isn't possible to initialize `maxReadLength` to `null`, `0` is used instead.

heuermh:
@akmorrow13 Ping for merge

adam-cli/src/main/scala/org/bdgenomics/adam/cli/TransformAlignments.scala
```diff
@@ -385,6 +388,10 @@ class TransformAlignments(protected val args: TransformAlignmentsArgs) extends B
 
     val stringencyOpt = Option(args.stringency).map(ValidationStringency.valueOf(_))
 
+    if (args.maxReadLength > 0) {
```
akmorrow13: It may make more sense to set the default args.maxReadLength to 10,000 and remove the if statement.

heuermh: If it is set to the default here in the cli class, then it is always set here, and will clobber the configuration property if defined on the command line or elsewhere.

akmorrow13:
Thanks @heuermh!

PR #2084, opened 2018-11-07
added pyspark mock modules for API documentation

akmorrow13:


akmorrow13:
I accidentally had pyspark in my venv. This will hopefully be the last of the mock modules that we need...

heuermh:
Thank you, @akmorrow13

PR #2092, opened 2018-11-16
bump adam-python version to 0.25.0a0

akmorrow13:


heuermh:
Thank you, @akmorrow13!

PR #2097, opened 2018-12-06
[ADAM-1986] Add java-specific methods where missing.

heuermh:
Fixes #1986.  Also adds java-specific methods where missing, most significantly `GenomicRDD.transformDataset(JFunction)`.

PR #2104, opened 2018-12-18
Add consensus targets to realignment targets

pauldwolfe:
Have been digging into ADAM indel realignment and differences between it and the GATK implementation. While doing so I found that the targets from known sites are never explicitly taken into account as the targetsToAdd method on the ConsensusGenerator trait is never called. This change calls that method as part of the target generation process.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/realignment/RealignIndels.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/rdd/read/realignment/RealignIndels.scala
@@ -55,7 +55,7 @@ private[read] object RealignIndels extends Serializable with Logging {
     unclipReads: Boolean = false): RDD[AlignmentRecord] = {
     new RealignIndels(
       rdd.context,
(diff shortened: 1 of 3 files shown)


heuermh:
Jenkins, retest this please

heuermh:
This looks reasonable to me, but I've introduced a conflict by merging #2097.  Can you rebase and squash commits on this branch?

The way I do it is
```
$ git add remote upstream https://github.com/bigdatagenomics/adam.git
$ git fetch upstream
$ git rebase -i upstream/master
... [p]ick the first commit, [f]ixup the rest as necessary to squash
... edit conflicts
$ git add ... conflicting files
$ git rebase --continue
$ git push --force
```

although admittedly my git-fu is not strong.  See https://git-scm.com/docs/git-rebase#_interactive_mode

heuermh:
Thanks @pauldwolfe, looks good to me.  I'll merge until tomorrow, pending additional review.

heuermh:
Thank you, @pauldwolfe!

PR #2109, opened 2019-01-04
removed anaconda requirement for venv during jenkins test

akmorrow13:


akmorrow13:
@heuermh this should fix the Jenkins problem:

```
InstallError: Error: the following specs depend on 'conda' and can only be installed
into the root environment: anaconda

```

heuermh:
Thank you, @akmorrow13!

PR #2116, opened 2019-01-16
Keep information for <NON_REF> alleles

karenfeng:
INFO and FORMAT fields with Number=<A/R/G> are not properly stored in ADAM for the <NON_REF> allele. This is a problem with both default and non-default fields. During conversion to ADAM, most fields lose any information related to the <NON_REF> allele. This is with the exception of PL, for which the majority of values is stored between `genotypeLikelihoods` and `nonRefLikelihoods`. However, this still drops the value for the ALT/<NON_REF> genotype.

As a result of this bug, the output list for most fields with Number=<A/R/G> does not contain enough values in the resulting gVCF (those with Number=A should have 2 but have 1, those with Number=R should have 3 but have 2, and those with Number=G should have 10 but have 6). For PL, the missing value for the ALT/<NON_REF> genotype contains `3223`.

To fix this problem, this PR does the following:
- The <NON_REF> allele index is passed into the FORMAT functions.
- For non-default fields with Number=<A/R>, the value for the <NON_REF> allele is added to the output list for addition to the attribute map (`VariantAnnotation.attributes` for INFO fields, and `VariantCallingAnnotations.attributes` for FORMAT fields).
- For default fields with Number=<A/R>, the value for the <NON_REF> allele is added to the attribute map keyed as "NON_REF_<FIELD_NAME>" and restored to the output list during extraction if present.
- For all fields with Number=G, the indices include those for REF, ALT, and <NON_REF>.

### DIFF
--- a/adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantContextConverter.scala
+++ b/adam-core/src/main/scala/org/bdgenomics/adam/converters/VariantContextConverter.scala
@@ -338,7 +338,7 @@ class VariantContextConverter(
     try {
       vc.getAlternateAlleles.toList match {
         case List(NON_REF_ALLELE) | List() => {
-          val (coreVariant, variant) = variantFormatFn(vc, None, 0, false)
+          val (coreVariant, variant) = variantFormatFn(vc, None, 0, None, false)
           val v = genotypeVariant(coreVariant, variant)
           val genotypes = vc.getGenotypes.map(g => {
             genotypeFormatFn(g, v, NON_REF_ALLELE, 0, Some(1), false)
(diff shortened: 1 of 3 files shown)


heuermh:
Hello @karenfeng!  As an alternative to this approach, have you considered splitting `<NON_REF>` alleles out to separate records?  Adding all of these nonref allele fields and attributes to every `Variant` record makes me feel like we have the model wrong.

heuermh:
@karenfeng @fnothaft Would it be possible to start a discussion around this issue?  I've done major refactoring of genotypes in https://github.com/bigdatagenomics/adam/pull/2117 but have yet to commit changes to non ref fields in that pull request, waiting for this to be resolved.

heuermh:
Ping @karenfeng @fnothaft for discussion

heuermh:
Closing as WontFix for lack of discussion.

PR #2119, opened 2019-01-21
Filter supplemental reads from scoring

pauldwolfe:
Currently a supplemental read can be added to the sum of base qualities
of a duplicate candidate, artificially inflating its base quality sum
when mark duplicates is trying to choose the best read from the bunch.

This is one (the simplest and most superficial) fix to the problem. Just
filter supplemental reads out when summing.

Another solution could be to use the supplemental flag when setting the
"not primary read flag" in SAMRecordConverter.

Yet another would be to create a separate field in SingleReadBucket for
supplemental reads.

Not sure what the downstream effects of these different approaches are.


heuermh:
See also thread on Gitter https://gitter.im/bigdatagenomics/adam?at=5c45dda9746d4a677aea7673

heuermh:
Thank you, @pauldwolfe!

PR #2126, opened 2019-02-21
Update python and R versions.

heuermh:
See #2089, looks like the same might also be an issue for `adam-python/version.py`.

Should this be set to the current snapshot version `0.27.0-SNAPSHOT` on git head?

### DIFF
--- a/adam-python/version.py
+++ b/adam-python/version.py
@@ -18,7 +18,7 @@
 
 from __future__ import print_function
 
-version = '0.26.0'
+version = '0.26.0a0'
 
 if __name__ == '__main__':
     print(version)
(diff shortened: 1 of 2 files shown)


adam-python/version.py
```diff
@@ -18,7 +18,7 @@
 
 from __future__ import print_function
 
-version = '0.25.0a0'
+version = '0.26.0'
```
akmorrow13: I think this should be '0.26.0a0' after it is pushed to pypi as 0.26.0

heuermh: ok, so that is the rough equivalent to 0.27.0-SNAPSHOT then?

akmorrow13: yep!

heuermh:
Thank you for the review, @akmorrow13

PR #2128, opened 2019-02-25
[ADAM-2127] Update python doc per GenomicRdd --> GenomicDataset change

### LINKED ISSUE #2127: python API does not match API
https://github.com/bigdatagenomics/adam/blob/master/adam-python/bdgenomics/adam/rdd.py#L28
Still has RDD API descriptions, but RDDs have been removed

heuermh:
Fixes #2127

### DIFF
--- a/adam-python/bdgenomics/adam/rdd.py
+++ b/adam-python/bdgenomics/adam/rdd.py
@@ -25,14 +25,14 @@ rdd
 
    GenomicDataset
    VCFSupportingGenomicDataset
-   AlignmentRecordRDD
-   CoverageRDD
-   FeatureRDD
-   FragmentRDD
-   GenotypeRDD
-   NucleotideContigFragmentRDD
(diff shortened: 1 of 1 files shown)


akmorrow13:
This LGTM. Also you were right @heuermh , this will need to be updated on pypi for the changes on readthedocs to be updated. I was looking into this issue last week, but it looks like all the API links on [readthedocs are broken ](https://adam.readthedocs.io/en/latest/api/overview/python.html). They are incorrectly linking to a folder (overview/) when they should link to api/python.html. I have not been able to replicate this issue on my local computer.

heuermh:
Bummer.  We can merge this and #2126 then, and if/when we come up with a fix for #2131, push the 0.26.0a0 release to PyPI.

heuermh:
Thank you for the review, @akmorrow13

PR #2132, opened 2019-03-04
Update htsjdk dependency version to 2.18.2

heuermh:


heuermh:
This newer version of htsjdk bumps the SAM version in the header on write to VN:1.6, causing some unit test comparisons to fail
```
"@HD VN:1.[5] SO:queryname" did not equal "@HD VN:1.[6] SO:queryname"
```

PR #2108, opened 2018-12-28
use parquet-scala_2.11 fork

ryan-williams:
[I built parquet-scala 1.8.3 against Scala 2.11](https://github.com/lasersonlab/parquet-mr/compare/apache-parquet-1.8.3...parquet-scala_2.11-1.8.3) and released it as [`org.lasersonlab.apache.parquet:parquet-scala_2.11:1.8.3`](https://oss.sonatype.org/content/repositories/releases/org/lasersonlab/apache/parquet/parquet-scala_2.11/1.8.3/)

This fixes a petrified old ADAM build-smell: using parquet-scala_2.10 in 2.11 builds.

I'll probably PR some of it upstream, but you can merge this first if you like. I put [a 2.10 version under the same forked `org.lasersonlab.apache.parquet` group](https://oss.sonatype.org/content/repositories/releases/org/lasersonlab/apache/parquet/parquet-scala_2.10/1.8.3/) so the existing cross-build here should work with it.

I started on this because I'm lifting my Spark stack onto 2.12, and `parquet-scala_2.10` does in fact break things there, so I also built and released [`parquet-scala_2.12` for `1.8.3`, `1.10.0`, and a `1.12.0` cut from HEAD](https://oss.sonatype.org/content/repositories/releases/org/lasersonlab/apache/parquet/parquet-scala_2.12/) while I was trying to figure out what I wanted.

Using that {`2.12`,`1.10.0`} version, [my old adam-core fork](https://github.com/hammerlab/adam/) builds+tests against {Spark 2.4, Scala 2.12}, which bodes well for #2044! I plan to publish a version of it soon so I will follow up if I encounter anything else of note.

PR #2140, opened 2019-04-19
Remove dependency on parquet-scala

karenfeng:
As mentioned in #2108 , the available version of parquet-scala is stuck in 2.10. The cleanest way to get around this issue is probably to just remove this dependency.

### DIFF
--- a/adam-cli/src/main/scala/org/bdgenomics/adam/cli/TransformAlignments.scala
+++ b/adam-cli/src/main/scala/org/bdgenomics/adam/cli/TransformAlignments.scala
@@ -19,7 +19,6 @@ package org.bdgenomics.adam.cli
 
 import htsjdk.samtools.ValidationStringency
 import java.time.Instant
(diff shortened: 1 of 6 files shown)


heuermh:
Thank you, @karenfeng.  This is a reasonable idea, as it doesn't look like the issue will be resolved upstream (i.e. there are two pull requests in parquet-mr and neither appear to be close to being merged).

The Jenkins failure seen here requires #2140 to be merged.  I'll trigger a new build here when that happens.

heuermh:
Hello @karenfeng, sorry but to pick up the changes in #2140, you'll need to rebase this pull request branch against master.

heuermh:
Ping @ryan-williams, would you agree this is better than waiting for changes in parquet-mr?

ryan-williams:
This just uses the Parquet Java APIs? That seems like a good workaround to me! Nice work @karenfeng

heuermh:
Thank you, @karenfeng

PR #2144, opened 2019-04-30
[ADAM-2143] Use fold instead of reduce when loading SAM/BAM/CRAM headers

### LINKED ISSUE #2143: Can encounter empty reduce when BAM header fails validation
When we use lenient/silent validation when loading a BAM file whose header contains invalid headerlines, we can trigger an empty reduce.

fnothaft:
Resolves #2143.


heuermh:
Thank you, @fnothaft

PR #2162, opened 2019-05-26
Convert fragment dataset to alignment dataset directly

heuermh:


heuermh:
Thank you @henrydavidge for the original contribution!

PR #2165, opened 2019-05-31
bump adam-python version

akmorrow13:


heuermh:
Oh but wait, this should go on the release branch, right?  Git head should be 0.28.0-something

akmorrow13:
@heuermh updated

PR #2239, opened 2019-12-27
bump python version to pre-release 0.31.0a0

akmorrow13:


heuermh:
Thank you, @akmorrow13

PR #2263, opened 2020-06-25
[ADAM-2262] Update Spark dependency version to 2.4.6

### LINKED ISSUE #2262: Update Spark dependency version to 2.4.6

heuermh:
Fixes #2262

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -21,7 +21,7 @@
     <avro.version>1.8.2</avro.version>
     <scala.version>2.11.12</scala.version>
     <scala.version.prefix>2.11</scala.version.prefix>
-    <spark.version>2.4.5</spark.version>
+    <spark.version>2.4.6</spark.version>
     <parquet.version>1.10.1</parquet.version>
     <!-- Edit the following line to configure the Hadoop (HDFS) version. -->
     <hadoop.version>2.7.5</hadoop.version>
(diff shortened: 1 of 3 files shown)


heuermh:
Hello @shaneknapp, I thought perhaps the issue with the first Jenkins test failure above might have been insecure `http` access, but that appears not to be the case.  Are there some other network restrictions in place?

```
+ curl \
-L https://www-us.apache.org/dist/hadoop/common/hadoop-2.7.7/hadoop-2.7.7.tar.gz \
-o hadoop-2.7.7.tar.gz
...
curl: (7) Failed to connect to downloads.apache.org port 443: Connection timed out
```
https://amplab.cs.berkeley.edu/jenkins/job/ADAM-prb/3099/HADOOP_VERSION=2.7.5,SCALAVER=2.12,SPARK_VERSION=2.4.5,label=ubuntu/console

shaneknapp:
sigh.  i wonder if we were blacklisted again.

heuermh:
Ah, perhaps I can rework it to use the `https://www.apache.org/dyn` mirror instead . . . will try shortly

shaneknapp:
yeah, try that...  but if we're blacklisted, that means that something
isn't cached locally that should be.  i don't think it's this build that's
broken, but you're just seeing the impact.

this has happened before...  :)

just opened an apache infra jira to help me figure it out:
https://issues.apache.org/jira/browse/INFRA-20463

shaneknapp:
word.

using the mirror is the best bet, and apache infra found a few servers that
still had us blacklisted and fixed that as well.

we should be g2g.

heuermh:
Awesome!  Thank you for your help, @shaneknapp, as always.

Issue #2321, opened 2021-06-09
saveAsSam format


hxdhan:
when I use this code to generate sam file 
```
val fragments = sc.loadFastq("36_11.fastq", Option("36_22.fastq"))
fragments.saveAsSam("test.sam", SAMFormat.SAM, true, false)
```

I got

@HD	VN:1.6	SO:unsorted
SRR8418436.2 2	77	*	0	0	*	*	0	0	GGAAATTTAAAAAAATACACATGGCCAGGCCCCAGCCCAAATCACTAATAAGAATCTCCAGGGCTTCACCTGTTAGACTGGCAAAAATCCAAAAGTAAACA	@@@BDA:?FHHHDHGFHBF9A?EGE@FHG<BG@DHGE;;B2=@GIDGDCFHHIBGHIIEEFC8?@B(.;>CA>>>;ACC<<?CCB9?9::>:>@C:@@CC>
SRR8418436.2 2	141	*	0	0	*	*	0	0	CACCAGCAATGTGTAGGAATACCTGTTTCTCCACAAAGTGTTTACTTTTGGATTTTTGCCAGTCTAACAGGTGAAGCCCTGGAGATTCTTATTAGTGATTT	?@<BDDABCCBFDHA9F@E@FHDEBEHI@DEBDAHIII1:D@GB@FHIB?@??DGDHG>CGB==C>;F@@C=C;??E=CBB?;AC<CCCDCCA>ACA35@@

but when I use another tool (example picard) I got
@HD     VN:1.6  SO:queryname
@RG     ID:A    SM:1
SRR8418436.2    77      *       0       0       *       *       0       0       GGAAATTTAAAAAAATACACATGGCCAGGCCCCAGCCCAAATCACTAATAAGAATCTCCAGGGCTTCACCTGTTAGACTGGCAAAAATCCAAAAGTAAACA   @@@BDA:?FHHHDHGFHBF9A?EGE@FHG<BG@DHGE;;B2=@GIDGDCFHHIBGHIIEEFC8?@B(.;>CA>>>;ACC<<?CCB9?9::>:>@C:@@CC>   RG:Z:A
[...]

PR #2325, opened 2021-07-03
trim sam format readname 

hxdhan:
https://github.com/bigdatagenomics/adam/issues/2321


heuermh:
Thank you for your contribution, @hxdhan.

I don't think we should change how ADAM parses read names across the board, rather for samples with odd names in them, one can simply transform the records before writing

```scala
import org.bdgenomics.adam.sql.Alignment

def trimReadName(a: Alignment): Alignment = {
  a.copy(readName = Some( /* ... do something with readName here ... */ ))
}

val transformed = alignments.transformDataset(ds => ds.map(trimReadName(_))
```

